PR #4, opened 2020-07-22
Updates from team feedback

gregjopa:
We had a great meeting on Monday about paypal-js. This PR incorporates that feedback. The README is probably the easist place to start with reviewing these changes. Changes include:

- Renaming the `getScript()` function to `loadScript()`
- Adding support for setting both attributes and query parameters through the `options` argument.
- Removing the `camelCaseToCabobCase()` functionality for the options. Just pass in the string keys just like how they are defined in [our documentation](https://developer.paypal.com/docs/checkout/reference/customize-sdk/#query-parameters) (ex: `{ params: { 'client-id': 'sb' }, attributes: { 'data-order-id': '12345' } }`).
- Using `defer` for script loading to prevent script execution from delaying page render time.
- Renaming the names of the js bundles in `dist` to make more sense. IIFE is confusing for everyone 😆 
- Removing the `window.paypalLoader` global window option. For that use case, we are proposing a `callback` solution w/ the JS SDK directly instead.

A couple other notes:
- I simplified the react example to just be a single basic component. I do think we should encourage preloading but I think the new react-paypal-js library is the place for that example and documentation as we figure out the design in our next sprint.
- I added some basic caching to avoid reloading the script when it's already loaded (https://engineering.paypalcorp.com/jira/browse/DTCHKINT-72). It's a good start but I think there are more edge cases to figure out with the "reloading" story and supporting multiple components on a page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,32 +26,3 @@ loadScript({ 'client-id': 'sb' })
         paypal.Buttons().render('#your-container-element');
     });
 ```
-
-### Using paypal-js with React
-
-```js
-import React, { useEffect, useState } from 'react';
-import ReactDOM from 'react-dom';
-import { loadScript } from '@paypal/paypal-js';
-
-const loadingPromise = loadScript({ 'client-id': 'sb' });
-let PayPalButtonsFromDriver;
-
-export default function Checkout() {
-    const [isLoaded, setIsLoaded] = useState(false);
-
-    useEffect(() => {
-        loadingPromise
-            .then(paypal => {
-                PayPalButtonsFromDriver = paypal.Buttons.driver('react', { React, ReactDOM });
-                setIsLoaded(true);
-            });
-    });
-
-    if (isLoaded) {
-        return <PayPalButtonsFromDriver />;
-    }
-
-    return 'loading';
-}
-```


README.md
```diff
@@ -45,50 +45,29 @@ window.paypalLoader.getScript({ clientID: 'sb' })
...
+export default function Checkout() {
     const [isLoaded, setIsLoaded] = useState(false);
 
     useEffect(() => {
-        props.loadingPromise
+        loadingPromise
             .then(paypal => {
-                paypal.Buttons().render('#paypal-button-container');
-                PayPalButtons = paypal.Buttons.driver('react', { React, ReactDOM });
+                PayPalButtonsFromDriver = paypal.Buttons.driver('react', { React, ReactDOM });
```
mnicpt: We should abstract this from the customer into a `PayPalButtons` component.  The `PayPalButtons` component is dependent on the script having loaded and will only display if it has otherwise throw a warning for the developer to ensure the script loads.

mnicpt: If the customer wants to load the script on Checkout, then they conditionally render the `PayPalButtons` component when loaded.

mnicpt: Something along these lines: 
```
function Checkout() {
  const [isLoaded, setLoaded] = useState(window.paypal);
  const paypalConfig = {
    params: {
      'client-id': 'sb',
      'currency': 'USD',
      'commit': true,
    },
    attributes: {
      'data-csp-nonce': 'yo'
    },
    async: true
  };

  useEffect(() => {
    if (!isLoaded) {
      loadPaypalScript(
        paypalConfig
      )
      .then(paypal => {
        console.log('Loaded:', JSON.stringify(paypal));
        setLoaded(true);
      })
      .catch(err => {
        console.error(`Error found: ${JSON.stringify(err)}`);
      });
    }
  }, [isLoaded, paypalConfig]);

  if (!isLoaded) return null;

  const createOrder = (data, actions) => {
    return actions.order.create({
      purchase_units: [
        {
          amount: {
            value: "0.01",
          },
        },
      ],
    });
  };

  const onApprove = (data, actions) => {
    return actions.order.capture();
  };

  return (
      <>
          <PayPalButtons
            createOrder={(data, actions) =>  createOrder(data, actions)}
            onApprove={(data, actions) => onApprove(data, actions)}
          />
      </>
  );
}

export default Checkout;
```

README.md
```diff
@@ -11,32 +11,32 @@ To get started, install paypal-js with npm.
...
 
 ```html
-<script src="dist/paypal.iife.min.js"></script>
+<script src="dist/paypal.min.js"></script>
 <div id="paypal-button-container"></div>
 ```
 
 ```js
-window.paypalLoader.getScript({ clientID: 'sb' })
+window.paypalLoader.loadScript({ 'client-id': 'sb' })
```
mstuart: Think we should use window.paypal? If all of the components safely merge onto `window.paypal` without over-writing it. If it's safe to use, maybe we can use it, so it's 1 less namespace to expect PayPal things to be on. I don't know if it's a big deal though. Maybe not! 

/cc @bluepnume

README.md
mstuart: For sure! It'd be nice to hide the driver and loading state stuff. I think this is an intermediate step while react-paypal-js is being built out. When we launch react-paypal-js, maybe we can update this section to point at that module instead?

e2e-tests/esm/index.test.js
```diff
@@ -14,14 +14,18 @@ describe('EcmaScript Module (ESM)', () => {
...
-
         const paypalVersion = await page.evaluate(() => {
             return window.paypal.version;
         });
 
         expect(paypalVersion.startsWith('5')).toBe(true);
     });
 
+    it('the script should have the data-order-id attribute', async () => {
+        const orderId = await page.$eval('head script', element => element.getAttribute('data-order-id'));
```
mstuart: puppeteer's so cool

README.md
mnicpt: I like that better.

src/constants.js
```diff
@@ -0,0 +1,23 @@
+export const SDK_BASE_URL = 'https://www.paypal.com/sdk/js';
+
+export const VALID_ATTRIBUTES = [
+    'data-partner-attribution-id',
+    'data-csp-nonce',
+    'data-order-id'
+];
+
+export const VALID_QUERY_PARAMETERS = [
```
mstuart: For a later enhancement, but I wonder if we should console.error() if these properties aren't passed? If someone were to mistakenly pass duplicate properties, would the first one be picked out?

I wonder if our opportunity to do that in the SDK has passed. I guess console.error() is a development error, not an error. If we started throwing errors, that'd probably block a bunch of payments :D

README.md
```diff
@@ -45,50 +45,29 @@ window.paypalLoader.getScript({ clientID: 'sb' })
...
-```
-
-```js
-// PayPalButtonsWrapper.js
 import React, { useEffect, useState } from 'react';
 import ReactDOM from 'react-dom';
+import { loadScript } from '@paypal/paypal-js';
 
-let PayPalButtons;
+const loadingPromise = loadScript({ 'client-id': 'sb' });
```
mnicpt: Would rather not set this into a promise variable and just have inline in `useEffect`.  I can't imagine the performance is that much greater.

src/constants.js
```diff
@@ -0,0 +1,23 @@
...
+    'client-id',
```
mstuart: I'm still on the fence on whether or not we should have a list of valid properies. Thinking it'll be out of date unless these Arrays were in paypal-sdk-constants and were pulled in here.

e2e-tests/esm/index.html
```diff
@@ -5,10 +5,12 @@
...
   </head>
   <body>
+    <h1>ESM Demo</h1>
+
     <script type="module">
-        import { getScript } from '../../dist/paypal.esm.js';
+        import { loadScript } from '../../dist/paypal.esm.js';
 
-        getScript({ clientID: 'sb' })
+        loadScript({ 'client-id': 'sb', 'data-order-id': '12345' })
```
mnicpt: `loadScript` to me is seeming to generic from a client perspective especially if it's loaded alongside others.  Maybe `PayPal.loadScript()` or `PayPal.script()` or as @mstuart suggested below, `window.paypal.loadScript()`?

src/constants.js
mnicpt: I don't think we should.  We should only use what is valid into the SDK and only throw errors if a required one is missing.   Then there is no maintenance.  The SDK does require a change to support this as we spoke about earlier.

src/utils.js
```diff
@@ -1,27 +1,49 @@
+import { VALID_ATTRIBUTES, VALID_QUERY_PARAMETERS } from './constants';
+
 function loadError() {
     throw new Error(`The script "${this.src}" didn't load correctly.`);
 }
 
-export function insertScriptElement(url, onloadFunction) {
+export function insertScriptElement({ url, attributes = {}, defer, callback }) {
```
mnicpt: So you're thinking that the query params come in with the url?  Just wondering if we should have a params arg too where they could define a config and pass the entire thing in?
```
  const paypalConfig = {
    params: {
      'client-id': 'sb',
      'currency': 'USD',
      'commit': true,
    },
    attributes: {
      'data-csp-nonce': 'yo'
    }
  };

  useEffect(() => {
    loadPaypalScript(
      paypalConfig
    )
```

README.md
```diff
@@ -11,32 +11,32 @@ To get started, install paypal-js with npm.
...
+<script src="dist/paypal.min.js"></script>
 <div id="paypal-button-container"></div>
 ```
 
 ```js
-window.paypalLoader.getScript({ clientID: 'sb' })
+window.paypalLoader.loadScript({ 'client-id': 'sb' })
     .then(paypal => {
         paypal.Buttons().render('#paypal-button-container');
     });
```
BriceShatzer: Since the `dist` folder doesn't exist without running a build, you can't really say that you're able to use `paypal-js` without a build tool. That being said there's two different approaches we could take, either:  

- include the dist folder as part of the repo, and explain that they will need to serve a copy of that file that can be referenced by the script tag on their page. 

- or provide a static copy of the built and minified version of the file via a CDN. This is what Stripe are [doing](https://github.com/stripe/stripe-js#manually-include-the-script-tag) with  `stripe-js` and I think this is by-far the most preferable solution. A grosser idea along these lines would be to include the the `dist` folder and then set up a DNS Alias pointing to the raw repo file  
(ie: `paypal.com/paypal-js` => `https://raw.githubusercontent.com/paypal/paypal-js/main/dist/paypal.min.js`)

e2e-tests/esm/index.html
mstuart: +1 for keeping it loadScript, plus they can rename it if it's confusing w/ another "loadSomething" function :)

It perfectly 👌 describes what it's doing. It's loading, not rendering or anything else!

README.md
gregjopa: I used `performance.now()` to test this out. In my basic create-react-app demo its about 30ms faster with this approach. I think the performance gain is worth it. 

Screen Shot 2020-07-22 at 5 17 15 PM

Screen Shot 2020-07-22 at 5 17 03 PM

src/utils.js
gregjopa: I'm open to that. We also need a way to pass in properties like `defer` since attributes and properties are different. Should the config have three sections like so?

```js
const paypalConfig = {
    params: {
      'client-id': 'sb',
      'currency': 'USD',
      'commit': true,
    },
    attributes: {
      'data-csp-nonce': 'yo'
    },
    properties: {
     defer: true 
    }
  };

loadScript(paypalConfig)
```

e2e-tests/esm/index.html
gregjopa: I vote for loadScript as well. I think the `import` statement makes it clear that it's coming from paypal.

```js
import { loadScript } from '@paypal/paypal-js';
```

src/constants.js
gregjopa: I struggled with this one too. The nice thing with these "valid lists" is they enable having a single flat object as the argument to loadScript like it is now. The downside is these "valid lists" will most likely get out-dated. 

I'm fine with removing them and using the nested object approach for namespacing things. Ex:
```js
const paypalConfig = {
    params: {
      'client-id': 'sb',
      'currency': 'USD',
      'commit': true,
    },
    attributes: {
      'data-csp-nonce': 'yo'
    },
    properties: {
     defer: true 
    }
  };
```

README.md
gregjopa: Yep I'd consider this temporary until react-paypal-js is built out. Eventually this paypal-js doc will point to that repo for guidance on implementing with react.

README.md
gregjopa: Sure I'm fine with changing it to `window.paypal.loadScript` for the IIFE bundle. I was just nervous about adding anything to the paypal namespace in-case there are naming conflicts down the road.

e2e-tests/esm/index.test.js
gregjopa: Agreed! It's great 💯

README.md
gregjopa: Great point @BriceShatzer! I do like the CDN suggestion.

README.md
gregjopa: Unfortunately it causes an error when trying to use `window.paypal.loadScript`. The JS SDK seems to throw the following error when `window.paypal` already exists and `paypal.version` doesn't match.

js-sdk-error-existing-namespace-screenshot

Let's keep it as `window.paypalLoader.loadScript` for now.

src/constants.js
gregjopa: I updated the PR to use this approach and dropped the constants.

mstuart: Do you guys prefer to pass in flattened parameters like `loadScript({ ... })` or `loadScript({ params: {}, attributes: {}, ... })`? The first option is pretty slick! Sorry, my first comment was about whether or not we should maintain a list of properties here or look at how we can keep them in sync w/ the SDK repos.

Know if the SDK throws errors for unknown params? If so, maybe they have a list somewhere that we can hoist up into paypal-sdk-constants, then share them w/ paypal-js.

README.md
mstuart: +1 for CDN for those that prefer it. More options the better!

src/constants.js
gregjopa: > Do you guys prefer to pass in flattened parameters like loadScript({ ... }) or loadScript({ params: {}, attributes: {}, ... })? The first option is pretty slick! 

I'm torn on this. I could go either way. Here's my thinking so far:

###  Flattened parameters `loadScript({ ... })`

**Pros**
- Simple and familiar. Just a flat object for configuration.
- Can depend on [paypal-sdk-constants](https://github.com/paypal/paypal-sdk-constants/blob/master/src/params.js) to know what query params and attributes are valid.
- Ability to `console.warn()` when invalid inputs are passed in.

**Cons**
- Potential to do the wrong thing in the future if there are naming collisions. A developer wants a key/value to be a query param and not a data attribute or vice versa. Not sure if that would ever actually happen 🤔  

###  Namespaced parameters `loadScript({ params: {}, attributes: {}, properties: {} })`

**Pros**
- Less code to maintain
- Developer can explicitly define whats a param vs what's an attribute.
- Can also integrate with `paypal-sdk-constants` to `console.warn()` when invalid inputs are passed. 

**Cons**
- Less familiar api for developers.  

### Developer Experience for JS SDK loading errors

> Know if the SDK throws errors for unknown params? If so, maybe they have a list somewhere that we can hoist up into paypal-sdk-constants, then share them w/ paypal-js.

This "error logging" use case is really interesting. @mnicpt mentioned this earlier in the thread as well. The JS SDK approach of throwing an error for invalid params doesn't work as I'd expect. I'd expect that code to execute and to see that error in the Dev Tools console, but I don't. For example, here's what the JS SDK script returns as a response when invalid input is passed in (note the 400 status code and the javascript in the response):

Screen Shot 2020-07-23 at 9 59 14 AM

I did some testing locally and it turns out that `<script src="https..."></script>` will only execute the contents of the response when a valid status code is returned. When it returns a 400 status code for bad input, It will not execute the contents of the response as javascript. So the developer is forced to go check the Dev Tools network tab to see the http response and learn exactly why their request failed. 

We could have paypal-js improve the dev experience around logging when there is bad input for either approach.

README.md
mnicpt: +1

README.md
mnicpt: @gregjopa Looks like you're loading the script twice.  Are you doing something like: `window.paypal.loadScript = ...` after the paypal script loads the first and only time?

README.md
mnicpt: But we really don't want to promote this approach at all for SPA's.  And for the odd case that they would want to, like they have don't care about performance as much so won't preload, then performance isn't an issue.  The next big issue I have with this is the name `loadingPromise`.  Speaks more to implementation details than what it is.

README.md
gregjopa: Here's the order of what I tried:

1. Load paypal-js and save it to the paypal namespace:
```js
  window.paypal = window.paypal || {};
  window.paypal.loadScript = window.paypalLoader.loadScript
```

2. Call `window.paypal.loadScript(options)` to load the JS SDK.

After doing those steps i get that error.


This error can also be reproduced without paypal js. Here's a minimal reproduction. You will get this error when running the following:

```html
<script>
    window.paypal = {};
</script>

<script src="https://www.paypal.com/sdk/js?client-id=sb"></script>
```

src/utils.js
```diff
@@ -15,31 +13,14 @@ export function insertScriptElement({ url, attributes = {}, defer, callback }) {
 
     document.head.appendChild(newScript);
     newScript.src = url;
-    newScript.defer = defer;
+    newScript.defer = properties.defer === undefined ? true : properties.defer;
 }
 
 export function findScript(url) {
```
mnicpt: Why do we need this?

README.md
```diff
@@ -11,32 +11,32 @@ To get started, install paypal-js with npm.
...
 
-getScript({ clientID: 'sb' })
+loadScript({ 'client-id': 'sb' })
     .then(paypal => {
         paypal.Buttons().render('#your-container-element');
     });
 ```
 
-### `getScript`
+### `loadScript`
```
mnicpt: I feel like this should be above the `loadScript` example, immediately after the import section.

mnicpt: I would also categorize this section with something like: `NPM` heading and `PayPal Snippet` inside `Installation and usage`.  So `Using paypal-js without a build tool` in this case would be replaced with `PayPal Snippet`.

src/utils.js
gregjopa: This was originally added to help support caching the JS SDK to avoid re-loading and evaluating it when it's already been loaded. I think you're right that we don't need this `findScript()` function. Originally I thought this was a valid use case that we need to support where the JS SDK script is loaded twice (once for `buttons` and once for `messages`):

```html
    <script src="https://www.paypal.com/sdk/js?client-id=sb&components=buttons"></script>
    <script src="https://www.paypal.com/sdk/js?client-id=sb&components=messages"></script>

    <script>
      paypal.Buttons().render('#paypal-buttons');
    </script>
    <!-- Credit Messaging -->
    <div id="paypal-credit-messaging" data-pp-message data-pp-placement="product" data-pp-amount="500.00"></div>

    <!-- PayPal Buttons -->
    <div id="paypal-buttons"></div>
```

When I tested this locally, it doesn't work and it also breaks in different ways depending on if buttons or messages is loaded first. Based on this, I've come to the conclusion that the JS SDK should be loaded only once per page with utilizing the comma separated list option for components (`components=buttons,messages`).

The `findScript()` function was added so it checked both the script url and the `window.paypal` object to determine if the scripts already loaded. Since we can only have one script on the page this seems unnecessary. I'll go ahead and remove this `findScript()` function. And just leave the `window.paypal` check to determine if the libraries already been loaded or not.

README.md
mnicpt: I'm struggling with this pattern.  Would love to have a conversation around this.  I like it when libraries that want to have an async way to load their scripts, to either provide a replay queue or an async snippet.  This eliminates any additions to the global namespace.  Even if we include triggering an event inside this snippet, the customer could take action upon its trigger.

src/constants.js
gregjopa: We talked through the Pro/Con list and decided to go with the flattened parameters approach. And using `data-` to determine whats an attribute vs a query param.

README.md
gregjopa: The CDN is a great idea. But no longer necessary for paypal js since we will solve for this use case directly in the JS SDK w/ the callback option.

src/utils.js
gregjopa: The `findScript()` function has been removed. Just checking `window.paypal` should be sufficient for determining if the script has already been loaded.

README.md
gregjopa: I separated out Install and and Usage into separate sections and also dropped the `Using paypal-js without a build tool` section all together.

README.md
gregjopa: The react implementation is still a WIP. We will iron on this details next week.

src/constants.js
gregjopa: This is definitely a possibility for a future enhancement. It will really depend on if we change the "400 bad input" behavior of the JS SDK or not.

README.md
gregjopa: We discussed alternatives like the replay queue and in the end, adding a callback to the JS SDK directly seems like the simplest and eliminates a need for paypal-js all together for the "no webpack or build tool" use case.

README.md
```diff
@@ -2,93 +2,56 @@
...
-        paypal.Buttons().render('#your-container-element');
-    });
-```
+## Usage
 
-### `getScript`
+Import the `loadScript` function for asynchronously loading the Paypal JS SDK.
 
-- accepts an object for passing query parameters to the JS SDK. The camelCase keys are converted to cabob-case before passing them as query parameters.
+`loadScript(options)`
```
mnicpt: I think we should make this a header as we will be adding to this section to load components and set page type I think.  `#### loadScript(options)`

gregjopa: Good point. I updated it to a heading 3 tag.

PR #6, opened 2020-07-24
Publish configuration

gregjopa:
This PR includes a few changes to ensure `npm publish` works properly:
- Adds a license
- Updates the package.json with main/module/files/license settings.
- Uses the `prepack` hook to build the project before publishing (https://docs.npmjs.com/misc/scripts)

To test this out locally you can run `npm pack` which does the same thing as `npm publish` but doesn't actually publish to any registry. I did this and copied the tarball over to my demo app and was able to successfully reference it as a dependency. 

Here's the demo app: https://github.paypal.com/gjopa/create-react-app-with-paypal-js
- note the syntax for referring to a local file: https://github.paypal.com/gjopa/create-react-app-with-paypal-js/blob/master/package.json#L6
- the `import` statement to bring in `loadScript`: https://github.paypal.com/gjopa/create-react-app-with-paypal-js/blob/master/src/Checkout.js#L3

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,18 +2,24 @@
   "name": "@paypal/paypal-js",
   "version": "0.1.0",
   "description": "Async loader for the JS SDK",
-  "main": "dist/paypal.js",
+  "main": "dist/paypal.node.js",
   "module": "dist/paypal.esm.js",
   "scripts": {
     "build": "rollup --config",
     "lint": "eslint .",
+    "prepack": "npm run build",
     "test": "jest --env=jsdom src",
     "test:e2e": "jest -c e2e-tests/jest.config.js",
     "test:e2e:start": "http-server"
   },
-  "keywords": [],
-  "author": "",
-  "license": "ISC",
+  "files": [
+    "dist"
+  ],
+  "license": "Apache-2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -2,18 +2,24 @@
...
-  "main": "dist/paypal.js",
+  "main": "dist/paypal.node.js",
   "module": "dist/paypal.esm.js",
   "scripts": {
     "build": "rollup --config",
     "lint": "eslint .",
+    "prepack": "npm run build",
     "test": "jest --env=jsdom src",
     "test:e2e": "jest -c e2e-tests/jest.config.js",
     "test:e2e:start": "http-server"
```
BriceShatzer, line 13: Would it make sense to chain these together into a single command? something like   
`"test:e2e": "http-server & jest -c e2e-tests/jest.config.js",`

gregjopa: From what I can tell, jest-puppeteer needs a separate script to call for starting up the server: https://github.com/paypal/paypal-js/blob/main/jest-puppeteer.config.js#L12. It handles the concurrency logic for us with running the web server in one process and the tests in another process.

Issue #5, opened 2020-07-24
Show how to set query, attribute, and options as params


mnicpt:
**Query**: loadScript({ 'client-id': 'sb' });

**Attribute**: loadScript({ 'client-id': 'sb', 'data-csp-nonce': 'asdf' });

**Options**: loadScript({ 'client-id': 'sb', 'data-csp-nonce': 'asdf', defer: true });

Link to available query params (https://developer.paypal.com/docs/checkout/reference/customize-sdk/#query-parameters) and attributes (https://developer.paypal.com/docs/checkout/reference/customize-sdk/#script-parameters).

PR #8, opened 2020-07-27
Add documentation for setting query params, script params, and defer

gregjopa:
This PR documents the `options` that can be passed into `loadScript()`. It resolves issue #5.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -17,13 +17,13 @@ export function loadScript(options) {
 
         isLoading = true;
 
-        const { queryString, attributes, properties } = processOptions(options);
+        const { queryString, dataAttributes, scriptAttributes } = processOptions(options);
         const url = `${SDK_BASE_URL}?${queryString}`;
 
         insertScriptElement({
             url,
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -17,13 +17,13 @@ export function loadScript(options) {
...
 
-        const { queryString, attributes, properties } = processOptions(options);
+        const { queryString, dataAttributes, scriptAttributes } = processOptions(options);
         const url = `${SDK_BASE_URL}?${queryString}`;
 
         insertScriptElement({
             url,
-            attributes,
-            properties,
+            dataAttributes,
```
gregjopa, line 25: Originally I thought `defer` and `async` were properties. I was wrong. They are script attributes (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Attributes). So I renamed:
- `properties` => `scriptAttributes`
- `attributes` => `dataAttributes`

Note that public `loadScript()` api has not changed. These variable names are mainly for the functions in utils.js

PR #9, opened 2020-07-28
Support reloading the <script> when attributes change

gregjopa:
## Problem

[Our current approach](https://github.com/paypal/paypal-js/blob/main/src/main.js#L16) for checking to see if the JS SDK is already loaded does not work for the reload use case when script query parameters and attributes need to change. 

## Solution

This PR updates the logic for determining if the JS SDK has already been loaded. In addition to checking for the existence of `window.paypal`, it also checks to see if that same url and query string have already been loaded in a script tag on the page.

## How to verify this works

This PR also includes new e2e tests for verifying that script reloading works with the JS SDK when the currency changes. ~~I had to comment out the `window.paypal` check in paypal-js to get this to work. With this demo I learned that we have to remove the existing script tag. I noticed that if this is not done the smart buttons don't actually pick up the currency changes. Not sure if this is a bug or by design~~

Screen Shot 2020-07-28 at 11 34 45 AM

Run the e2e tests with `npm run test:e2e`

## Additional Info

Note how we don't call `removeChild()` to remove the previous JS SDK script that's in the DOM. The JS SDK appears to take care of this during the second pass. Here's a gif of how the previous script tag is removed whenever the currency changes. The code I wrote does not remove it. This seems to be done by the JS SDK.

currency-demo

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -18,7 +18,7 @@ export function insertScriptElement({ url, dataAttributes = {}, scriptAttributes
     document.head.insertBefore(newScript, document.head.firstElementChild);
 
     newScript.src = url;
-    newScript.defer = scriptAttributes.defer === undefined ? true : scriptAttributes.defer;
+    newScript.defer = scriptAttributes.defer ?? true;
 }
 
 export function processOptions(options = {}) {


e2e-tests/reload-script/index.test.js
```diff
@@ -0,0 +1,32 @@
+import { baseURL } from '../test-helper';
+
+describe('Reload script', () => {
+    let sdkResponse, smartButtonsResponse;
+
+    beforeEach(async () => {
+        [, sdkResponse, smartButtonsResponse] = await Promise.all([
+            page.goto(`${baseURL}/e2e-tests/reload-script/index.html`, {waitUntil: 'networkidle2'}),
+            page.waitForResponse(response => response.url().startsWith('https://www.paypal.com/sdk/js')),
+            page.waitForResponse(response => response.url().startsWith('https://www.sandbox.paypal.com/smart/buttons'))
```
gregjopa, line 10: This e2e test verifies that the same currency value is also passed through to the smart buttons (https://www.sandbox.paypal.com/smart/buttons) request. I noticed that if you don't prepend the new sdk script element then currency doesn't pass through to smart buttons as expected.

src/utils.js
```diff
@@ -11,7 +15,8 @@ export function insertScriptElement({ url, dataAttributes = {}, scriptAttributes
         newScript.setAttribute(key, dataAttributes[key]);
     });
 
-    document.head.appendChild(newScript);
+    document.head.insertBefore(newScript, document.head.firstElementChild);
+
     newScript.src = url;
     newScript.defer = scriptAttributes.defer === undefined ? true : scriptAttributes.defer;
```
mnicpt: `scriptAttributes.defer ?? true`...I'm pretty sure we can use this now.

gregjopa: Great point! 🔥

I applied this change and it seems to transpile with no issues. Ex:

```js
 newScript.defer = scriptAttributes.defer ?? true;
```

Becomes:

```js
newScript.defer = (_scriptAttributes$def = scriptAttributes.defer) !== null && _scriptAttributes$def !== void 0 ? _scriptAttributes$def : true;
```

Issue #13, opened 2020-08-05
Browser Support - String.prototype.startsWith() is not polyfilled


gregjopa:
Babel/Rollup is not transpiling `startsWith()` which breaks in older browsers (`es.string.starts-with { "chrome":"49", "edge":"18", "ie":"11" }`)

Possible solutions:
1. Wire up babel preset-env to dynamically include polyfills like this based on the browser support we target.
2. Manually include a polyfill for `startsWith()`
3. Don't use `key.startsWith('data-')` and instead use `key.substring(0, 5) === 'data-'`

One thing to keep in mind with options 1 and 2 is we want to keep the overall bundle size small. Including too many polyfills will bloat this micro-module. I also tested this out in browserstack in IE11. It techincally works if we provide a promise polyfill and don't use `startsWith()`.

Thoughts on this?

PR #14, opened 2020-08-06
Rollup updates

gregjopa:
This PR updates the browser support for paypal-js and resolves issue #13 

It includes a few specific changes:
- Improves browser support - The [babel-preset-env](https://babeljs.io/docs/en/babel-preset-env) option is used to add in polyfills at compile time based on the [browserslist](https://github.com/browserslist/browserslist) file.
- Adds dist/paypal.browser.js - this new output file is designed to be used from the CDN. It also makes it way easier to test in IE 11 with Browserstack.
- Bundle-size plugin - display the bundle-sizes in the terminal when building with `npm run build`

### Testing in IE 11 w/ Browserstack
Screen Shot 2020-08-06 at 1 29 59 PM
 
### Bundle size metrics from build time

Screen Shot 2020-08-06 at 1 34 29 PM

### DIFF
--- a/e2e-tests/browser-global/index.html
+++ b/e2e-tests/browser-global/index.html
@@ -7,9 +7,7 @@
   <body>
     <h1>Demo with window.paypalLoadScript</h1>
 
-    <!-- bring your own promise polyfill -->
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/promise-polyfill/8.1.3/polyfill.js"></script>
-    <script src="../../dist/paypal.browser.js"></script>
+    <script src="../../dist/paypal.browser.min.js"></script>
     <script>
         var options = {
             'client-id': 'sb',


.browserslistrc
```diff
@@ -0,0 +1,7 @@
+# Copied from https://github.com/krakenjs/grumbler-scripts/blob/master/config/.babelrc-browser
```
gregjopa, line 1: I learned that browserslist supports sharable configs. We could create a shareable browerslist module in the future: https://github.com/browserslist/browserslist#shareable-configs

rollup.config.js
```diff
@@ -20,6 +29,13 @@ export default {
         {
             file: pkg.main,
             format: 'cjs'
+        },
+        {
+            file: 'dist/paypal.browser.js',
+            format: 'iife',
+            name: 'paypalLoadScript',
+            // update the default export to be the loadScript() function
+            footer: 'paypalLoadScript = paypalLoadScript.loadScript;'
```
gregjopa: This `footer` option is kind of a hack. It's used to simplify the export that's assigned to `window.paypalLoadScript` so it's just the main function. Without this change it would need to be called like so:

```js
window.paypalLoadScript.loadScript()
```

babel.config.js
```diff
@@ -1,12 +1,33 @@
...
+                '@babel/preset-env',
+                {
+                    useBuiltIns: 'usage',
+                    corejs: {
+                        version: '3.6',
+                        proposals: true
+                    },
+                    modules: false,
+                    exclude: [
+                        "es.promise"
```
gregjopa: This promise polyfill nearly doubles the size of the bundle. It's purposely excluded with the idea that developers would bring their own Promise polyfill to support IE 11.

e2e-tests/browser-global/index.html
```diff
@@ -0,0 +1,25 @@
...
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Demo with window.paypalLoadScript | PayPal JS</title>
+  </head>
+  <body>
+    <h1>Demo with window.paypalLoadScript</h1>
+
+    <!-- bring your own promise polyfill -->
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/promise-polyfill/8.1.3/polyfill.js"></script>
```
gregjopa: Note this Promise polyfill so this html page works in IE 11.

babel.config.js
mnicpt: @gregjopa Let's chat about this.

gregjopa: @mnicpt I agree. Let's sync up and figure out the best approach. I was researching the polyfill subject more and wish we could do something simple like this:

```html
<!-- Smaller polyfill bundle for browsers with ES2015+ support -->
<script type="module" src="../../dist/paypal.browser.modern.js"></script>

<!-- Full polyfill bundle for old browsers -->
<script nomodule src="../../dist/paypal.browser.legacy.js"></script>
```

But Edge and Safari don't fully support this yet: https://gist.github.com/jakub-g/5fc11af85a061ca29cc84892f1059fec.

gregjopa: Here's a couple interesting posts on the subject:
- https://3perf.com/blog/polyfills/
- https://jasonformat.com/modern-script-loading/

PR #19, opened 2020-08-12
Reduce bundle size

gregjopa:
## Summary

This PR reduces the bundle-size for paypal-js and makes the bundles about 75% smaller.

There are tradeoffs to this approach and it's unclear if we want to merge this PR or not. I created this PR to get feedback from the team.

## Details

The thought is paypal-js is a micro library and we should do everything we can to keep it as small as possible while still supporting the same browsers as the JS SDK. Developers will need to be mindful of this strategy in order for this approach to be successful.

### When to use new JS features

In order support IE 11, our JS code needs to be traspiled down to ES3 since that's the way babel-preset-env/core-js/browserslist work to fully support IE 11. Many new features of JS easily transpile down to ES3 and don't require polyfills. We can continue using these features like `const/let`, destructuring, arrow functions, and default function argument values. 

### When to avoid new JS features

There are features we need to actively avoid which require polyfills. For example, `Object.keys` and `Array.prototype.forEach` require polyfills so the code in utils.js has been refactored to avoid them. We also need to polyfill the Promise API since it doesn't exist in IE 11. The core-js polyfill for promises comes with a bunch of other stuff and is huge so I'm using [promise-polyfill](https://www.npmjs.com/package/promise-polyfill) instead which is much smaller and supports the ponyfill option to avoid manipulating the global namespace.

### New e2e tests to make this maintainable

Two new e2e tests have been added:
1. eslint is used to lint the bundle using the ES3 parser to ensure no new JS features sneak into the bundle. Babel-preset-env should polyfill them anyways but it's still nice for the extra smoke test.
2. the file size of dist/paypal-browser.min.js is measured to ensure it doesn't exceed the 5KB threshold. 

Ideally these tests would be run automatically to help keep the bundle size small and compatible with IE 11. These tests would need some more love if we decide to move forward with this approach.

### New Bundle Sizes

Existing sizes
existing-sizes

New Smaller Sizes
new-sizes

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -60,7 +60,7 @@ export function objectToQueryString(params) {
 // uses es3 to avoid requiring polyfills for Array.prototype.forEach and Object.keys
 function forEachObjectKey(obj, callback) {
     for (let key in obj) {
-        if (Object.prototype.hasOwnProperty.call(obj, key)) {
+        if (obj.hasOwnProperty(key)) {
             callback(key);
         }
     }


src/utils.js
```diff
@@ -22,17 +22,23 @@ export function insertScriptElement({ url, dataAttributes = {}, scriptAttributes
...
-            return accumulator;
-        }, { queryParams: {}, dataAttributes: {}, scriptAttributes: {} });
+    const processedOptions = {
+        queryParams: {},
+        dataAttributes: {},
+        scriptAttributes: {}
+    };
+
+    forEachObjectKey(options, key => {
+        if (key.substring(0, 5) === 'data-') {
```
mnicpt, line 32: Do you prefer `substring` over `indexOf`?  Seems `key.indexOf('data-') === 0` indicates more of a `startsWith` notion, where `substring` is more of a from here to there kind of thing.  Super small and just my opinion so either way is fine with me.

src/utils.js
```diff
@@ -42,10 +48,20 @@ export function processOptions(options = {}) {
...
+    });
+    return queryString;
+}
 
-            return accumulator += `${key}=${params[key]}`;
-        }, '');
+// uses es3 to avoid requiring polyfills for Array.prototype.forEach and Object.keys
+function forEachObjectKey(obj, callback) {
+    for (let key in obj) {
+        if (Object.prototype.hasOwnProperty.call(obj, key)) {
```
mnicpt: Why are you using the `call` function here rather than using the `obj` instance: `obj.hasOwnProperty(key)`?

src/utils.js
gregjopa: Unfortunately babel-preset-env polyfills `indexOf` because it thinks it may be for Array.prototype.indexOf() vs String.prototype.indexOf(). For example, here's the debug output showing how using indexOf introduces another polyfill:

index-of-polyfill-screenshot

src/utils.js
gregjopa: Good point. `obj.hasOwnProperty(key)` seems to be working fine when testing locally and it doesn't introduce any additional polyfill. I'll update it to use `obj.hasOwnProperty(key)` instead since it reads way better and is shorter.

PR #23, opened 2020-08-13
Add the ability to configure the async attribute

gregjopa:
This PR exposes the `async` script attribute and defaults it to true. The reason for this change is to:
- Make it super clear in the docs that we load the script asynchronously by default.
- Support the edge case where a user would want to set `async=false` to support sequential script loading.

## Details

Technically the script already loads async because that's the default behavior when adding a script to the DOM programmatically.

> Scripts that are dynamically created and added to the document are async by default, they don’t block rendering and execute as soon as they download, meaning they could come out in the wrong order. 

source: https://www.html5rocks.com/en/tutorials/speed/script-loading/

Setting async to false does change the behavior and can be used to load scripts in order programmatically:

> To request script-inserted external scripts be executed in the insertion order in browsers where the document.createElement("script").async evaluates to true (such as Firefox 4), set async="false" on the scripts you want to maintain order.

source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

Other script loading libraries seem to set async to true by default:
- https://github.com/myspace-nu/resLoader#options (this library also offers a `blocking` option which appends the script with `document.write()`. We shouldn't do this but just wanted to mention the approach)
- https://github.com/ded/script.js/blob/05860e2fcd07259da61319a1a30b2a9ad5f19d3a/src/script.js#L79

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ loadScript({ 'client-id': 'YOUR_CLIENT_ID', 'currency': 'EUR' });
 Which will render the following `<script>` tag to the DOM:
 
 ```html
-<script src="https://www.paypal.com/sdk/js?client-id=YOUR_CLIENT_ID&currency=EUR" defer></script>
+<script src="https://www.paypal.com/sdk/js?client-id=YOUR_CLIENT_ID&currency=EUR" async defer></script>
 ```
 
 View the [full list of supported query parameters](https://developer.paypal.com/docs/checkout/reference/customize-sdk/#query-parameters).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -42,7 +42,7 @@ loadScript({ 'client-id': 'YOUR_CLIENT_ID', 'currency': 'EUR' });
 Which will render the following `<script>` tag to the DOM:
 
 ```html
-<script src="https://www.paypal.com/sdk/js?client-id=YOUR_CLIENT_ID&currency=EUR" defer></script>
+<script src="https://www.paypal.com/sdk/js?client-id=YOUR_CLIENT_ID&currency=EUR" async defer></script>
```
mnicpt, line 45: If the default is to have both async and defer on the tag, then async will take precedence on modern browsers, and don't think that's what we want in the preload scenario.

gregjopa: Hmm that's a good point. That makes me wonder if `defer` is not working as expected on the main branch 🤔. If the script is loaded `async` by default when loading programmatically, then it may ignore the `defer` option in modern browsers. I'll put together a demo to see.

src/utils.js
```diff
@@ -18,6 +18,7 @@ export function insertScriptElement({ url, dataAttributes = {}, scriptAttributes
     document.head.insertBefore(newScript, document.head.firstElementChild);
 
     newScript.src = url;
+    newScript.async = scriptAttributes.async ?? true;
```
mnicpt, line 21: Need to ensure either one or the other and not allow both programmatically or need to let the devs know that if they specify both as true that it is the same as just async=true.  We really need to think about having them set both.  I think it is introducing more confusion.

PR #29, opened 2020-08-31
Document CDN Usage

gregjopa:
This PR replaces CDNX with [UNPKG](https://unpkg.com/) as our CDN solution.

### Why UNPKG instead of CDNX?

- UNPKG is a CDN for the public npm registry. With this design, we get CDN functionality for free since we already publish paypal-js to the public registry. We don't have to write any bash code to publish to the CDN like we would need to with cdnx.
- We get versioned urls!
- It seems stable and safe to recommend. Microsoft is using it as their recommended CDN for their new component library: https://www.fast.design/docs/components/getting-started#from-cdn

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # PayPal JS
 
-An async loader for the PayPal JS SDK.
+An async loader for the [PayPal JS SDK](https://developer.paypal.com/docs/checkout/).
 
 <a href="https://www.npmjs.com/package/@paypal/paypal-js"><img src="https://img.shields.io/npm/v/@paypal/paypal-js?style=flat-square" alt="npm version"></a>
 <a href="https://github.com/paypal/paypal-js/blob/main/LICENSE.txt"><img src="https://img.shields.io/npm/l/@paypal/paypal-js?style=flat-square" alt="github license"></a>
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -67,3 +67,27 @@ Which will load the following `<script>` asynchronously:
...
+
+### Using a CDN
+
+The paypal-js script is also available on the [UNPKG CDN](https://unpkg.com/). The paypal.browser.js build assigns the `loadScript` function to the window object as `window.paypalLoadScript`. Here's an example:
+
+```html
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <script src="https://unpkg.com/@paypal/paypal-js/dist/paypal.browser.min.js"></script>
```
gregjopa, line 79: Note the absence of the version number here. This is done on purpose so this code snippet always uses the latest version from UNPKG. There's redirect functionality provided by UNPKG when you leave out the version number. It redirects to the latest version.

Screen Shot 2020-08-31 at 11 30 24 AM

mnicpt: @gregjopa I like the simplicity of this approach and using what's already available to us.  How many edge servers to they have and where?  And can we configure a CNAME to map to this IP to customize the url?

gregjopa: > How many edge servers to they have and where?

I was wondering that too. The [About section on the website](https://unpkg.com/) mentions the unpkg CDN is powered by Cloudflare Enterprise and Cloudflare sponsor's the work. I found this talk from the author that goes into more detail: https://youtu.be/2rhkgB8Cohc?t=1737

> And can we configure a CNAME to map to this IP to customize the url?

I imagine we could but I'm unsure it would be worth the effort. The majority of developers will be using this module via npm and bundling it with their own assets.

mnicpt: Oh, right!  Good point!

PR #31, opened 2020-09-01
Document release process

gregjopa:
This PR creates a couple helper scripts for publishing a new release to npm. It also documents the release process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,11 @@
   "scripts": {
     "build": "rm -rf dist && rollup --config",
     "lint": "eslint .",
-    "prepublishOnly": "npm run lint && npm test && npm run build && npm run test:bundle && npm run test:e2e",
+    "prepublishOnly": "echo \"\nPlease use 'npm run release' instead\n\" && exit 1",
+    "release": "node ./scripts/publish",
+    "release:patch": "node ./scripts/publish patch",
+    "release:minor": "node ./scripts/publish minor",
+    "release:major": "node ./scripts/publish major",
     "test": "jest --env=jsdom src",
     "test:bundle": "jest --env=jsdom e2e-tests/bundle/**",
     "test:e2e": "jest -c e2e-tests/jest.config.js",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -7,7 +7,11 @@
   "scripts": {
     "build": "rm -rf dist && rollup --config",
     "lint": "eslint .",
-    "prepublishOnly": "npm run lint && npm test && npm run build && npm run test:bundle && npm run test:e2e",
+    "prepublishOnly": "echo \"\nPlease use 'npm run release' instead\n\" && exit 1",
```
mnicpt, line 10: Are you doing this for backward compatibility? Wondering why we need this.

scripts/publish.js
```diff
@@ -0,0 +1,23 @@
...
+exec('npm run lint && npm test && npm run build');
+
+// ensure bundle size is below threshold
+exec('npm run test:bundle');
+
+// run puppeteer tests
+exec('npm run test:e2e');
+
+// bump version
+const newVersionArg = process.argv[2] || 'patch';
```
mnicpt, line 16: Are we unable to get the current version and automatically bump it?

package.json
gregjopa: I added this incase someone tries to run `npm publish` in this repo. This will prevent that operation and recommend `npm run release` instead.

scripts/publish.js
gregjopa: This is here to support what type of version bump to do. Similar to how [npm version](https://docs.npmjs.com/cli/version) takes in patch, minor, or major.

package.json
mnicpt: What catches someone trying to run `publish` and then fires off `npm run prepublishOnly`?

gregjopa: Ah yes, that's built into the npm publish command workflow itself. Here's the definition of prepublishOnly:

> prepublishOnly: Run BEFORE the package is prepared and packed, ONLY on npm publish

https://docs.npmjs.com/misc/scripts

mnicpt: Ah, figured it might be something like that.  Thanks for clarifying.

PR #34, opened 2020-09-14
Improve error messaging when the script fails to load

gregjopa:
@mnicpt and I were hacking on the error use case this morning. This PR improves the error messages you see in the Dev Tools Console when paypal-js fails to load the script. 

For example, if a bad query param is passed in the developer will see the following error in the console:
Screen Shot 2020-09-14 at 1 58 41 PM

### Additional Details

The JS SDK does return useful error messages. The problem is you have to check the Network Tab in Dev Tools to see that response. The error doesn't show up in the console because `<script src="....">` only executes when a 200 status code is returned. This `fetch` when there's an error approach allows us to get at the response status code and response body so we can log it to the console. This way the developer is more likely to see the error.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -27,11 +27,31 @@ export function loadScript(options) {
             onSuccess: () => {
                 isLoading = false;
                 if (window.paypal) return resolve(window.paypal);
-                return reject(new Error('The window.paypal global variable is not available.'));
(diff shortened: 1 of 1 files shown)


gregjopa:
I'm going to close this one out and re-think error messaging.

PR #36, opened 2020-09-24
Update cache logic to check data attributes

gregjopa:
This PR updates the cache logic for the reload use case. This change ensures that both query params and data attributes are validated for the reload use case. 
- If the query params or data attributes change, it will reload the script. 
- If both are the same, it will avoid reloading the script and will return the existing `window.paypal` reference.

### DIFF
--- a/e2e-tests/bundle/bundle-size.test.js
+++ b/e2e-tests/bundle/bundle-size.test.js
@@ -1,12 +1,14 @@
 import filesize from 'filesize';
 import fs from 'fs';
 
+const maxBundleSizeInKiloBytes = 6;
+
 describe('bundle size', () => {
-    it('paypal.browser.min.js should be less than 5 KB', () => {
+    it(`paypal.browser.min.js should be less than ${maxBundleSizeInKiloBytes} KB`, () => {
         const { size: sizeInBytes } = fs.statSync('dist/paypal.browser.min.js');
         const [sizeInKiloBytes, label] = filesize(sizeInBytes, {output: "array"});
         console.log(`paypal.browser.min.js: ${sizeInKiloBytes} ${label}`);
 
-        expect(sizeInKiloBytes).toBeLessThan(18);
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -55,3 +68,14 @@ function forEachObjectKey(obj, callback) {
         }
     }
 }
+
+function createScriptElement(url, dataAttributes = {}) {
```
gregjopa, line 64: Note that this new function creates the script but does not insert it into the DOM. So no network request is triggered by this function. It's reused twice above and makes it easier to compare data attributes in `findScript()`

src/utils.js
```diff
@@ -1,17 +1,31 @@
-export function findScript(url) {
-    return document.querySelector(`script[src="${url}"]`);
+export function findScript(url, dataAttributes) {
+    const currentScript = document.querySelector(`script[src="${url}"]`);
+    if (!currentScript) return null;
+
+    const nextScript = createScriptElement(url, dataAttributes);
+
+    // check if the new script has the same number of data attributes
+    if (objectSize(currentScript.dataset) !== objectSize(nextScript.dataset)) {
```
gregjopa, line 8: @mnicpt thanks for pairing on this. I added the `objectSize` helper function and some comments to improve readability while still keeping it ES3. Please give this another review.

e2e-tests/bundle/bundle-size.test.js
```diff
@@ -1,12 +1,14 @@
 import filesize from 'filesize';
 import fs from 'fs';
 
+const maxBundleSizeInKiloBytes = 6;
+
 describe('bundle size', () => {
-    it('paypal.browser.min.js should be less than 5 KB', () => {
+    it(`paypal.browser.min.js should be less than ${maxBundleSizeInKiloBytes} KB`, () => {
```
gregjopa, line 7: I noticed this e2e test was wrong. It said it was checking for a 5 KB threshold but actually checked for 18 KB 🤭.

The actual minified size is 5.12 KB so I set the threshold to 6 KB for some breathing room.

PR #43, opened 2020-10-16
Update publish script

elizabethmv:
## What
Updates the`publish.js` script to use a callback to pass in the OTP flag with `npm publish`.

## Why
Attempting to run `npm publish` with the `exec` command from `shelljs` with 2FA enabled exits before allowing the user to enter an OTP.


gregjopa (approved):
💯

PR #48, opened 2020-11-23
Bring your own Promise Polyfill

gregjopa:
## The Problem

Currently paypal-js includes a promise-polyfill. This polyfill doubles the size of the library. More than 95% of users have browsers that support Promises (source: https://caniuse.com/promises) These users shouldn't have to pay a performance penalty when this polyfill when it's not needed. 

## Solution

Don't ship with a Promise Polyfill. Instead let's provide a way for users to bring their own Promise implementation to support legacy browsers like IE 11. This change makes the paypal-js over 50% smaller.

### Before
Screen Shot 2020-11-23 at 1 07 46 PM

### After
Screen Shot 2020-11-23 at 1 06 56 PM

See the readme changes for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# PayPal JS
+# paypal-js
 
 > A client-side loader for the [PayPal JS SDK](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-reference/)
 
@@ -82,6 +82,29 @@ Which will load the following `<script>` asynchronously:
 
 View the [full list of supported script parameters](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-configuration/#script-parameters).
 
+### Legacy Browser Support
+
+This library relies on [JavaScript Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). To support legacy browsers like IE 11, you must provide your own Promise polyfill. With a Promise polyfill this library will [support the same browsers as the JS SDK](https://developer.paypal.com/docs/business/checkout/reference/browser-support/#supported-browsers-by-platform).
+
+The `loadScript()` function takes in a second parameter for providing a Promise ponyfill. It defaults to the global `Promise` object if it exists. There are two options for polyfilling the Promise object:
+
+1. Use a global polyfill strategy that monkey patches the `window.Promise` API implementation.
+2. Use a [ponyfill strategy](https://github.com/sindresorhus/ponyfill) that passes a Promise library into `loadScript()` without affecting other code:
+
+```js
+import { loadScript } from '@paypal/paypal-js';
+import PromisePonyfill from 'promise-polyfill';
+
+loadScript(options, PromisePonyfill)
+    .then(paypalObject => {})
+```
+
+We also provide a legacy build that includes the [promise-polyfill](https://github.com/taylorhakes/promise-polyfill) library. You can reference it from the CDN here:
+
+```html
+<script src="https://unpkg.com/@paypal/paypal-js/dist/paypal.legacy.browser.min.js"></script>
+```
+
 ### Using a CDN
 
 The paypal-js script is also available on the [unpkg CDN](https://unpkg.com/). The paypal.browser.js build assigns the `loadScript` function to the window object as `window.paypalLoadScript`. Here's an example:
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -82,6 +83,29 @@ Which will load the following `<script>` asynchronously:
 
 View the [full list of supported script parameters](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-configuration/#script-parameters).
 
+### Legacy Browser Support
+
+This library relies on [JavaScript Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). To support legacy browsers like IE 11, you must provide your own Promise polyfill. With a Promise polyfill this library will support the same browsers as the JS SDK. Here's the [full list of supported browsers](https://developer.paypal.com/docs/business/checkout/reference/browser-support/#supported-browsers-by-platform).
+
+The `loadScript()` function takes in a second parameter for providing a Promise ponyfill. It defaults to the global `Promise` object if it exists. There are two options for polyfilling the Promise object:
```
mnicpt, line 90: Might want to provide a link to `ponyfill` here.  That through me for a loop.  Thought you initially misspelled it.

README.md
```diff
@@ -82,6 +83,29 @@ Which will load the following `<script>` asynchronously:
...
+
+1. Use a global polyfill strategy that sets `window.Promise` when it doesn't exist.
+2. Use a [ponyfill strategy](https://github.com/sindresorhus/ponyfill) that passes a Promise library into `loadScript()`:
```
mnicpt, line 93: `polyfill`

README.md
```diff
@@ -82,6 +83,29 @@ Which will load the following `<script>` asynchronously:
 
 View the [full list of supported script parameters](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-configuration/#script-parameters).
 
+### Legacy Browser Support
+
+This library relies on [JavaScript Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). To support legacy browsers like IE 11, you must provide your own Promise polyfill. With a Promise polyfill this library will support the same browsers as the JS SDK. Here's the [full list of supported browsers](https://developer.paypal.com/docs/business/checkout/reference/browser-support/#supported-browsers-by-platform).
```
BriceShatzer: Nit-picking, but I think it could be made clearer that the "full list of supported browsers" that is provided is explicitly referring to the list of browsers that the full JS SDK supports. Maybe not having that link live in separate sentence? 
>This library relies on [JavaScript Promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise). To support legacy browsers like IE 11, you must provide your own Promise polyfill. With a Promise polyfill this library will [support the same browsers](https://developer.paypal.com/docs/business/checkout/reference/browser-support/#supported-browsers-by-platform) as the JS SDK.

mnicpt (approved):

BriceShatzer (changes requested):
not sure if you've thought about the versioning with this, but if you're following [semver](https://semver.org/) I think this probably counts as a major version in terms of backwards compatible changes. 

If updating `package.json` to 2.x messes with anything from a deployment/cdn/etc. perspective, it might be worth looking into usage data to see if there's a potential to fake this promise behavior is as always being a part of 1.x

README.md
gregjopa: @BriceShatzer thanks for the suggestion. That is definitely better. I just pushed up the change.

README.md
gregjopa: I agree that ponyfill is confusing at first. I updated the description and it includes the link to https://github.com/sindresorhus/ponyfill

gregjopa:
> if you're following semver I think this probably counts as a major version in terms of backwards compatible changes.

I agree. This will definitely be a major release since it's not backwards compatible. I think this will be rolled out with v2.

> If updating package.json to 2.x messes with anything from a deployment/cdn/etc. perspective, it might be worth looking into usage data to see if there's a potential to fake this promise behavior is as always being a part of 1.x

This major version bump will not break anything with our CDN strategy. We use unpkg for the CDN functionality which versions urls like so: https://unpkg.com/@paypal/paypal-js@1.0.4/dist/paypal.browser.min.js. Folks can technically always link to the latest on unpkg by not including the version number like so (https://unpkg.com/@paypal/paypal-js/dist/paypal.browser.min.js). But that's not recommended and we have a note about that in our README

BriceShatzer (approved):

PR #55, opened 2020-12-21
Use TypeScript and also manage type declarations for the JS SDK

gregjopa:
This PR does two things:
- Upgrades the library to use TypeScript
- Exports type declarations for both paypal-js and the `window.paypal` namespace

### How to Test the new Types

This branch is currently published to npm using the beta tag. Install it using the following command:

`npm install @paypal/paypal-js@beta`

Here are some screenshots of my editor to show the new helpful auto-suggestions TypeScript devs will get:

Types for `loadScript()`:
Screen Shot 2020-12-29 at 3 25 41 PM

Types for `window.paypal`:
Screen Shot 2020-12-29 at 3 26 09 PM

Types for `window.paypal` reference returned from loadScript Promise:
Screen Shot 2020-12-29 at 3 25 26 PM

### Why convert to TypeScript

There is a high demand for TypeScript types for the JS SDK. For example, the checkout.js library is typed on DefinitelyTyped and gets about 13K downloads a week (https://www.npmjs.com/package/@types/paypal-checkout-components). Let's host the JS SDK types in this repo and be the source of truth for those types. It's also an incentive for more folks to use this library.

#### TypeScript Definition files

When designing a library with TypeScript, you have two options:
1. Use `declarations: true` in your tsconfig.json to dynamically generate them from your source code.
2. Manually define them using definition files.

My goal was to use option 1 with auto-generating the types for `loadScript()` and also the `window.paypal` namespace that isn't managed in this code base. I ran into problems with this approach with the TypeScript complier being smart about only including declarations for functionality that is actually exports. This problem led me to use option 2 instead (I think this is why [stripe-js](https://github.com/stripe/stripe-js/blob/master/package.json#L8) also uses option 2). There is some redundancy between what's in the /types/ folder and the /src/ folder but I think it's worth it. The benefit to this design is we can add TypeScript definitions for all our components (`window.paypal.Buttons`, `window.paypal.Messages`, etc...) and manage them in this single repo.

There is a lot more work to do for adding type declarations for `window.paypal`. Right now only the `Buttons` component is fully defined. The goal of this PR is to lay the foundation for our TypeScript strategy for the PayPal JS SDK. Future PRs can add types for things like `window.paypal.Marks`, `window.paypal.Messages`, braintree integration support, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Client-side loader for the PayPal JS SDK",
   "main": "dist/paypal.node.js",
   "module": "dist/paypal.esm.js",
+  "types": "types/index.d.ts",
   "scripts": {
     "build": "rm -rf dist && rollup --config",
     "lint": "eslint .",
@@ -15,10 +16,13 @@
     "test:bundle": "jest --env=jsdom e2e-tests/bundle/**",
     "test:e2e": "jest -c e2e-tests/jest.config.js",
     "test:e2e:start": "serve",
-    "validate": "npm run build && npm run lint && npm test -- --coverage && npm run test:bundle"
+    "typecheck": "tsc --noEmit && tsc types/*.ts --noEmit",
+    "validate": "npm run typecheck && npm run build && npm run lint && npm test -- --coverage && npm run test:bundle"
   },
   "files": [
-    "dist"
+    "dist",
+    "types",
+    "!types/**/*.test.ts"
   ],
   "keywords": [
     "paypal",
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -15,7 +16,8 @@
     "test:bundle": "jest --env=jsdom e2e-tests/bundle/**",
     "test:e2e": "jest -c e2e-tests/jest.config.js",
     "test:e2e:start": "serve",
-    "validate": "npm run build && npm run lint && npm test -- --coverage"
+    "typecheck": "tsc",
```
camsjams: Can update this to include `noEmit`:
`"typecheck": "tsc --noEmit"`

Since its just a quality check, no need to emit an output.

See https://www.typescriptlang.org/docs/handbook/compiler-options.html

types/components/buttons.d.ts
```diff
@@ -0,0 +1,6 @@
+export interface ButtonsComponentSettings {
+    createOrder?: (data: object, actions: object) => void;
```
camsjams: Should be able to enhance this with the Types from https://gist.github.com/camsjams/b843637df7a30c260b341daf97a96d60#file-paypal-react-paypal-js-d-ts-L147

camsjams (commented):
Looks like a great pass - thanks for making these! 

I've been pulled to another project so have minimal availability at this time of year.

:tada:

gregjopa:
Thanks for the code review @camsjams! The [types](https://gist.github.com/camsjams/b843637df7a30c260b341daf97a96d60#file-paypal-react-paypal-js-d-ts-L147) you created are a great starting point. 💯

pedroapfilho (approved):
Looks very good. The only thing that would keep me from approving this PR is the amount of `@ts-expect-error` tags, but I've checked and they don't compromise the type safety of the application itself, so this is good to go.

gregjopa:
> The only thing that would keep me from approving this PR is the amount of @ts-expect-error tags, but I've checked and they don't compromise the type safety of the application itself, so this is good to go

@pedroapfilho thank you for the code review! I refactored the tests and now we only have 3 instances using  "ts-expect-error". I agree that these "ts-expect-error" compiler exceptions should be used sparingly. They are only in the test file and we can always fix them later after better learning how to use TypeScript w/ HJest.

PR #66, opened 2021-03-24
Added HostedFields to interface PayPalNamespace

getTobiasNielsen:
HostedFields will be required for <PayPalHostedFields>

Issue: https://github.com/paypal/react-paypal-js/issues/49

This addition also depends on two types, that will need to be created.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -15,6 +15,7 @@ import type { getFundingSources } from "./components/funding-eligibility";
 
 export interface PayPalNamespace {
     Buttons?: (options?: PayPalButtonsComponentProps) => PayPalButtonsComponent;
+    HostedFields?: (options?: PayPalButtonsComponentProps) => PayPalButtonsComponent;
     Marks?: (options?: PayPalMarksComponentProps) => PayPalMarksComponent;
     Messages?: (
         options?: PayPalMessagesComponentProps


types/index.d.ts
```diff
@@ -15,6 +15,7 @@ import type { getFundingSources } from "./components/funding-eligibility";
 
 export interface PayPalNamespace {
     Buttons?: (options?: PayPalButtonsComponentProps) => PayPalButtonsComponent;
+    HostedFields?: (options?: PayPalButtonsComponentProps) => PayPalButtonsComponent;
```
mnicpt, line 18: HostedFields returns a PayPalButtonsComponent and takes in PayPalButtonsComponentProps?

mnicpt: Seeing now that you are having a conversation with @gregjopa about this in `react-paypal-js`.

gregjopa: Let's create a new `PayPalHostedFieldsComponent` since this is a different interface than Buttons.
```suggestion
    HostedFields?: PayPalHostedFieldsComponent;
```
hosted-fields-screenshot

I'm thinking we can create a new file here: types/components/hosted-fields.d.ts. And it will contain the following:
```ts
export interface PayPalHostedFieldsComponent {
    isEligible: () => boolean;
    render: (container: HTMLElement | string) => Promise<void>;
}
```

gregjopa:
Thanks for the PR @getTobiasNielsen! It's great to have more contributors. I left a comment with feedback.

types/index.d.ts
getTobiasNielsen: How did it change those to Buttons, I swear I changed it to 
`HostedFields?: (options?: PayPalHostedFieldsComponentProps) => PayPalHostedFieldsComponent;`

It's not a function that returns `PayPalHostedFieldsComponent` like buttons?

gregjopa: > It's not a function that returns PayPalHostedFieldsComponent like buttons?

Unfortunately the hosted-fields component does not follow the same pattern. Instead it returns an object. The source code for this components lives here: https://github.com/paypal/paypal-card-components/blob/6d099d7157d1ba26cfe2f2e0858c6ec24450a20d/src/component.js#L152-L168

getTobiasNielsen: I see.

gregjopa:
@getTobiasNielsen thanks for getting this hosted-fields work started. I added some initial types in #72 so let's close this one out.

PR #76, opened 2021-04-21
fix: override tsconfig target in legacy build

dali546:
The legacy paypal-js library, which supposedly should work on IE11, does not compile to ES5.
This overrides the `target: "esnext"` in `tsconfig.json`


gregjopa:
Great catch @dali546! 👏  Could you add this `target: 'es5'` change to the overrides section here? https://github.com/paypal/paypal-js/blob/main/rollup.config.js#L20. We want to output es5 for all dist files.

gregjopa (approved):

gregjopa:
Thanks @dali546! I'll merge this and publish a bugfix.

PR #122, opened 2021-09-27
docs: update docs to use async/await syntax

wsbrunson:
updated README to include Async/Await examples. I left the original examples and labelled them as Promises. Open to removing them, changing the labels, or anything else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,22 @@ Import the `loadScript` function for asynchronously loading the Paypal JS SDK.
 -   accepts an object for passing query parameters and attributes to the JS SDK.
 -   returns a Promise that resolves with `window.paypal` after the JS SDK is finished loading.
 
+#### Async/Await
+
+```js
+import { loadScript } from "@paypal/paypal-js";
+
+try {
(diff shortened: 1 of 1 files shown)


westeezy (approved):
good idea showing both formats. lgtm

README.md
```diff
@@ -56,6 +56,20 @@ Import the `loadScript` function for asynchronously loading the Paypal JS SDK.
 -   accepts an object for passing query parameters and attributes to the JS SDK.
 -   returns a Promise that resolves with `window.paypal` after the JS SDK is finished loading.
 
+#### Async/Await
+```js
```
gregjopa, line 61: Awesome first PR @wsbrunson! 💯 🎸

Can you run `npm run format` locally? I think there are some minor formatting tweaks that prettier will apply (I thought this would get auto-formatted by the precommit hook but did not). Please push that up and then this should be good to merge!

wsbrunson: I hadn't run `npm install` yet so it committed without running any of the pre-commit hooks. Should be good now!

PR #142, opened 2021-12-28
chore(types): refactor tests for types

gregjopa:
This PR includes a few improvements to the type tests:
- Replaces the single test file `types/index.d.test.ts ` with multiple files, one per component.
- Uses async/await instead of Promises. I couldn't figure out how to get top level async/await working so these test files all wrap the code with `async function main()`.
- Leverages optional chaining for null checks


package.json
```diff
@@ -20,7 +20,7 @@
         "test:bundle": "jest --env=jsdom e2e-tests/bundle/**",
         "test:e2e": "jest -c e2e-tests/jest.config.js",
         "test:e2e:start": "serve",
-        "typecheck": "tsc --noEmit && tsc types/*.ts --noEmit",
```
gregjopa, line 23: The `types` folder is now included in the tsconfig: https://github.com/paypal/paypal-js/blob/main/tsconfig.json#L15

PR #208, opened 2022-03-25
Replace babel with TypeScript for transpiling to ES5

gregjopa:
This library is a great candidate for using TypeScript instead of babel since it doesn't rely on polyfills. 

This PR includes the following changes:
- removes all of the babel configuration code and uses TypeScript to transpile to ES5. 
- uses ts-jest for using TypeScript with the test suite.
- refactors the configuration of the puppeteer e2e test suite to play nice without babel.


e2e-tests/loading-errors/index.test.js
```diff
@@ -1,10 +1,11 @@
-import { baseURL } from "../test-helper";
-
 describe("Validation Errors", () => {
     beforeEach(async () => {
-        await page.goto(`${baseURL}/e2e-tests/loading-errors/index.html`, {
-            waitUntil: "networkidle2",
-        });
+        await page.goto(
+            `http://localhost:${process.env.PORT}/e2e-tests/loading-errors/index.html`,
```
westeezy: Should this use `process.env.BASE_URL`?

gregjopa: `process.env.BASE_URL` does make more sense. I'll update it.

package.json
```diff
@@ -17,7 +17,7 @@
         "postrelease": "git push && git push --follow-tags && npm run build && npm publish",
         "test": "jest --env=jsdom src",
         "test:bundle": "jest --env=jsdom e2e-tests/bundle/**",
-        "test:e2e": "jest -c e2e-tests/jest.config.js",
+        "test:e2e": "PORT=4444 jest -c e2e-tests/jest.config.js",
```
westeezy: Should the port be removed here so users can set?

gregjopa: This has been updated to the following:
```
"test:e2e": "BASE_URL=http://localhost:4444 jest -c e2e-tests/jest.config.js",
```

It is missing the ability for users to be able to override. But I don't think anyone ever used that feature. Let's drop it for simplicity. The real e2e tests live here and are much more configurable: https://github.com/paypal/paypal-sdk-e2e-tests

westeezy (approved):
lgtm!

wsbrunson (commented):
nice!

PR #258, opened 2022-06-30
feat(types): type out funding sources

gregjopa:
This PR adds types for all the possible funding source values. These values were copied over from the sdk-constants repo: https://github.com/paypal/paypal-sdk-constants/blob/main/src/funding.js#L3.

Also, the `createOrder(data, actions)` callback now has `data.paymentSource` for letting the merchant know which button was clicked. I added types for that.


westeezy (approved):
:chef-kiss:

PR #268, opened 2022-07-25
Update orders.d.ts

TuomoKonu:
As you are stating in your comment " * Either type or options but not both may be present".
Currently Paypal doesn't allow type and options to be used at the same time, but types it paypal-js are enforcing developers to use both.
I want to use paypal-js types over defining empty type declaration in my local types folder. so i'm suggesting this fix to paypal-js types.

### DIFF
--- a/types/apis/orders.d.ts
+++ b/types/apis/orders.d.ts
@@ -158,12 +158,12 @@ export type ShippingInfo = {
      * The method by which the payer wants to get their items from the payee e.g shipping, in-person pickup.
      * Either type or options but not both may be present
      */
-    type: string;
+    type?: string;
     /**
      * An array of shipping options that the payee or
      * merchant offers to the payer to ship or pick up their items
      */
-    options: ShippingInfoOption[];
+    options?: ShippingInfoOption[];
     address: Address;
 };
 


wsbrunson:
@TuomoKonu I think the type should actually look like this:

```
export type ShippingInfoBase = {
    /**
     * The name of the party
     */
    name: Partial<{
        /**
         * When the party is a person, the party's full name
         */
        full_name: string;
    }>;
    email_address: string;
    phone_number: {
        /**
         * The national number, in its canonical international [E.164 numbering plan format](https://www.itu.int/rec/T-REC-E.164/en).
         * The combined length of the country calling code (CC) and the national number must not be greater than 15 digits.
         * The national number consists of a national destination code (NDC) and subscriber number (SN).
         */
        national_number: string;
    };
    address: Address;
};

type ShippingInfoWithType = {
    ...ShippingInfoBase,
    type: string;
}

type ShippingInfoWithOptions = {
    ...ShippingInfoBase,
    options: ShippingInfoOption[];
}

export type ShippingInfo = ShippingInfoWithType | ShippingInfoWithOptions;
```

Then Typescript will know that `type` or `options` is required but won't let you use both

TuomoKonu:
> @TuomoKonu I think the type should actually look like this:
> 
> ```
> export type ShippingInfoBase = {
>     /**
>      * The name of the party
>      */
>     name: Partial<{
>         /**
>          * When the party is a person, the party's full name
>          */
>         full_name: string;
>     }>;
>     email_address: string;
>     phone_number: {
>         /**
>          * The national number, in its canonical international [E.164 numbering plan format](https://www.itu.int/rec/T-REC-E.164/en).
>          * The combined length of the country calling code (CC) and the national number must not be greater than 15 digits.
>          * The national number consists of a national destination code (NDC) and subscriber number (SN).
>          */
>         national_number: string;
>     };
>     address: Address;
> };
> 
> type ShippingInfoWithType = {
>     ...ShippingInfoBase,
>     type: string;
> }
> 
> type ShippingInfoWithOptions = {
>     ...ShippingInfoBase,
>     options: ShippingInfoOption[];
> }
> 
> export type ShippingInfo = ShippingInfoWithType | ShippingInfoWithOptions;
> ```
> 
> Then Typescript will know that `type` or `options` is required but won't let you use both

Hi [wsbrunson](https://github.com/wsbrunson), I think your suggestion would do, if TypeScript would not enforce us to use both at the same time. Then we should add this change instead in this PR.

wsbrunson:
> Hi [wsbrunson](https://github.com/wsbrunson), I think your suggestion would do, if TypeScript would not enforce us to use both at the same time as it is currently doing and for this reason my build fails in Github actions pipeline becaues of this current enforcement.. Then we should add this change instead in this PR.

Sounds good! If you are interested in making the change, I'll review it and merge

gregjopa:
Hi @TuomoKonu thanks for raising this PR. This should be solved in the latest release ([5.1.1](https://github.com/paypal/paypal-js/commit/f464bb6f0c43e0b03e0bb21a364a2b54a2a2e1ab)). Please give it a try. If you still run into type errors with the latest release we can reopen the issue.

TuomoKonu:
> Hi @TuomoKonu thanks for raising this PR. This should be solved in the latest release ([5.1.1](https://github.com/paypal/paypal-js/commit/f464bb6f0c43e0b03e0bb21a364a2b54a2a2e1ab)). Please give it a try. If you still run into type errors with the latest release we can reopen the issue.

@gregjopa Hi, thank you for your reply. I will try it right away. I was on summer holiday. Tought to contribute this fix now. but you made it. :)

Issue #269, opened 2022-07-25
[Bug] PayPalMessagesComponentOptions not up to date


CoderBrad:
### 🐞 Describe the Bug

The current PayPalMessagesComponentOptions does not match the documentation found [in the sdk docs](https://developer.paypal.com/docs/checkout/pay-later/us/integrate/reference/#link-messageconfigurationobject).

### 🔬 Minimal Reproduction

Compare current type to the docs.

### 😕 Actual Behavior

Current type is missing options mentioned in the docs.

### 🤔 Expected Behavior

To be able to reference options present in the docs.

### 🌍 Environment

N/A

gregjopa:
This was resolved in #274

PR #274, opened 2022-07-29
Update messages type to match current documentation

CoderBrad:
I have added the missing options and removed account since it is not listed in the docs.
https://github.com/paypal/paypal-js/issues/269

### DIFF
--- a/types/components/messages.d.ts
+++ b/types/components/messages.d.ts
@@ -16,9 +16,9 @@ export interface PayPalMessagesComponentOptions {
         ratio?: "1x1" | "1x4" | "8x1" | "20x1";
     };
     placement?: "home" | "category" | "product" | "cart" | "payment";
-    onApply?: Function;
-    onClick?: Function;
-    onRender?: Function;
+    onApply?: (data: Record<string, unknown>) => void;
+    onClick?: (data: Record<string, unknown>) => void;
+    onRender?: (data: Record<string, unknown>) => void;
 }
 
 export interface PayPalMessagesComponent {
--- a/types/components/messages.d.ts
+++ b/types/components/messages.d.ts
@@ -1,6 +1,6 @@
 export interface PayPalMessagesComponentOptions {
-    account?: string;
     amount?: number | string;
+    currency?: "USD" | "GBP" | "EUR";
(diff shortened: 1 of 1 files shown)


types/components/messages.d.ts
```diff
@@ -13,7 +13,12 @@ export interface PayPalMessagesComponentOptions {
             size?: 10 | 11 | 12 | 13 | 14 | 15 | 16;
             align?: "left" | "center" | "right";
         };
+        ratio?: "1x1" | "1x4" | "8x1" | "20x1";
     };
+    placement?: "home" | "category" | "product" | "cart" | "payment";
+    onApply?: Function;
```
gregjopa: Let's avoid using `Function` and instead type these out as functions accept an object as the arg and return nothing.
```suggestion
    onApply?: (data: Record<string, unknown>) => void;
```

types/components/messages.d.ts
```diff
@@ -13,7 +13,12 @@ export interface PayPalMessagesComponentOptions {
...
+    onClick?: Function;
```
gregjopa:
```suggestion
    onClick?: (data: Record<string, unknown>) => void;
```

types/components/messages.d.ts
```diff
@@ -13,7 +13,12 @@ export interface PayPalMessagesComponentOptions {
...
+    onRender?: Function;
```
gregjopa:
```suggestion
    onRender?: (data: Record<string, unknown>) => void;
```

gregjopa: When testing locally I noticed that these callbacks do have data that's passed in. I don't think we need to type it all out at this point but it's good to call it out with `data: Record<string, unknown>`. Ex:
Screen Shot 2022-07-29 at 9 50 49 AM

CoderBrad: Yea, that was my delay in accepting. I was noticing paypal-messaging-components tests seems to have some data being passed in but tracing what and where it goes was getting a little confusing.

gregjopa (approved):

gregjopa:
Thanks for your contribution @CoderBrad! 💯

PR #276, opened 2022-08-02
fix(types): update order shipping options to match spec

gregjopa:
This PR is a follow up to https://github.com/paypal/paypal-js/pull/268 and includes the following changes:

- Update shipping name, address, phone_number, and email to be optional (https://developer.paypal.com/docs/api/orders/v2/#definition-shipping_detail)
- Update shipping types to enforce the rule of having either` shipping.type` or `'shipping.options` but not both. This enforces the following requirement from the rest api docs:
  > The method by which the payer wants to get their items from the payee e.g shipping, in-person pickup. Either type or options but not both may be present.

### DIFF
--- a/types/apis/orders.d.ts
+++ b/types/apis/orders.d.ts
@@ -135,7 +135,7 @@ export type ShippingInfoOption = {
     selected: boolean;
 };
 
-export interface ShippingInfoBase {
+interface ShippingInfoBase {
     /**
      * The name of the party
      */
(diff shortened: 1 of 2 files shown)


types/apis/orders.d.ts
```diff
@@ -135,37 +135,39 @@ export type ShippingInfoOption = {
     selected: boolean;
 };
 
-export type ShippingInfo = {
+export interface ShippingInfoBase {
```
wsbrunson: we could stop exporting this type in favor of just `ShippingInfo`

wsbrunson (approved):

types/apis/orders.d.ts
gregjopa: Great point! Updated.

westeezy (approved):

PR #305, opened 2022-11-11
chore(tests): replace puppeteer with playwright

gregjopa:
This PR upgrades the e2e test suite to use [Playwright](https://playwright.dev/). It also includes a GitHub Action for running the e2e tests.


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,51 @@
...
+              uses: styfle/cancel-workflow-action@0.11.0
+              with:
+                  access_token: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: ⬇️ Checkout repo
+              uses: actions/checkout@v3
+              with:
+                  fetch-depth: 0
+
+            - name: 🤝 Set Node version from .nvmrc
```
westeezy, line 24: We should probably use this in more of our workflows since we always have the .nvmrc file anyways!

wsbrunson (approved):
very cool, i hear that playwright is the way to go

.github/workflows/playwright.yml
gregjopa: Yeah it's great having our Node.js version defined in one place (.nvmrc) and having all other things like GitHub Actions read from it.

PR #321, opened 2022-12-02
fix: update load-script.js

dtjones404:
Updated environment checks to check for `document` rather than `window` to make code Deno-safe, as Deno has a [window](https://deno.land/api@v1.28.3?s=Window) object.


wsbrunson (approved):

gregjopa:
Awesome work @dtjones404! 💯

PR #326, opened 2022-12-19
fix: update github actions badge url

dtjones404:


gregjopa:
@dtjones404 can you update the commit to be chore(docs) instead of fix? That will be better for the changelog behavior with semver. Please do the same for react-paypal-js

PR #347, opened 2023-03-21
fix(types): add type for minimal response from orders api

### LINKED ISSUE #337: [Types] Add support for different preferred server responses
### Improving Types

The [PayPal v2 Orders API](https://developer.paypal.com/docs/api/orders/v2/#orders-create-header-parameters) supports passing a request header to define the preferred server response shape. This is the `Prefer` request header and it support two types of values:

1. `return=minimal`
2. `return=representation`

For example, the v2 Orders API create operation defaults to returning a minimal response. Where as the v2 Orders API capture operation defaults to returning the large representation response.

The solution is to support both of these types of responses. Currently we just support one: https://github.com/paypal/paypal-js/blob/main/types/apis/orders.d.ts#L241.

One thought is to export two different types so the merchant can choose which on they want. Ex:
```typescript
// new type for return=minimal
export type OrderResponseBodyMinimal = {}
// use existing type for return=representation 
export type OrderResponseBody = {}
```

You can use this example repo to test out the different responses from the orders api and play around with the types: https://github.com/paypal-examples/paypal-sdk-server-side-integration/blob/main/src/order/create-order.ts#L43-L59

gregjopa:
Resolves #337

### DIFF
--- a/types/apis/orders.d.ts
+++ b/types/apis/orders.d.ts
@@ -219,9 +219,6 @@ export type LinkDescription = {
     href: string;
     rel: string;
     method?: string;
-    title?: string;
-    mediaType?: string;
-    encType?: string;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


wsbrunson (approved):

xuanwan05 (approved):
LG2M

za8457 (commented):
#172

PR #360, opened 2023-04-11
fix(types): update patch request body for order actions

jocmp:
Updates the "patch" function for button callbacks to be consistent with the documented behavior for v2 of the PayPal orders API.

Reference:

- https://developer.paypal.com/docs/api/orders/v2/#orders_patch
- https://developer.paypal.com/docs/checkout/advanced/customize/update-order-details/#link-updateorderdetails

### DIFF
--- a/types/apis/orders.d.ts
+++ b/types/apis/orders.d.ts
@@ -235,6 +235,31 @@ export type CreateOrderRequestBody = {
     application_context?: OrderApplicationContext;
 };
 
+export type UpdateOrderOperation =
+    | {
+          op: "add" | "replace" | "copy" | "test";
+          /**
+           * The JSON Pointer to the target document location at which to complete the operation.
+           */
+          path: string;
+          value: Record<string, unknown>;
+      }
+    | {
+          op: "move";
+          path: string;
+          value: Record<string, unknown>;
+          /**
+           * The JSON Pointer to the target document location from which to move the value.
+           */
+          from: string;
+      }
+    | {
+          op: "remove";
+          path: string;
+      };
+
+export type UpdateOrderRequestBody = UpdateOrderOperation[];
+
 export type OrderResponseBodyMinimal = {
     /**
      * The ID of the order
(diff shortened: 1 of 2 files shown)


types/apis/orders.d.ts
```diff
@@ -235,6 +235,30 @@ export type CreateOrderRequestBody = {
     application_context?: OrderApplicationContext;
 };
 
+export type UpdateOrderRequestBody = {
```
gregjopa: Thanks for sending this PR @jocmp! @xuanwan05 can you help review?

Also, we do have this shipping change example to cross-reference: https://github.com/paypal-examples/paypal-sdk-server-side-integration/blob/main/src/order/patch-order.ts#L28

jocmp: Hey thanks for the link @gregjopa. 

I re-pushed the commit to include the body as an array instead of an object based on https://github.com/paypal-examples/paypal-sdk-server-side-integration/blob/93cd4e03c354f3140807ee694e43189509a80c38/src/order/patch-order.ts#L36.

types/apis/orders.d.ts
```diff
@@ -235,6 +235,32 @@ export type CreateOrderRequestBody = {
...
+    /**
+     * The value to apply. The `remove` operation does not require a value.
+     */
+    value?: Record<string, any>;
+    /**
+     * The JSON Pointer to the target document location from which to move the value.
+     * Required for the `move` operation.
+     */
+    from?: string;
+};
```
wsbrunson:
```suggestion
export type UpdateOrderOperation =
    | {
          op: "add" | "replace" | "copy" | "test";
          /**
           * The JSON Pointer to the target document location at which to complete the operation.
           */
          path: string;
          value: Record<string, any>;
      }
    | {
          op: "move";
          path: string;
          value: Record<string, any>;
          /**
           * The JSON Pointer to the target document location from which to move the value.
           */
          from: string;
      }
    | {
          op: "remove";
          path: string;
      };
```

wsbrunson (commented):
one suggestion to codify the required fields for each operation using types

types/apis/orders.d.ts
jocmp: Updated!

wsbrunson (approved):

xuanwan05 (approved):
LG2M

PR #361, opened 2023-04-12
chore(types): add test for patching order

gregjopa:
This PR adds a test to validate the new types added in #360 for the [Update Order patch operation](https://developer.paypal.com/docs/api/orders/v2/#orders_patch).

It also removes the "move" operation from the patch api call types. Based on [this list](https://developer.paypal.com/docs/api/orders/v2/#orders_patch), no attributes actually support the "move" operation.

### DIFF
--- a/types/tests/server.test.ts
+++ b/types/tests/server.test.ts
@@ -168,3 +168,13 @@ updateOrder({
         },
     ],
 });
+
+updateOrder({
+    orderID: "42P22220TW111111R",
+    requestBody: [
+        {
+            op: "remove",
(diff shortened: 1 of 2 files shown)


types/tests/server.test.ts
```diff
@@ -145,3 +154,17 @@ createOrder({
         },
     ],
 });
+
+updateOrder({
```
wsbrunson, line 158: you might think about adding a case for `move` and `remove` as well

gregjopa: Test added for "remove". I ended up removing the type for "move" since I don't see any attributes supporting that operation.

PR #376, opened 2023-06-01
chore(tests): use vitest

gregjopa:
This PR replaces Jest with [vitest](https://vitest.dev/).

I stuck with jsdom since happy-dom executes `<script>` tags by default and was difficult to switch out. There are also some differences in type definitions for the DOM. Sticking with jsdom made this upgrade pretty easy.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,13 +15,13 @@
         "prerelease": "npm run validate",
         "release": "standard-version --commit-all",
         "postrelease": "git push && git push --follow-tags && npm run build && npm publish",
-        "test": "jest src",
-        "test:bundle": "jest bundle-tests/**",
+        "test": "vitest src",
+        "test:bundle": "vitest bundle-tests/**",
         "test:e2e": "playwright test",
         "test:e2e:debug": "PWDEBUG=1 playwright test --project=\"chromium\"",
         "test:e2e:start": "node e2e-tests/http-server.js",
         "typecheck": "tsc --noEmit",
-        "validate": "npm run check-node-version && npm run format:check && npm run typecheck && npm run build && npm run lint && npm test -- --coverage && npm run test:bundle"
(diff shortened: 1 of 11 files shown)


src/load-script.test.ts
```diff
@@ -319,17 +319,4 @@ describe("loadCustomScript()", () => {
         );
         expect(PromisePonyfill).toHaveBeenCalledTimes(1);
     });
-
-    test("should throw an error when the Promise implementation is undefined", () => {
```
gregjopa, line 323: Admittedly, this is a bad test that had to delete the global Promise object. I ended up removing this test since I was unable to get it working with vitest. 

I'm thinking we can actually get rid of this custom error message and just let the browser throw a standard error message in IE11 when `window.Promise` is undefined. https://github.com/paypal/paypal-js/blob/main/src/load-script.ts#L117-L119

package.json
```diff
@@ -15,13 +15,13 @@
         "prerelease": "npm run validate",
         "release": "standard-version --commit-all",
         "postrelease": "git push && git push --follow-tags && npm run build && npm publish",
-        "test": "jest src",
-        "test:bundle": "jest bundle-tests/**",
+        "test": "vitest src",
+        "test:bundle": "vitest bundle-tests/**",
```
wsbrunson, line 19: whats the difference between these test suites? is one really slow?

I would recommend just running all tests in `test` unless the bundle tests are really slow

wsbrunson (approved):

package.json
gregjopa: The one gotcha with `npm run test:bundle` is it expects the `npm run build` command to run first so the `dist` folder exists. It expects the built assets to be available to read from the file system to check the size.

wsbrunson: ah yeah that would be annoying then. LGTM

PR #391, opened 2023-06-28
Update deps and other misc changes

gregjopa:
- Update deps
- Upgrade to Node 18
- Refactor tests to use mock over spy
- ~Only attempt to parse the validation error message for 400 status code~


src/load-script.ts
```diff
@@ -96,7 +96,7 @@ export function loadCustomScript(
                 // Fetch the error reason from the response body for validation errors
                 return fetch(url)
                     .then((response) => {
-                        if (response.status === 200) {
+                        if (response.status !== 400) {
```
wsbrunson: would this reject on a 200 too?

gregjopa: Yes this code is kinda gnarly. If this `onError()` callback happens, it means the script failed to load. A couple years ago we added this fetch logic to attempt to get at the response body to find out why the script failed to load. It was added to try and show the validation error message to the end user (ex: https://www.paypal.com/sdk/js?client-id=foo). If a 2xx/3xx status code comes back from this fetch(), we can't do much at this point. Maybe we could add retry logic for the original `insertScriptElement()` operation 🤔.

dtjones404 (approved):

src/load-script.ts
gregjopa: I was thinking more about this. I'm gonna revert this src/load-script.ts change in this PR. And send a new PR to remove all of this `fetch()` onError logic all together. I think it can just be replaced with a message to devs telling them to check the Dev Tools Network tab to see the status code, response body, and paypal-debug-id to better troubleshoot.

wsbrunson: nice sounds good to me

wsbrunson (approved):

PR #419, opened 2023-08-25
Minor type defintion changes for script loading options

CBoensel:
See issue description of #418 and commit messages below. 

Hope that works. 

@gregjopa Let me know if anything else is needed from my end. Thanks.

PR #420, opened 2023-08-26
fix(types): query params as an array or string

gregjopa:
Follow up to #419. Applies the same array or string type logic to `components` and also fixes the types for `disableFunding` and `integrationDate`.

cc: @CBoensel


CBoensel:
Thanks for having a look. `components` makes a lot of sense as well.

PR #427, opened 2023-09-18
feat: add type definitions for two new callbacks

jshawl:
This PR adds type definitions for two new callbacks that were implemented in https://github.com/paypal/paypal-smart-payment-buttons/pull/679

Also see the react-paypal-js changes, which consume these changes here: https://github.com/paypal/react-paypal-js/pull/388

### DIFF
--- a/types/components/buttons.d.ts
+++ b/types/components/buttons.d.ts
@@ -108,6 +108,81 @@ export type OnShippingChangeActions = {
     };
 };
 
+type CurrencyCodeAndValue = {
+    currencyCode: string;
+    value: string;
+};
+
+type ShippingOption = {
+    amount: CurrencyCodeAndValue;
+    id?: string;
+    label: string;
+    selected: boolean;
+    type: "SHIPPING" | "PICKUP";
+};
+
+export type OnShippingOptionsChangeData = {
+    orderID?: string;
+    paymentID?: string;
+    paymentToken?: string;
+    selectedShippingOption?: ShippingOption;
+};
+
+type BuildOrderPatchPayloadOptions = {
+    discount?: string;
+    handling?: string;
+    insurance?: string;
+    itemTotal?: string;
+    option?: ShippingOption;
+    shippingOptions?: ShippingOption[];
+    shippingDiscount?: string;
+    taxTotal?: string;
+};
+
+export type OnShippingOptionsChangeActions = {
+    buildOrderPatchPayload: ({
+        discount,
+        handling,
+        insurance,
+        itemTotal,
+        shippingOptions,
+        shippingDiscount,
+        taxTotal,
+    }: BuildOrderPatchPayloadOptions) => UpdateOrderRequestBody;
+    reject: () => Promise<void>;
+};
+
+export type OnShippingAddressChangeData = {
+    amount: CurrencyCodeAndValue;
+    orderID?: string;
+    paymentID?: string;
+    paymentToken?: string;
+    shippingAddress: {
+        city: string;
+        state: string;
+        /** The [two-character ISO 3166-1 code](/docs/integration/direct/rest/country-codes/) that identifies the country or region. */
+        countryCode: string;
+        /**
+         * The postal code, which is the zip code or equivalent.
+         * Typically required for countries with a postal code or an equivalent.
+         */
+        postalCode: string;
+    };
+};
+
+export type OnShippingAddressChangeActions = {
+    buildOrderPatchPayload: ({
+        discount,
+        handling,
+        insurance,
+        itemTotal,
+        shippingOptions,
+        shippingDiscount,
+        taxTotal,
+    }: BuildOrderPatchPayloadOptions) => UpdateOrderRequestBody;
+    reject: () => Promise<void>;
+};
+
 export interface PayPalButtonsComponentOptions {
     /**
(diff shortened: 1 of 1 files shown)


types/components/buttons.d.ts
```diff
@@ -108,6 +108,81 @@ export type OnShippingChangeActions = {
     };
 };
 
+type CurrencyCodeAndValue = {
+    currencyCode: string;
+    value: string;
+};
```
jshawl, line 114: there is a similar type here: https://github.com/paypal/paypal-js/blob/aa10eff7be5b7cad8d3bd0b446ceec29fcfd736a/types/apis/commons.d.ts#L1-L12 but I wasn't sure if this new type should be added there, as it is not shared with any other types

types/components/buttons.d.ts
```diff
@@ -108,6 +108,81 @@ export type OnShippingChangeActions = {
...
+        discount,
+        handling,
+        insurance,
+        itemTotal,
+        shippingOptions,
+        shippingDiscount,
+        taxTotal,
+    }: BuildOrderPatchPayloadOptions) => UpdateOrderRequestBody;
+    reject: () => Promise<void>;
+};
```
jshawl, line 193: I don't think these types need to be exported if they are not imported from anywhere else. I'd lean toward removing the `export` in this case..

types/components/buttons.d.ts
nbierdeman: I think you're good here on defining a separate `type`, since it's not intended to be extended or used as a class contract like with an `interface`.

types/components/buttons.d.ts
```diff
@@ -108,6 +108,81 @@ export type OnShippingChangeActions = {
...
+    value: string;
+};
+
+type ShippingOption = {
+    amount: CurrencyCodeAndValue;
+    id?: string;
+    label: string;
+    selected: boolean;
+    type: "SHIPPING" | "PICKUP";
+};
```
nbierdeman: How would you feel about naming this `CheckoutShippingOption` to match the [equivalent type](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/types.js#L381-L390) in SPB?

types/components/buttons.d.ts
```diff
@@ -108,6 +108,81 @@ export type OnShippingChangeActions = {
...
+type BuildOrderPatchPayloadOptions = {
+    discount?: string;
+    handling?: string;
+    insurance?: string;
+    itemTotal?: string;
+    option?: ShippingOption;
+    shippingOptions?: ShippingOption[];
+    shippingDiscount?: string;
+    taxTotal?: string;
+};
```
nbierdeman: How about using `BuildOrderPatchPayloadArgs` [like in SPB](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/props/onShippingAddressChange.js#L34-L42)?

types/components/buttons.d.ts
```diff
@@ -108,6 +108,81 @@ export type OnShippingChangeActions = {
...
+
+export type OnShippingOptionsChangeActions = {
+    buildOrderPatchPayload: ({
+        discount,
+        handling,
+        insurance,
+        itemTotal,
+        shippingOptions,
```
nbierdeman: I'm thinking this should be singular, `shippingOption`, like [here](https://github.com/paypal/paypal-smart-payment-buttons/blob/main/src/props/onShippingOptionsChange.js#L34).

types/components/buttons.d.ts
nbierdeman: I'm with you on that, though I left at least one `export` in SPB even though it wasn't being imported anywhere. 😄

types/components/buttons.d.ts
```diff
@@ -165,11 +240,26 @@ export interface PayPalButtonsComponentOptions {
...
     ) => Promise<void>;
+    /**
+     * Called when the buyer selects a new shipping option on PayPal.
+     */
+    onShippingOptionsChange?: (
+        data: OnShippingOptionsChangeData,
+        actions: OnShippingOptionsChangeActions,
+    ) => Promise<void>;
+    /**
+     * Called when the buyer their shipping address on PayPal.
```
nbierdeman:
```suggestion
     * Called when the buyer updates their shipping address on PayPal.
```

nbierdeman (approved):
Nice work! 🪨 ⭐

types/components/buttons.d.ts
```diff
@@ -165,11 +249,26 @@ export interface PayPalButtonsComponentOptions {
...
     ) => Promise<void>;
+    /**
+     * Called when the buyer selects a new shipping option on PayPal.
+     */
+    onShippingOptionsChange?: (
+        data: OnShippingOptionsChangeData,
+        actions: OnShippingOptionsChangeActions,
+    ) => Promise<void>;
+    /**
+     * Called when the buyer updates their shipping address on PayPal.
```
jshawl, line 266: should this be "changes" instead of "updates"?

nbierdeman: Oh man, I don't know the difference. A buyer can either add one to begin with or select a different one that's already on file. Does that help? Probably not. 🤣

wsbrunson (approved):

gregjopa (approved):

PR #435, opened 2023-11-13
feat: add support for crossorigin script attribute

### LINKED ISSUE #434: How to inject crossorigin="anonymous" in script tag when using loadScript
I am using Nuxt 3 and @paypal/paypal-js Module below is my code and trying to figure out how can i inject crossorigin in the generated script tag.
We are trying to solve this issue https://github.com/Baroshem/nuxt-security/issues/255
```
<script setup>
import { loadScript } from '@paypal/paypal-js';
const paypalSdk = await loadScript({
    clientId: paypalClientID,
    currency: 'USD',
    disableFunding: 'card'
  });
  paypalSdk.Buttons({
    // Styling
    style: {
      layout: 'vertical',
      color: 'gold',
      shape: 'rect',
      label: 'buynow',
    },
    // Sets up the transaction when a payment button is clicked
    createOrder: async function (data, actions) {
      const { data: createdProductDetail } = await useFetch(api_url, 'POST', { 'slug': slug });
      return createdProductDetail.value.id;
    },
    // Finalize the transaction after payer approval
    onApprove: async function (data, actions) {
      const { data: orderData } = await useFetch(api_url + `${data.orderID}/capture/`, 'POST', { 'slug': slug });
      if (orderData.value.status === "COMPLETED") {
          return navigateTo('/success', { replace: true })
      }
    },
  }).render("#paypal-button-container");
}
</script>
<template>
  <div id="paypal-button-container"></div>
</template>
```

gregjopa:
This PR adds support for the [crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) script attribute. 

For example, the following `loadScript()` call will create the following script tag in the DOM:

```js
loadScript({ clientId: "test", crossorigin: "anonymous" })
```
```html
<script src="https://www.paypal.com/sdk/js?client-id=test" crossorigin="anonymous"></script>
```

Resolves #434

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -64,7 +64,7 @@ export function insertScriptElement({
 
 export function processOptions(options: PayPalScriptOptions): {
     url: string;
-    dataAttributes: StringMap;
+    attributes: StringMap;
 } {
     let sdkBaseUrl = "https://www.paypal.com/sdk/js";
 
@@ -76,7 +76,7 @@ export function processOptions(options: PayPalScriptOptions): {
     const optionsWithStringIndex =
         options as PayPalScriptOptionsWithStringIndex;
 
-    const { queryParams, dataAttributes } = Object.keys(optionsWithStringIndex)
+    const { queryParams, attributes } = Object.keys(optionsWithStringIndex)
         .filter((key) => {
             return (
                 typeof optionsWithStringIndex[key] !== "undefined" &&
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -64,7 +64,7 @@ export function insertScriptElement({
 
 export function processOptions(options: PayPalScriptOptions): {
     url: string;
-    dataAttributes: StringMap;
+    attributes: StringMap;
```
gregjopa, line 67: I renamed `dataAttributes` to `attributes` to better reflect that both data attributes and script attributes are included.

nbierdeman (approved):
💯 🚢

wsbrunson (approved):

PR #439, opened 2023-11-22
fix: add type definitions for createVaultSetupToken

westeezy:
Adding support for the `createVaultSetupToken` callback to our type definitions. Documented [here](https://developer.paypal.com/docs/checkout/save-payment-methods/purchase-later/js-sdk/paypal/#link-clientsidecodesample)

### DIFF
--- a/types/components/buttons.d.ts
+++ b/types/components/buttons.d.ts
@@ -214,7 +214,7 @@ export interface PayPalButtonsComponentOptions {
         actions: CreateSubscriptionActions,
     ) => Promise<string>;
     /**
-     * Save payment methods to charge payers after a set amount of time. For example, you can offer a free trial and charge payers after the trial expires. Payers don't need to be present when charged. No checkout required.a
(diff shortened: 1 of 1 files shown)


wsbrunson (approved):

elizabethmv (approved):

types/components/buttons.d.ts
```diff
@@ -213,6 +213,11 @@ export interface PayPalButtonsComponentOptions {
         data: Record<string, unknown>,
         actions: CreateSubscriptionActions,
     ) => Promise<string>;
+    /**
+     * Save payment methods to charge payers after a set amount of time. For example, you can offer a free trial and charge payers after the trial expires. Payers don't need to be present when charged. No checkout required.a
```
gregjopa: Looks like a trailing `a`. Or maybe you meant `eh?` if you really want a [canadianism](https://www.cs.cmu.edu/~steffan/canadianisms.html) at the end of this awesome comment.
```suggestion
     * Save payment methods to charge payers after a set amount of time. For example, you can offer a free trial and charge payers after the trial expires. Payers don't need to be present when charged. No checkout required.
```

PR #443, opened 2023-11-27
Auto generate api types with openapi specs

gregjopa:
## Summary

PayPal released the Open API 3.0 specs earlier this year: https://github.com/paypal/paypal-rest-api-specifications. Now we can auto-generate the TypeScript types for PayPal API requests and responses. This PR uses [openapi-typescript](https://github.com/drwpow/openapi-typescript) to auto generate the v2 Orders API types and the v1 Subscriptions API types. 

## Details

These new API types will make it much easier for TypeScript developers to understand the PayPal APIs. It will also make it much easier for us to maintain. We will no longer need to manually write types to stay in sync with what's documented on https://developer.paypal.com/api/rest/.

The best way to understand this change is to take a look at this real world example of writing server-side code in TypeScript that wraps the PayPal APIs: https://github.com/paypal-examples/paypal-sdk-server-side-integration/pull/37

Here's a snippet with wrapping the create order api call for reference:

```ts
// use this library to get the types for the request headers, request body, response body, and more
import type {
  CreateOrderRequestBody,
  OrderSuccessResponseBody,
  OrderSuccessResponseBodyMinimal,
  OrderErrorResponseBody,
  CreateOrder,
} from "@paypal/paypal-js";

type CreateOrderSuccessResponse = {
  status: "ok";
  data: OrderSuccessResponseBody | OrderSuccessResponseBodyMinimal;
  httpStatusCode: 200 | 201;
};

type CreateOrderErrorResponse = {
  status: "error";
  data: OrderErrorResponseBody;
  httpStatusCode: number;
};

// the Authorization header is missing from the headers list
type CreateOrderRequestHeaders = Partial<
  CreateOrder["parameters"]["header"]
> & {
  Authorization?: string;
};

type CreateOrderOptions = {
  body: CreateOrderRequestBody;
  headers?: CreateOrderRequestHeaders;
};

export default async function createOrder({
  body,
  headers = {},
}: CreateOrderOptions): Promise<
  CreateOrderSuccessResponse | CreateOrderErrorResponse
> {

  const requestHeaders = {
    "Content-Type": "application/json",
    Authorization: `Bearer ${accessToken}`,
    "Accept-Language": "en_US",
    Prefer: "return=minimal",
    ...headers,
  };

  let response;
  try {
    response = await fetch(`${apiBaseUrl}/v2/checkout/orders`, {
      method: "POST",
      headers: requestHeaders,
      body: JSON.stringify(body),
    });

    const data = await response.json();

    if (response.ok) {
      return {
        status: "ok",
        data:
          requestHeaders.Prefer === "return=minimal"
            ? (data as OrderSuccessResponseBodyMinimal)
            : (data as OrderSuccessResponseBody),
        httpStatusCode: response.status,
      } as CreateOrderSuccessResponse;
    } else {
      return {
        status: "error",
        data,
        httpStatusCode: response.status,
      } as CreateOrderErrorResponse;
    }
  } catch (error) {
    const httpError: HttpErrorResponse =
      error instanceof Error ? error : new Error(defaultErrorMessage);
    httpError.statusCode = response?.status;
    throw httpError;
  }
}
```

### Is this a breaking change?

Yes

### How can I test this?

Use this beta to test out the new type contracts.

```
npm install @paypal/paypal-js@8.0.0-beta.5
```

### DIFF
--- a/types/components/buttons.d.ts
+++ b/types/components/buttons.d.ts
@@ -90,7 +90,7 @@ export type OnCancelledActions = {
     redirect: () => void;
 };
 
-type PartialShippingAddress = {
+export type PartialShippingAddress = {
     city: string;
     state: string;
     country_code: string;


types/tests/openapi.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    UpdateOrderRequestBody,
+} from "../index";
+
+import type {
+    components
+} from "../apis/openapi/checkout_orders_v2";
+
+function createOrder(
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    createOrderRequestBody: components["schemas"]["order_request"],
```
gregjopa: This is the format for accessing the api types: ` components["schemas"]["order_request"]`

package.json
```diff
@@ -46,6 +47,7 @@
     "devDependencies": {
         "@commitlint/cli": "17.7.1",
         "@commitlint/config-conventional": "17.7.0",
+        "@paypal/paypal-rest-api-specifications": "github:paypal/paypal-rest-api-specifications",
```
gregjopa: Ideally we'd get these specs published to the public npm registry. Pulling directly from github for now: https://github.com/paypal/paypal-rest-api-specifications

package.json
```diff
@@ -21,7 +21,8 @@
         "test:e2e:debug": "PWDEBUG=1 playwright test --project=\"chromium\"",
         "test:e2e:start": "node e2e-tests/http-server.js",
         "typecheck": "tsc --noEmit",
-        "validate": "npm run check-node-version && npm run format:check && npm run typecheck && npm run build && npm run lint && npm test -- --coverage --watch=false && npm run test:bundle -- --watch=false"
+        "validate": "npm run check-node-version && npm run format:check && npm run typecheck && npm run build && npm run lint && npm test -- --coverage --watch=false && npm run test:bundle -- --watch=false",
+        "openapi": "npx openapi-typescript node_modules/@paypal/paypal-rest-api-specifications/openapi/checkout_orders_v2.json -o types/apis/openapi/checkout_orders_v2.d.ts"
```
gregjopa: If this goes well we should update this command to work for multiple schemas. These is a convention around this with using a redocly.yaml file: https://openapi-ts.pages.dev/cli#multiple-schemas

package.json
gregjopa: The @paypal/paypal-rest-api-specifications package has been published to the public npm registry.

gregjopa:
@paypal/checkout-sdk please review this PR.

nbierdeman (approved):
Wow, so much less to maintain. Nice!!! 🥇

wsbrunson (approved):
pretty dang cool

subvertallchris:
I was so relieved to find this, thank you! It would be great to add the original example in your PR to the docs. It provides great 
background and example usage. For anyone struggling with errors related to import, you might need to use `import type` like this:

```
import type { CheckoutOrdersV2, CreateOrderRequestBody, OrderResponseBody } from '@paypal/paypal-js';
```

With an OpenAPI spec, you could go a step further and generate a type-safe API client. Is there any interest in that?

PR #453, opened 2024-02-01
Convert paypal-js to a monorepo with Changesets and npm workspaces

wsbrunson:
The motivation for this change is to create a central repository for all of PayPal's JS SDK libraries that are intended for merchant use. We'd like this to be the one place to go to find framework-specific SDK wrappers, helper libraries, and constants/types. We'd also like it to be the single place where developers can open Github Issues about the PayPal SDK.

There are 2 changes to discuss that impact the workflow. We are heavily borrowing from [grumbler-scripts](https://github.com/krakenjs/grumbler-scripts) if you'd like to see a current working example.

**Note**: I'm turning off prettier checking for now so that we don't mess up the diff too badly. Once this is merged, I'm going to create a separate PR to upgrade the version of prettier and remove any config options so that we are using the base prettier options.

### Moving to Changesets
Before, this repo was using standard commits to manage versioning and changelogs. Its a nice tool for simple usecases but something like Changesets makes managing packages in a monorepo incredibly simple. It also doesn't require any formatting for commit messages, it has a cli helper to create a changelog entry and determine the type of change (patch, minor, major). Changesets will handle versioning each package and deploying them to npm.

### NPM workspaces
The root package.json has a `workspaces` property now pointing to the `/packages` folder. You can learn more about [workspaces here](https://docs.npmjs.com/cli/v8/using-npm/workspaces). The general gist is that they are a great way to handle repositories with multiple npm packages provided directly from the npm team.

In the future, we could consider a solution like Turbo if our requirements get more complicated. For now though, workspaces should suffice.


ravishekhar (approved):
🎉

.changeset/README.md
```diff
@@ -0,0 +1,8 @@
+# Changesets
```
gregjopa, line 1: Should we add a comment to the existing CHANGELOG.md file mentioning that it will no longer be used and to view this changeset folder instead?

.github/workflows/main.yml
```diff
@@ -30,14 +24,16 @@ jobs:
 
             - name: 📥 Download deps
               uses: bahmutov/npm-install@v1
-              with:
-                  useLockFile: false
 
-            - name: ▶️ Run validate script
-              run: npm run validate
```
gregjopa, line 37: I think `npm run validate` also does the prettier formatting check and the TypeScript type checks. I'm open to changing this but really want to keep CI doing the type checking and formatting check

README.md
```diff
@@ -1,298 +1,52 @@
-# paypal-js
+## PayPal JS Monorepo
 
-> Loading wrapper and TypeScript types for the [PayPal JS SDK](https://developer.paypal.com/sdk/js/reference/)
+This is a collection of libraries intended to help developers more easily integrate with PayPal's JS SDK
```
gregjopa, line 3: ❤️

.changeset/README.md
westeezy: No need, `changesets` will write to the existing changelog on the top. this file just talks about how to set up `changesets`

README.md
```diff
@@ -1,298 +1,52 @@
...
-```
-
-## TypeScript Support
-
-This package includes TypeScript type definitions for the PayPal JS SDK. This includes types for the `window.paypal` namespace. We support projects using TypeScript versions >= 3.8.
-
-## Releasing
-
-Run `npm run release` to publish a new release. The version number is determined by the git commits which follow [conventional commits spec](https://www.conventionalcommits.org).
+There is no Github Action for alpha release at this time. Because this repo utilizes changesets we can follow their process locally in the meantime. This document can be seen [here](https://github.com/changesets/changesets/blob/main/docs/prereleases.md).
```
westeezy, line 52: Ive been using the `pre` mode on changesets with success which does alphas. fwiw

westeezy (approved):

.changeset/README.md
wsbrunson: Oh yeah, we should. changesets will create and start using CHANGELOG.mds in each package folder. We can do what grumbler-scripts did

```
WARNING This is the legacy changelog from before v8. For up to date changes see individual packages
```

root: https://github.com/krakenjs/grumbler-scripts/blob/main/CHANGELOG.md

package: https://github.com/krakenjs/grumbler-scripts/blob/main/packages/babel-config-grumbler/CHANGELOG.md

gregjopa: > No need, changesets will write to the existing changelog on the top. this file just talks about how to set up changesets

Great to hear. I didn't know that. Maybe we move the existing paypal-js changelog file into the paypal-js package folder. That way the new changeset functionality can continue to append to it and we don't need a deprecation message.

README.md
wsbrunson: oh nice. i pulled this from grumbler-scripts

.github/workflows/main.yml
wsbrunson: I think `npm run build` would handle typechecking for typescript packages but I will definitely add a `npm run format` task

package.json
```diff
@@ -1,86 +1,37 @@
...
-        "rollup": "4.9.1",
-        "semver": "7.5.4",
-        "standard-version": "9.5.0",
-        "tslib": "2.6.2",
-        "typescript": "5.3.3",
-        "vitest": "1.1.0"
+    "scripts": {
+        "build": "npm run build --workspaces --if-present",
+        "changeset:add": "changeset add",
+        "changeset:release": "npm run build && changeset publish",
```
jshawl, line 16: do we still have the ability to publish alpha releases with this change?

jshawl (approved):
looks great! just one small q about publishing alpha releases (I don't _need_ this functionality at the moment but just thinking ahead) 🚀 :shipit:

package.json
wsbrunson: we do, we'll just need to follow a different pattern: https://github.com/changesets/changesets/blob/main/docs/prereleases.md

that or just publish alpha versions directly in each package if we don't want changesets to be involved.

We could still potentially make a GH action for this, it would need to take branch and app name into account

gregjopa (approved):
💯

cgdibble (approved):

elizabethmv (approved):

Basheer-alsyani:
اريد السحب عبر بايبال

PR #465, opened 2024-03-15
Set data-js-sdk-library default value

nbierdeman:
### Description

In service of DTPPCPSDK-1482, this PR sets a data-js-sdk-library value for tracking library usage.

### DIFF
--- a/packages/paypal-js/e2e-tests/load-cached-script.html
+++ b/packages/paypal-js/e2e-tests/load-cached-script.html
@@ -13,6 +13,7 @@
             const options = {
                 clientId: "test",
                 dataPageType: "checkout",
+                dataJsSdkLibrary: "paypal-js",
             };
 
             // initial load and render
--- a/packages/paypal-js/e2e-tests/load-cached-script.html
+++ b/packages/paypal-js/e2e-tests/load-cached-script.html
@@ -13,7 +13,6 @@
             const options = {
                 clientId: "test",
                 dataPageType: "checkout",
-                dataJsSdkLibrary: "paypal-js",
             };
 
             // initial load and render
--- a/packages/paypal-js/src/load-script.ts
+++ b/packages/paypal-js/src/load-script.ts
@@ -27,10 +27,6 @@ export function loadScript(
         return PromisePonyfill.resolve(existingWindowNamespace);
     }
 
-    if (!attributes["data-js-sdk-library"]) {
-        attributes["data-js-sdk-library"] = "paypal-js";
-    }
-
     return loadCustomScript(
         {
             url,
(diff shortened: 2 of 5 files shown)


packages/paypal-js/src/load-script.ts
```diff
@@ -19,6 +19,14 @@ export function loadScript(
     if (typeof document === "undefined") return PromisePonyfill.resolve(null);
 
     const { url, attributes } = processOptions(options);
+
+    if (
```
gregjopa: I'd prefer to skip the validation on the existing value and only set it if the data attribute value is undefined. That way we can add new libraries in the future like vue-paypal-js and not have to worry about updating this code path.
```js
    if (!attributes["data-js-sdk-library"]) {
        attributes["data-js-sdk-library"] = "paypal-js";
    }
```

nbierdeman: Okay, cool. I'll need to look into this, because it ends up inserting a new script in the DOM even when there's already one with the same params.

packages/paypal-js/src/load-script.ts
```diff
@@ -64,6 +64,10 @@ export function loadCustomScript(
 
     const { url, attributes } = options;
 
+    if (attributes && !attributes["data-js-sdk-library"]) {
```
nbierdeman: `loadCustomScript` will always have `attributes` when called by `react-paypal-js`/`paypal-js` -> `loadScript`, but checking for `attributes` here, since `react-paypal-js` calls `loadCustomScript` without `attributes` in `getBraintreeNamespace` [here](https://github.com/paypal/react-paypal-js/blob/e365997c6c31bf459438f2b191884a1507b20b97/src/components/braintree/utils.ts#L96-L97).

packages/paypal-js/e2e-tests/load-cached-script.html
```diff
@@ -13,6 +13,7 @@ <h1>Load Cached Script</h1>
             const options = {
                 clientId: "test",
                 dataPageType: "checkout",
+                dataJsSdkLibrary: "paypal-js",
```
nbierdeman: This smells funky, because `dataJsSdkLibrary` is not a required script option, but if we don't add it here, the "load cached script" tests fail.

packages/paypal-js/src/load-script.ts
nbierdeman: @gregjopa, I tried adding the `data-js-sdk-library` default value in `loadScript`, `processOptions`, `loadCustomScript`, and `insertScriptElement`, and ended up moving it back to `loadScript`. Definitely open to trying something else too, but marking this ready for review for now.

gregjopa (approved):

gregjopa (approved):

packages/paypal-js/src/load-script.ts
gregjopa: I think you added it in the right spot in loadScript() @nbierdeman  💯

wsbrunson (approved):

PR #475, opened 2024-03-26
Update onApprove Param Type 👾

spencersablan:
**JIRA:** [LI-40638](https://paypal.atlassian.net/browse/LI-40638)

This will help us align with the types in our docs 😄 

### Helpful note
I didn't realize this repo auto-formats when you commit. So here's the lines I actually changed:
- [Made facilitatorAccessToken optional in onApproveData](https://github.com/paypal/paypal-js/pull/475/files#diff-66eb7d7ed93786f74449389e63635c5bddcd2d12912a143640e96416e3498a51R49)
- [Changed onApprove param from err to data with onApproveData as the type](https://github.com/paypal/paypal-js/pull/475/files#diff-3cd5a8f75e518348830337978d300aa8b7c7f29407d2b1cd629aaa4fc1b30a46R129)

### DIFF
--- a/packages/paypal-js/types/components/card-fields.d.ts
+++ b/packages/paypal-js/types/components/card-fields.d.ts
@@ -1,5 +1,3 @@
-import { OnApproveData } from "./buttons";
-
 export interface PayPalCardFieldsStyleOptions {
     appearance?: string;
     color?: string;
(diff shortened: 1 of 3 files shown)


packages/paypal-js/types/components/card-fields.d.ts
```diff
@@ -1,3 +1,5 @@
+import { OnApproveData } from "./buttons";
```
gregjopa: I worry about sharing the exact same onApproveData object from buttons. Could we have cardFields have its own unique type for it? That way they can be independent in the future and any future changes need to be explicit for both.

wsbrunson: 👍 they may be slightly different. if we end up with 3+ onApprove definitions we could at that time write some type of `baseOnApproveData`

spencersablan: Ahh that's a great idea. 😄

nbierdeman (approved):
👽

gregjopa (approved):

PR #485, opened 2024-04-15
Introduce env parameter to support sandbox and paypal 

ravishekhar:
* Introduce a new option in the `loadScript` method to allow passing `env = sandbox or production`.
*  If `env=sandbox` is passed as an option, then the sdk script will be loaded from `www.sandbox.paypal.com/sdk/js`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 /test-results/
 /playwright-report/
 /playwright/.cache/
+.idea
--- a/packages/paypal-js/src/load-script.ts
+++ b/packages/paypal-js/src/load-script.ts
@@ -106,7 +106,11 @@ function validateArguments(options: unknown, PromisePonyfill?: unknown) {
     }
     const { environment } = options as PayPalScriptOptions;
 
-    if (environment && !["production", "sandbox"].includes(environment)) {
+    if (
+        environment &&
+        environment !== "production" &&
+        environment !== "sandbox"
+    ) {
         throw new Error(
             'The `environment` option must be either "production" or "sandbox".'
         );
--- a/packages/paypal-js/src/load-script.ts
+++ b/packages/paypal-js/src/load-script.ts
@@ -104,6 +104,13 @@ function validateArguments(options: unknown, PromisePonyfill?: unknown) {
     if (typeof options !== "object" || options === null) {
         throw new Error("Expected an options object.");
     }
+    const { environment } = options as PayPalScriptOptions;
+
+    if (environment && !["production", "sandbox"].includes(environment)) {
+        throw new Error(
+            'The `environment` option must be either "production" or "sandbox".'
+        );
+    }
 
     if (
         typeof PromisePonyfill !== "undefined" &&
--- a/packages/paypal-js/src/utils.ts
+++ b/packages/paypal-js/src/utils.ts
@@ -66,15 +66,15 @@ export function processOptions(options: PayPalScriptOptions): {
     url: string;
     attributes: StringMap;
 } {
-    const { env } = options;
+    const { environment } = options;
     // Keeping production as default to keep maintain backward compatibility.
     // In the future this logic needs to be changed to use sandbox domain as default instead of production.
     let sdkBaseUrl =
-        env === "sandbox"
+        environment === "sandbox"
             ? "https://www.sandbox.paypal.com/sdk/js"
             : "https://www.paypal.com/sdk/js";
     // env is not an allowed key for sdk/js url.
-    delete options.env;
+    delete options.environment;
 
     if (options.sdkBaseUrl) {
         sdkBaseUrl = options.sdkBaseUrl;
--- a/packages/paypal-js/src/utils.ts
+++ b/packages/paypal-js/src/utils.ts
@@ -66,7 +66,15 @@ export function processOptions(options: PayPalScriptOptions): {
     url: string;
     attributes: StringMap;
 } {
-    let sdkBaseUrl = "https://www.paypal.com/sdk/js";
+    const { env } = options;
+    // Keeping production as default to keep maintain backward compatibility.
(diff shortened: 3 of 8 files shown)


packages/paypal-js/types/script-options.d.ts
```diff
@@ -21,6 +21,7 @@ interface PayPalScriptQueryParameters {
 }
 
 interface PayPalScriptDataAttributes {
+    env?: "sandbox" | "production";
```
gregjopa: Would you be open to renaming to `environment`? I try to avoid abbreviations whenever possible. 
```suggestion
    environment?: "sandbox" | "production";
```

packages/paypal-js/src/utils.ts
```diff
@@ -66,7 +66,15 @@ export function processOptions(options: PayPalScriptOptions): {
     url: string;
     attributes: StringMap;
 } {
-    let sdkBaseUrl = "https://www.paypal.com/sdk/js";
+    const { env } = options;
+    // Keeping production as default to keep maintain backward compatibility.
```
nbierdeman:
```suggestion
    // Keeping production as default to maintain backward compatibility.
```

packages/paypal-js/types/script-options.d.ts
nbierdeman: 🔤 
```suggestion
    dataClientToken?: string;
    dataCspNonce?: string;
    dataClientMetadataId?: string;
    dataJsSdkLibrary?: string;
    // loadScript() supports an array and will convert it
    // to the correct dataMerchantId string values
    dataMerchantId?: string[] | string;
    dataNamespace?: string;
    dataPageType?: string;
    dataPartnerAttributionId?: string;
    dataSdkIntegrationSource?: string;
    dataUid?: string;
    dataUserIdToken?: string;
    env?: "sandbox" | "production";
```

.gitignore
```diff
@@ -6,3 +6,4 @@ node_modules
 /test-results/
 /playwright-report/
 /playwright/.cache/
+.idea
```
nbierdeman, line 9: Not using VS Code, eh? 😁

nbierdeman (approved):
Just a couple of small nitpicks, but LGTM. 🚀

packages/paypal-js/src/utils.ts
```diff
@@ -66,7 +66,15 @@ export function processOptions(options: PayPalScriptOptions): {
     url: string;
     attributes: StringMap;
 } {
-    let sdkBaseUrl = "https://www.paypal.com/sdk/js";
+    const { env } = options;
```
jshawl: I liked Greg's suggestion to use `environment` but also am curious about validating this value like:

```js
if (environment) {
  if (!["sandbox", "production"].includes(environment)) {
    throw new Error('The `environment` option must be either "sandbox" or "production"')
  }
}
```

jshawl (approved):
Looks great! 1 small non-blocking suggestion. :shipit:

packages/paypal-js/src/utils.ts
ravishekhar: Ah! great catch. I forgot the typescript interface isn't available in transpiled code.

packages/paypal-js/types/script-options.d.ts
ravishekhar: Done 🎉

ravishekhar: Yes sounds good to me ✅ .  I have also moved the type to the PayPalScriptOptions type

gregjopa: @ravishekhar I'm still seeing the name as `env` in the diff. Can we update to `environment`?

ravishekhar: Ah! I see my git push got denied earlier. Just did that again. Thanks for catching this.

packages/paypal-js/types/script-options.d.ts
```diff
@@ -45,4 +45,5 @@ export interface PayPalScriptOptions
         PayPalScriptDataAttributes,
         ScriptAttributes {
     sdkBaseUrl?: string;
+    environment?: "production" | "sandbox";
```
nbierdeman, line 47: 🔤 There ought to be a linting rule that does this for us. 😁 
```suggestion
    environment?: "production" | "sandbox";
    sdkBaseUrl?: string;
```

packages/paypal-js/src/load-script.ts
```diff
@@ -104,6 +104,13 @@ function validateArguments(options: unknown, PromisePonyfill?: unknown) {
     if (typeof options !== "object" || options === null) {
         throw new Error("Expected an options object.");
     }
+    const { environment } = options as PayPalScriptOptions;
+
+    if (environment && !["production", "sandbox"].includes(environment)) {
```
gregjopa: I prefer to remove this validation code for two reasons:
1. The logic in utils.ts will ignore any bad values and will default the production environment.
2. paypal-js is designed to compile to ES3 so using Array.prototype.includes() would be a breaking change. If we really want to keep this, let's replace the usage of includes() with something that transpiles nicely to ES3 and does not require a polyfill.

This loader is meant to be as small as possible on top of the PayPal JS SDK.

ravishekhar: I agree with the challenges with `Array.prototype.includes()` . We can definitely use an alternative there 👍 .

About not having validations on input, once production stops the support for sandbox, then this may become a source of errors, if there is any typos in the input options.

gregjopa: I see your point. Let's replace the usage of `Array.prototype.includes()` with some old school JS then :)

ravishekhar: done ✔️

gregjopa (approved):

PR #489, opened 2024-05-03
Throw error for missing client ID 👾

spencersablan:
- 👽 _**Added more meaningful error handling**_ - Before, without a client ID, if you tried to load the PayPal SDK, it would throw a generic error.  It would be helpful to explicitly point out that the developer did not include the client ID.

### DIFF
--- a/packages/paypal-js/src/load-script.ts
+++ b/packages/paypal-js/src/load-script.ts
@@ -14,11 +14,17 @@ export function loadScript(
     PromisePonyfill: PromiseConstructor = Promise
 ): Promise<PayPalNamespace | null> {
     validateArguments(options, PromisePonyfill);
-
     // resolve with null when running in Node or Deno
     if (typeof document === "undefined") return PromisePonyfill.resolve(null);
 
+    if (!options.clientId) {
+        throw new Error(
+            `Expected clientId in options object: ${JSON.stringify(options)}`
+        );
+    }
+
     const { url, attributes } = processOptions(options);
+
     const namespace = attributes["data-namespace"] || "paypal";
     const existingWindowNamespace = getPayPalWindowNamespace(namespace);
 
@@ -104,6 +110,7 @@ function validateArguments(options: unknown, PromisePonyfill?: unknown) {
     if (typeof options !== "object" || options === null) {
         throw new Error("Expected an options object.");
     }
+
     const { environment } = options as PayPalScriptOptions;
 
     if (
--- a/packages/paypal-js/src/load-script.ts
+++ b/packages/paypal-js/src/load-script.ts
@@ -17,13 +17,13 @@ export function loadScript(
     // resolve with null when running in Node or Deno
     if (typeof document === "undefined") return PromisePonyfill.resolve(null);
 
-    if (!options.clientId) {
+    const { url, attributes, queryParams } = processOptions(options);
+
+    if (!queryParams["client-id"]) {
         throw new Error(
             `Expected clientId in options object: ${JSON.stringify(options)}`
         );
     }
-
-    const { url, attributes } = processOptions(options);
 
     const namespace = attributes["data-namespace"] || "paypal";
     const existingWindowNamespace = getPayPalWindowNamespace(namespace);
--- a/.changeset/beige-tables-pull.md
+++ b/.changeset/beige-tables-pull.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
+---
+
+Improved error handling for missing clientId
--- a/packages/paypal-js/src/load-script.test.ts
+++ b/packages/paypal-js/src/load-script.test.ts
@@ -142,6 +142,14 @@ describe("loadScript()", () => {
         expect(() => loadScript({}, {})).toThrow(
             "Expected PromisePonyfill to be a function."
         );
+
+        const missingClientIdOptions = { components: "buttons" };
+        // @ts-expect-error ignore invalid arguments error
+        expect(() => loadScript(missingClientIdOptions)).toThrow(
+            `Expected clientId in options object: ${JSON.stringify(
+                missingClientIdOptions
+            )}`
+        );
     });
 });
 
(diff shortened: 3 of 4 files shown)


packages/paypal-js/src/load-script.ts
```diff
@@ -14,11 +14,17 @@ export function loadScript(
     PromisePonyfill: PromiseConstructor = Promise
 ): Promise<PayPalNamespace | null> {
     validateArguments(options, PromisePonyfill);
-
     // resolve with null when running in Node or Deno
     if (typeof document === "undefined") return PromisePonyfill.resolve(null);
 
+    if (!options.clientId) {
+        throw new Error(
+            `Expected clientId in options object: ${JSON.stringify(options)}`
```
jshawl, line 24: I like how you added what the existing options were, as it will make it easier to identify where the problem can be fixed!

packages/paypal-js/src/load-script.ts
```diff
@@ -104,6 +110,7 @@ function validateArguments(options: unknown, PromisePonyfill?: unknown) {
     if (typeof options !== "object" || options === null) {
         throw new Error("Expected an options object.");
     }
+
```
jshawl, line 113: should the `options.clientId` check come after line 112? I just noticed that `throw new Error("Expected an options object.");` will never be hit and/or the error message might be something like "cannot read attribute `clientId` of undefined"

packages/paypal-js/src/load-script.ts
```diff
@@ -14,11 +14,17 @@ export function loadScript(
     PromisePonyfill: PromiseConstructor = Promise
 ): Promise<PayPalNamespace | null> {
     validateArguments(options, PromisePonyfill);
-
     // resolve with null when running in Node or Deno
     if (typeof document === "undefined") return PromisePonyfill.resolve(null);
 
+    if (!options.clientId) {
```
imbrian: Do we want to move this into `loadCustomScript` with the other validations? Seems like this might start breaking on incorrectly-implemented sites where another namespace exists and [is used below](https://github.com/paypal/paypal-js/blob/f6511f9dca73b56fec9ed8dcfe32b8ae8284b8c2/packages/paypal-js/src/load-script.ts#L35).

jshawl (approved):
🚀 one small suggestion about moving the check down but I like the improvement!

packages/paypal-js/src/load-script.ts
spencersablan: I originally had it there but I realized that (per the README)`loadCustomScript` is just a generic script loader that should work with any URL. Whereas `loadScript` is specifically for the JS SDK script.

imbrian: Good catch. Maybe to line 39 then? At least would give
```js
// resolve with the existing global paypal namespace when a script with the same params already exists
    if (findScript(url, attributes) && existingWindowNamespace) {
        return PromisePonyfill.resolve(existingWindowNamespace);
 }
```
 a chance to gracefully handle an incorrect implementation in production today?

packages/paypal-js/src/load-script.ts
spencersablan: Good thought! We do call `validateArguments()` on line 16 as well, so we will still get a that `Expected an options object`  error if a developer doesn't pass anything into `loadScript()`

It is a bit odd that we are calling `validateArguments()` at the top of both `loadScript` and `loadCustomScript` when `loadScript` uses `loadCustomScript`... but there's probably a good reason for it. 😅

packages/paypal-js/src/load-script.ts
imbrian: Oh actually, that wouldn't find the existing instance, because the URL would be different due to missing `clientId`.

packages/paypal-js/src/load-script.ts
jshawl: ah you're right! pls ignore my comment as 112 will throw before the client id check

spencersablan:
@jshawl @imbrian 

I moved this logic down below `processOptions()` so that client ID would always be formatted as `"client-id"`.  Otherwise, it was throwing the error for: `{ "client-id": "abc" }` or `{ clientID: "abc" }`, which are both valid.

Although, @gregjopa brought up that we _used to_ have logic to fetch the error message if /sdk/js returns a 400 status code: [See here](https://github.com/paypal/paypal-js/blob/0a85965eeb57d54ddd6f41b143e9c0fb05897f49/src/load-script.ts#L88-L111)

This might be cool to reimplement since it would be able to log errors for missing client ID's, invalid client ID's, and so much more.  It was removed for maintenance purposes but Greg said if we think it would be valuable, we can reimplement it.

I'd love to know your thoughts. Is this needed?

wsbrunson (approved):

gregjopa:
> Although, @gregjopa brought up that we used to have logic to fetch the error message if /sdk/js returns a 400 status code: [See here](https://github.com/paypal/paypal-js/blob/0a85965eeb57d54ddd6f41b143e9c0fb05897f49/src/load-script.ts#L88-L111)

If we want to add input validation to this library, I recommend we bring back the fallback `fetch()` approach that was originally in paypal-js. That will solve for things like missing and invalid client-ids, invalid query param keys and invalid query param values. It also keeps the JS SDK itself responsible for owning the validation logic. And this library is just a thin wrapper on top.

All the code for this is in version history along with tests. Happy to help bring it back. I think we learned the hard way with the JS SDK that any time you build an API that returns JavaScript, you should always return back with a 200 status code so the browser can read the response. Script tags will ignore the content of 400/500 http responses and will not execute the JS that gets returned. So this solution is a workaround to that problem.

At a high level, here's the logic:
1. If the script fails to get inserted into the DOM, use the script `onerror` callback to fetch the details of the error.
2. The fetch() call made to /sdk/js with the same exact query params will enable us to read the http response body.
3. When we get back a 400/500 error, parse the body as plain text and pull out the error message along with the debug id.
4. Return this error message and debug id when rejecting the problem.
5. Celebrate improving the dev ex 🎉

PR #496, opened 2024-06-06
Migration/react paypal js

sebastianfdz:
Internal ticket reference: `DTPPCPSDK-2407`

- We are migrating our `react-paypal-js` repo to be a package inside of `paypal-js`.
- We want to keep the commit history of `react-paypal-js` in this migration so that is the reason why we are seeing all of the commits in this PR.


sebastianfdz:
Git issues

PR #552, opened 2024-08-09
Fix formatting

hnguyen510:
- remove prettier dependency from packages/paypal-js and packages/react-paypal-js
- update prettier version in the root from 2.7.1 to 3.1.1
- remove format:check from package.json and other references
- update node version from v16 to v18
- fixed some codes to fix build errors (will add comments where the file changed)
- run `npm run format` to fix formatting


packages/react-paypal-js/lint-staged.config.js
```diff
@@ -1,13 +1,13 @@
 module.exports = {
     "**/*.ts?(x)": (filenames) => [
         "npm run typecheck",
-        `prettier --write ${filenames.join(" ")}`,
+        // `prettier --write ${filenames.join(" ")}`,
```
imbrian: What's the reason for omitting these?

hnguyen510: Since i removed the prettier dependency in react-paypal-js, I thought we have to remove it. I was trying it out while fixing the build errors. Should i put it back?

packages/react-paypal-js/.nvmrc
```diff
@@ -1 +1 @@
-v16
+v18
```
wsbrunson: oh you know what? we should remove this and just rely on an `.nvmrc` at the root of the project.

Good idea updating this though

packages/react-paypal-js/lint-staged.config.js
wsbrunson: we should actually remove this whole file, all of the `lint-staged` stuff should be handled by the root of the project

wsbrunson (commented):
Great work, not only did you get all the prettier stuff removed and updated, you found some other parts that needed updating as well!

Just left 2 comments, once we resolve those we can merge this in

wsbrunson (approved):

package-lock.json
```diff
@@ -2275,6 +2275,21 @@
                 "graceful-fs": "^4.1.6"
             }
         },
+        "node_modules/@changesets/apply-release-plan/node_modules/prettier": {
+            "version": "2.8.8",
+            "resolved": "https://artifactory.g.devqa.gcp.dev.paypalinc.com/artifactory/api/npm/npm-all/prettier/-/prettier-2.8.8.tgz",
```
wsbrunson, line 2280: I have no idea how this would have happened. there is an `.npmrc` file at the root of the project which should force you to use the public npm registry. somehow it used the private artifactory registry though.

do you have a public npm account? If so i would do the following
* `npm uninstall --save prettier`
* `npm login` <- use your public account creds
* `npm install --save prettier@3`

imbrian (approved):

imbrian (approved):

PR #553, opened 2024-08-12
Export type for buttons components props

hnguyen510:
Split up each prop in  into its own type to export

// Old way
const styles: PayPalButtonsComponentOptions["styles"] = { ...your_button_styles }

// New way
const styles: PayPalButtonsStyles = { ...your_button_styles }

### DIFF
--- a/packages/paypal-js/types/components/buttons.d.ts
+++ b/packages/paypal-js/types/components/buttons.d.ts
@@ -271,9 +271,7 @@ export type PayPalButtonStyle = {
     shape?: "rect" | "pill" | "sharp";
     tagline?: boolean;
 };
-
 export type PayPalButtonDisplayOnly = DisplayOnlyOptions[];
-
 export type PayPalButtonMessage = {
     amount?: number;
     align?: "center" | "left" | "right";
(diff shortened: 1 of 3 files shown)


packages/paypal-js/types/components/buttons.d.ts
```diff
@@ -201,122 +201,155 @@ type OnShippingAddressChangeActions = {
 
 export type DisplayOnlyOptions = "vaultable";
 
+export type PayPalButtonCreateOrder = (
```
spencersablan, line 204: This is totally my fault for putting this in the ticket, but wanted to throw this out there:
Do you think we should prefix these types with `PayPalButtons___` instead of `PayPalButton___` to match `PayPalButtonsComponentOptions`? 🤔   

Open to either option 😄  But just thought I'd start the conversation.

spencersablan (approved):
This is going to be a great developer experience upgrade for TypeScript devs 😄  Great work.

One comment for conversation ⬇️

packages/paypal-js/types/components/buttons.d.ts
hnguyen510: I think `PayPalButton__` is better since we split them into a single button.

imbrian, line 204: Was there any discussion around why these types live in `paypal-js` instead of alongside the components they describe? I'm sure there's some history there, and it's not a blocker on this PR.

spencersablan: I don't know about any discussion around this.  Were you thinking of importing it from checkout-components?

PR #594, opened 2024-12-11
Refactor `processOptions` to prevent modifying its arguments

matheusrufca:
Removed `delete` statements from `processOptions` function to prevent modifying its argument and having side-effects.

Deleting `sdkBaseUrl` property is causing to React script loader to not properly set a custom `sdkBaseUrl` because `useEffect` is running twice.

### DIFF
--- a/packages/paypal-js/src/utils.ts
+++ b/packages/paypal-js/src/utils.ts
@@ -131,6 +131,16 @@ export function objectToQueryString(params: StringMap): string {
     return queryString;
 }
 
+export function processSdkBaseUrl(
+    environment: PayPalScriptOptions["environment"],
+): string {
+    // Keeping production as default to maintain backward compatibility.
+    // In the future this logic needs to be changed to use sandbox domain as default instead of production.
+    return environment === "sandbox"
+        ? "https://www.sandbox.paypal.com/sdk/js"
+        : "https://www.paypal.com/sdk/js";
+}
+
 function createScriptElement(
     url: string,
     attributes: StringMap = {},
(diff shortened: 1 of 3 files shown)


packages/paypal-js/src/utils.ts
```diff
@@ -62,24 +62,15 @@ export function insertScriptElement({
...
 } {
-    const { environment } = options;
-    // Keeping production as default to maintain backward compatibility.
-    // In the future this logic needs to be changed to use sandbox domain as default instead of production.
-    let sdkBaseUrl =
-        environment === "sandbox"
-            ? "https://www.sandbox.paypal.com/sdk/js"
-            : "https://www.paypal.com/sdk/js";
-    // env is not an allowed key for sdk/js url.
-    delete options.environment;
```
imbrian, line 77: do we still need to keep this line though? if `environment` isn't an allowed param for sdk/js URL.

matheusrufca: Should we do it in the same PR that is addressing this issue?
I'm not sure why this code if there and if we would have side effects by removing it.

imbrian: It looks like this change is removing the `delete options.environment`, which appears to have been added because it's not an allowed key. I think we only want to delete `options.sdkBaseUrl` for this change.

matheusrufca: I removed `delete options.environment` but also extracted it on line 67 so there is not need to delete the object key as this property will not be included in spreaded `options`.

imbrian: Oh yep, I see now; line 67

wsbrunson (approved):

PR #604, opened 2025-01-15
(fix) proxy function props to prevent stale state in callbacks

### LINKED ISSUE #590: [Bug] PaypalButtons Event handlers hold stale state
With react-paypal-js (`"@paypal/react-paypal-js": "^8.7.0"`), the event handlers seem to create a closure and hold old state - its not clear from the docs on how to avoid this, or update.

An example:

```jsx
function Paypal() {
  const [count, setCount] = useState(0);
  return (
    <>
      <button onClick={() => setCount(count + 1)}>Count: {count}</button>
      <PayPalButtons
        onClick={() => {
          console.log(count) // Always 0
        }}
      />
    </>
  );
}
```

Click the button a few times, state updates, but when any of the event handlers of PaypalButton are called, it's the stale state.

Seems you are creating a closure of the event handler functions. 

This is a problem because people 

1. go to my checkout, enter their info (which updates cart state)
2. click Paypal as a payment option - this renders the PaypalButtons component
3. Then they go back to their checkout and change things like coupon, name or email address.
4. They click the paypal button

So the state is stale back to when the button was initially rendered. 

Am I supposed to force a re-render of the button every time my state changes? 

I'd consider this a bug as I'd expect the onClick to just use the state, but if it's not a bug, it needs documentation as it doesn't behave as React event handlers do.

imbrian:
## context

Callback props (e.g. `createOrder`, `onApprove`) are passed to `paypal.Buttons` in a [useEffect](https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/components/PayPalButtons.tsx#L89), where the SDK Button component is then rendered. While the SDK provides a `forceReRender` prop to re-invoke the effect, which will update the callback functions passed to the SDK Button, this renders the SDK again (triggering animations, etc). The SDK Button otherwise never receives new functions as the parent component updates, creating stale closures.

## change
This PR provides a stable object from which to read the callback functions without requiring an SDK re-render.

resolves #590


wsbrunson (approved):

gregjopa (approved):
Awesome work @imbrian! 💯  I tested the new storybook demo and I see the item quantity updating and it does not cause the buttons to re-render.
Screenshot 2025-01-15 at 3 49 56 PM

PR #605, opened 2025-01-15
create new types for createOrder and creatVaultToken

### LINKED ISSUE #580: [Bug] Type and Docs not sync
### Library used

`'@paypal/paypal-js'` (for types) `script-tag`

### 🐞 Describe the Bug

Docs and types are not in sync

### 🔬 Minimal Reproduction
```ts
const cardFieldsOptions : PayPalCardFieldsComponentOptions = {
    // actions are in docs example but not within types
    //                               vvvvvv  https://developer.paypal.com/sdk/js/reference/#link-requestcaptureorderfromserver
    onApprove: (data, actions)=>{}
    // onCancel is documented but not in types
    // vvvvvvv can't link to it, because links are broken (see screenshot 1)
    onCancel: () => {}
}
const fields = paypal.CardFields.(cardFieldsOptions)
```
```ts
// example from docs https://developer.paypal.com/sdk/js/reference/#link-requestcreateorderfromserver
const onApprove = (data, actions) => {
    return fetch('/api/paypal/order/capture', {
        method: 'POST',
        body: JSON.stringify({
            orderID: data.orderID
        })
    }).then(res => {
        return res.json();
    }).then(json => {
        // Show a success page
    });
};
// this example does return a promise with data. type for onApprove is not a Promise, its just void
```

### 😕 Actual Behavior

see above

### 🤔 Expected Behavior

Not sure whats the right behavior, i guess types are right, and docs wrong?

### 🌍 Environment

-   Node.js/npm: 22 + pnpm latest
-   OS: ubuntu 24
-   Browser: chrome latest

### ➕ Additional Context

image

The sdk docs are a little broken. most links:
image

exists at least twice and always link to the first one.

### LINKED ISSUE #587: [Bug] Unable to use createVaultSetupToken with PayPalCardFieldsProvider
### Library used

react-paypal-js

### 🐞 Describe the Bug

I want to save payment methods for future use without an immediate transaction using the `PayPalCardFieldsProvider` component. This component does not allow to call only the `createVaultSetupToken` prop because the `createOrder` prop is mandatory.

I tried to specify both functions (createVaultSetupToken and createOrder) but there is an error shown in the console.

Error in console when trying to use `createVaultSetupToken` with `PayPalCardFieldsProvider` component even though its available in the provided interface.

### 🔬 Minimal Reproduction

Add a `PayPalCardFieldsProvider` component and try to add the `createVaultSetupToken` prop without `createOrder` or try to add both `createVaultSetupToken` and `createOrder`.

### 😕 Actual Behavior

- When trying to specify only `createVaultSetupToken` : 
image

- **When trying to specify both `createVaultSetupToken`  and `createOrder`:** Error is thrown during runtime : `Uncaught Error: Cannot pass both createVaultSetupToken and createOrder`
image

### 🤔 Expected Behavior

Being able to use the `createVaultSetupToken` prop in `PayPalCardFieldsProvider` component without having to specify a `createOrder` prop.

### 🌍 Environment

-   Node.js/npm: - 20.14.0
-   OS: - Windows 11
-   Browser: - Edge

hnguyen510:
This pr is to fix this [LI-73058](https://paypal.atlassian.net/browse/LI-73058)

Problem: the createOrder callback is require when createVaultToken is implement when it should be one or the other and not both. 

Closes #587
Closes #580

### DIFF
--- a/.changeset/silly-bears-boil.md
+++ b/.changeset/silly-bears-boil.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": major
+---
+
+create new type for createOrder, createVaultToken and add onCancel to card-field
--- a/.changeset/silly-bears-boil.md
+++ b/.changeset/silly-bears-boil.md
@@ -1,5 +1,0 @@
----
-"@paypal/paypal-js": major
----
-
-create new type for createOrder, createVaultToken and add onCancel to card-field
--- a/packages/paypal-js/types/components/card-fields.d.ts
+++ b/packages/paypal-js/types/components/card-fields.d.ts
@@ -127,16 +127,31 @@ export interface PayPalCardFieldsIndividualField {
     close: () => Promise<void>;
 }
 
-export interface PayPalCardFieldsComponentOptions {
-    createOrder: () => Promise<string>;
+export interface PayPalCardFieldsComponentBasics {
     onApprove: (data: CardFieldsOnApproveData) => void;
     onError: (err: Record<string, unknown>) => void;
-    createVaultSetupToken?: () => Promise<string>;
     inputEvents?: PayPalCardFieldsInputEvents;
     style?: Record<string, PayPalCardFieldsStyleOptions>;
 }
 
-export interface PayPalCardFieldsComponent {
+export interface PayPalCardFieldsComponentCreateOrder
+    extends PayPalCardFieldsComponentBasics {
+    createOrder: () => Promise<string>;
+    createVaultSetupToken?: never;
+}
+
+export interface PayPalCardFieldsComponentCreateVaultSetupToken
+    extends PayPalCardFieldsComponentBasics {
+    createOrder?: never;
+    createVaultSetupToken: () => Promise<string>;
+}
+
+export type PayPalCardFieldsComponentOptions =
+    | PayPalCardFieldsComponentCreateOrder
+    | PayPalCardFieldsComponentCreateVaultSetupToken;
+
+export interface PayPalCardFieldsComponent
+    extends PayPalCardFieldsComponentOptions {
     getState: () => Promise<PayPalCardFieldsStateObject>;
     isEligible: () => boolean;
     submit: () => Promise<void>;
--- a/packages/paypal-js/types/components/card-fields.d.ts
+++ b/packages/paypal-js/types/components/card-fields.d.ts
@@ -130,6 +130,7 @@ export interface PayPalCardFieldsIndividualField {
 export interface PayPalCardFieldsComponentBasics {
     onApprove: (data: CardFieldsOnApproveData) => void;
     onError: (err: Record<string, unknown>) => void;
+    onCancel?: () => Promise<void> | void;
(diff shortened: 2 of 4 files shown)


packages/paypal-js/types/components/card-fields.d.ts
```diff
@@ -127,16 +127,31 @@ export interface PayPalCardFieldsIndividualField {
...
-export interface PayPalCardFieldsComponent {
+export interface PayPalCardFieldsComponentCreateOrder
+    extends PayPalCardFieldsComponentBasics {
+    createOrder: () => Promise<string>;
+    createVaultSetupToken?: never;
+}
+
+export interface PayPalCardFieldsComponentCreateVaultSetupToken
+    extends PayPalCardFieldsComponentBasics {
+    createOrder?: never;
```
ravishekhar, line 146: lets revisit if we need  createOrder?: never;

ravishekhar (approved):
Looks good overall.

wsbrunson:
@hnguyen510 can you also add `onCancel` to Card Fields props? https://github.com/paypal/paypal-js/issues/580

packages/paypal-js/types/components/card-fields.d.ts
hnguyen510: I am not sure if `createOrder?: never` add any value, so i feel like we can remove that

packages/paypal-js/types/tests/card-fields.test.ts
```diff
@@ -0,0 +1,48 @@
...
+            {
+                createVaultSetupToken: vi.fn(),
+                onApprove: vi.fn(),
+                onError: vi.fn(),
+            };
+
+        expect(createVaultSetupTokenCallback.createOrder).toBeUndefined();
+    });
+
+    test.skip("Can't have both", () => {
```
imbrian: Are these test still supposed to be skipped?

packages/paypal-js/types/components/card-fields.d.ts
imbrian: There might be value in future clarity.

.changeset/silly-bears-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/paypal-js": major
```
wsbrunson: this is probably a `minor` change since its not breaking any existing usage of the sdk

packages/paypal-js/types/tests/card-fields.test.ts
hnguyen510: It still need to be skip right now because there is assertionError. Not sure what assertion call I should do instead.

.changeset/silly-bears-boil.md
hnguyen510: I updated the changeset, how do i remove the old one?

wsbrunson: you can delete this file

ravishekhar (approved):

PR #622, opened 2025-02-08
chore: update `repository` field in `package.json` files

andipaetzold:
Two small changes:
- The `repository.url` of `react-paypal-js` was still pointing to the old repository. That URL is used for the "Homepage" link on npm. 
- Added `repository.directory` to point to the correct folder within this repository


wsbrunson (approved):
thank you!

PR #636, opened 2025-04-10
Upgrade/react bt sdk version

delbarrow:
The BT SDK version is statically typed to an old version in our React SDK. As a short-term fix, we need to upgrade this version to the latest statically. The BT SDK was updated to 3.117.0 in the React SDK.

Jira ticket: [https://paypal.atlassian.net/browse/DTPPCPSDK-3082](url)

### DIFF
--- a/.changeset/afraid-eyes-impress.md
+++ b/.changeset/afraid-eyes-impress.md
@@ -1,0 +1,7 @@
+---
+"@paypal/react-paypal-js": major
+---
+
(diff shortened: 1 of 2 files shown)


gregjopa (approved):

.changeset/afraid-eyes-impress.md
```diff
@@ -0,0 +1,7 @@
+---
+"@paypal/react-paypal-js": major
```
imbrian: we should make this a patch version. the cli is a little confusing the first time

imbrian (changes requested):

imbrian (approved):

PR #659, opened 2025-08-20
Feature: Add V6 Types

EvanReinstein:
This adds a set of Typescript types that support the PayPal V6 JS SDK, limited to PayPal and Venmo One Time Payments.


packages/paypal-js/package.json
```diff
@@ -3,11 +3,22 @@
...
     "publishConfig": {
         "access": "public"
     },
+    "exports": {
+        ".": {
+            "types": "./types/index.d.ts",
+            "import": "./dist/v5/esm/paypal-js.js",
+            "require": "./dist/v5/cjs/paypal-js.js"
+        },
+        "./v6": {
```
gregjopa: Thoughts on renaming this so its clear that its referring to the v6 web sdk?
```suggestion
        "./sdk-v6": {
```

EvanReinstein: Yea, all good with this 👍

packages/paypal-js/README.md
```diff
@@ -293,6 +293,25 @@ loadCustomScript({
...
+    SdkInstance,
+    PaymentSessionInputs,
+    OnApproveData,
+    OnCompleteData,
+    CreateInstanceInputs,
+    Components,
+    EligiblePaymentMethods
+} from "@paypal/paypal-js/types/v6";
+```
+
```
EvanReinstein: @gregjopa I'll update this path but figured I'd bring your attention to the README, lmk if you think this can/should be enhanced.

packages/paypal-js/README.md
```diff
@@ -293,6 +293,25 @@ loadCustomScript({
 
 This package includes TypeScript type definitions for the PayPal JS SDK. This includes types for the `window.paypal` namespace. We support projects using TypeScript versions >= 3.8.
 
+### V6 TypeScript Types
+
+The library includes TypeScript types for the PayPal JS SDK V6 API. These types provide full type safety when working with the V6 payment flows.
```
kand: Do we usually refer to our library as an API or SDK?

kand (approved):
One small nitpick to take or leave; otherwise looks good!

packages/paypal-js/src/index-v6.ts
```diff
@@ -0,0 +1,4 @@
+import { loadCustomScript } from "./load-script";
```
hamzanasir: Would it be better to nest this file in a v6 folder? So possibly this:

src -> v6 -> index.ts

This would allow us to potentially add v6 specific code to this directory and barrel it with this index file

EvanReinstein: Great call, will do 👍

packages/paypal-js/e2e-tests/reload-script.test.ts
```diff
@@ -14,13 +14,15 @@ test("Reload script", async ({ page }) => {
...
+            request.url().startsWith("https://www.paypal.com/sdk/js") &&
+            request.url().includes("currency=EUR"),
+    );
 
     await page.locator("select#currency").selectOption("EUR");
-    expect(sdkRequest.includes("currency=EUR")).toBe(true);
+
+    const sdkRequest = await requestPromise;
+    expect(sdkRequest.url().includes("currency=EUR")).toBe(true);
 });
```
EvanReinstein, line 28: I updated this b/c I kept getting `reload-script` test failures for chrome and firefox. I checked on the `main` branch locally and got the same failures so decided to keep this part of the `reload-script.test.ts` update from the previous PR.

packages/paypal-js/types/v6/components/payments.d.ts
```diff
@@ -0,0 +1,101 @@
...
+    ) => Promise<void>;
+    onShippingOptionsChange?: (
+        data: OnShippingOptionsChangeData,
+    ) => Promise<void>;
+    savePayment?: boolean;
+};
+
+export type SavePaymentSessionInputs = {
+    testBuyerCountry?: string;
+    commit?: boolean;
```
gregjopa: This flag does not exist for the save payment flows.
```suggestion
```

packages/paypal-js/types/v6/components/payments.d.ts
```diff
@@ -0,0 +1,101 @@
...
+export type ShippingOptionsErrorMessages = Record<
+    ShippingOptionsErrorType,
+    string
+>;
+
+export type OnShippingAddressChangeData = {
+    errors: ShippingAddressErrorMessages;
+    orderId: string;
+    shippingAddress: {
+        city: string;
```
gregjopa: Technically, countryCode is the only required field and `city`, `postalCode`, and `state` are all optional based on all the different types of addresses around the world.

gregjopa:
```suggestion
        city?: string;
```

packages/paypal-js/types/v6/components/payments.d.ts
```diff
@@ -0,0 +1,101 @@
...
+        countryCode: string;
+        postalCode: string;
```
gregjopa:
```suggestion
        postalCode?: string;
```

packages/paypal-js/types/v6/components/payments.d.ts
```diff
@@ -0,0 +1,101 @@
...
+        state: string;
```
gregjopa:
```suggestion
        state?: string;
```

packages/paypal-js/types/v6/components/billing.d.ts
```diff
@@ -0,0 +1,14 @@
+import {
+    OnApproveData,
+    OnCompleteData,
+    OnCancelData,
+    OnErrorData,
+} from "./payments";
+
+export type BillingSessionInputs = {
+    onApprove?: (data: OnApproveData) => Promise<void>;
```
gregjopa: I just checked and technically `orderId`, `billingToken`, and `payerId` are what get returned for this legacy billing agreement without purchase flow.
```suggestion
    onApprove?: (data: OnApproveDataOneTimePayments) => Promise<void>;
```

EvanReinstein: Ah! Thanks for catching this!

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,137 @@
...
+    cancel: () => void;
+};
+
+export type BillingSessionOutput = {
+    start: (
+        options: StartSessionInput,
+        billingSessionOptionsPromise: Promise<{ billingToken: string }>,
+    ) => Promise<void>;
+    destroy: () => void;
+};
```
EvanReinstein: Noticed that `BillingAgreementWithoutPurchaseSession` does not have a `cancel` method.

packages/paypal-js/README.md
```diff
@@ -293,6 +293,25 @@ loadCustomScript({
...
+    SdkInstance,
+    PaymentSessionInputs,
+    OnApproveData,
+    OnCompleteData,
+    CreateInstanceInputs,
+    Components,
+    EligiblePaymentMethods
+} from "@paypal/paypal-js/types/sdk-v6";
+```
+
```
gregjopa, line 334: Can we update this example showing how someone would use `window.paypal.createInstance()` with these types?

gregjopa: Here's an idea for the readme. I'm thinking that most people only need `PayPalV6Namespace` and the types for their callback functions for knowing what `data` is.
```ts
import type {
    PayPalV6Namespace,
    CreateInstanceInputs,
    OnApproveDataOneTimePayments,
    OnShippingAddressChangeData
} from "@paypal/paypal-js/types/sdk-v6";


declare global {
    interface Window {
        paypal?: PayPalV6Namespace | null;
    }
}

const createInstanceInputs: CreateInstanceInputs = {
    clientToken,
    components: ["paypal-payments", "venmo-payments"]
}

const sdkInstance = await window.paypal.createInstance(createInstanceInputs);
```

EvanReinstein: Thanks @gregjopa for this suggestion, I'll enhance the README accordingly.

packages/paypal-js/types/v6/components/billing.d.ts
```diff
@@ -0,0 +1,14 @@
+import {
+    OnCompleteData,
+    OnCancelData,
+    OnErrorData,
+    OnApproveDataOneTimePayments,
+} from "./payments";
+
+export type BillingSessionInputs = {
+    onApprove?: (data: OnApproveDataOneTimePayments) => Promise<void>;
```
EvanReinstein: Could create an alias for this, that is more legacy billing agreement-centric @EvanReinstein

packages/paypal-js/types/v6/components/payments.d.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export type OnCancelData = {
+    orderId: string;
+};
+
+export type OnErrorData = Error;
+
+export type PaymentSessionInputs = {
+    onApprove?: (data: OnApproveDataOneTimePayments) => Promise<void>;
+    onCancel?: (data?: { orderId: string }) => void;
```
EvanReinstein: update with onCancelData @EvanReinstein

packages/paypal-js/types/v6/components/payments.d.ts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+export type OnApproveDataSavePayments = {
+    vaultSetupToken: string;
+    payerId?: string;
+};
+
+export type OnApproveData =
+    | OnApproveDataOneTimePayments
+    | OnApproveDataSavePayments;
```
EvanReinstein: remove export @EvanReinstein

packages/paypal-js/types/v6/components/paypal-legacy-billing-agreemens.ts
```diff
@@ -0,0 +1,14 @@
+import {
+    OnCompleteData,
+    OnCancelData,
+    OnErrorData,
+    OnApproveDataOneTimePayments,
+} from "./paypal-payments";
+
+export type BillingSessionInputs = {
```
bywood: We seem to have a mix of `Inputs` and `Options` in our naming. I think I'd prefer to standardize on `Options`. I think that's how we typically refer to these types of objects.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,138 @@
...
+    findEligibleMethods: (
+        findEligibleMethodsOptions: FindEligibleMethodsInputs,
+    ) => Promise<EligiblePaymentMethodsOutput>;
+};
+
+export type FindEligibleMethodsInputs = {
+    currencyCode?: string;
+};
+
+export type SessionOutput = {
```
bywood: Since these `create...PaymentSession` methods create and return instances of payment session classes, could we call this `PaymentSession` instead?

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,138 @@
...
+    | "modal"
+    | "payment-handler"
+    | "popup"
+    | "redirect";
+
+export type VenmoPresentationModes = "auto" | "modal" | "popup";
+
+export type SessionStartOptions = {
+    presentationMode?: PayPalPresentationModes | VenmoPresentationModes;
+    targetElement?: string | EventTarget;
```
bywood: Not sure about `EventTarget` here. Do we want it to be that broad? Why not `HTMLElement`?
```
EventTarget <- Node <- Element <- HTMLElement
```

packages/paypal-js/README.md
```diff
@@ -293,6 +293,42 @@ loadCustomScript({
...
+In addition to the type definitions above, this package also includes Typescript type definitions for the PayPal JS SDK V6.
+
+#### Importing and Using V6 Types
+
+A basic example showing data-typing for PayPal One Time Payment.
+
+```js
+import type {
+    PayPalV6Namespace,
+    CreateInstanceInputs,
```
bywood: Looks like you addressed the `Inputs` vs `Options` while I've been reviewing, but this file still needs to be updated to use `CreateInstanceOptions`.

packages/paypal-js/types/v6/index.d.ts
EvanReinstein: Yes! I lifted this from the v6 example and didn't consider renaming. But this makes sense 👍

packages/paypal-js/types/v6/index.d.ts
EvanReinstein: Will do 👍

packages/paypal-js/README.md
EvanReinstein: Thanks for your 🦅 👁️ @bywood !

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,138 @@
...
+    | "paypal_credit"
+    | "paypal"
+    | "venmo";
+
+export interface EligiblePaymentMethodsOutput {
+    isEligible: (paymentMethod: EligiblePaymentMethods) => boolean;
+}
+
+export type SdkInstance = {
+    // "paypal-legacy-billing-agreements" component
```
gregjopa: Is this comment needed?
```suggestion
```

gregjopa: If so, I think we could add the component name to the deprecated comment below.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export type EligiblePaymentMethods =
+    | "paypal_pay_later"
+    | "paypal_credit"
+    | "paypal"
+    | "venmo";
+
+export interface EligiblePaymentMethodsOutput {
+    isEligible: (paymentMethod: EligiblePaymentMethods) => boolean;
+}
```
gregjopa, line 108: Can we add the `getDetails()` here? That is required for paylater and credit to work. Ex: https://github.com/paypal-examples/v6-web-sdk-sample-integration/blob/main/client/components/paypalPayments/oneTimePayment/html/src/recommended/app.js#L19-L20

packages/paypal-js/types/v6/index.d.ts
gregjopa: ahh I see why the comment is there. Each component has a comment like this. Please ignore. I need more coffee

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,138 @@
...
+    findEligibleMethods: (
+        findEligibleMethodsOptions: FindEligibleMethodsOptions,
+    ) => Promise<EligiblePaymentMethodsOutput>;
+};
+
+export type FindEligibleMethodsOptions = {
+    currencyCode?: string;
+};
+
+export type PaymentSession = {
```
gregjopa: I think we need these to be payment flow specific too.
```suggestion
export type PaymentSessionOneTimePayment = {
```

For save payment flow, we will have:
```ts
export type PaymentSessionSavePayment = {
    start: (
        options: SessionStartOptions,
        vaultSetupTokenPromise: Promise<{ vaultSetupToken: string }>,
    ) => Promise<void>;
    destroy: () => void;
    cancel: () => void;
};
```

gregjopa: same applies to the legacy billing agreements too:
```ts
export type PaymentSessionLegacyBillingAgreement = {
    start: (
        options: SessionStartOptions,
        billingTokenPromise: Promise<{ billingToken: string }>,
    ) => Promise<void>;
    destroy: () => void;
    cancel: () => void;
};
```

packages/paypal-js/types/v6/index.d.ts
EvanReinstein: Do you have any feelings related to adding all of the funding source strings?

EvanReinstein: I should clarify, is there any reason why we don't want all of the current values included in the funding source type also included here?

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,148 @@
...
+    findEligibleMethods: (
+        findEligibleMethodsOptions: FindEligibleMethodsOptions,
+    ) => Promise<EligiblePaymentMethodsOutput>;
+};
+
+export type FindEligibleMethodsOptions = {
+    currencyCode?: string;
+};
+
+export type PaymentSessionOneTimePayment = {
```
bywood: I'd recommend having these reflect the names used in the methods that create them:
```suggestion
export type OneTimePaymentSession = {
```

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,148 @@
...
+    start: (
+        options: SessionStartOptions,
+        orderIdPromise: Promise<{ orderId: string }>,
+    ) => Promise<void>;
+    destroy: () => void;
+    cancel: () => void;
+};
+
+export type PaymentSessionSavePayment = {
```
bywood: Same as above:
```suggestion
export type SavePaymentSession = {
```

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,148 @@
...
+    start: (
+        options: SessionStartOptions,
+        vaultSetupTokenPromise: Promise<{ vaultSetupToken: string }>,
+    ) => Promise<void>;
+    destroy: () => void;
+    cancel: () => void;
+};
+
+export type PaymentSessionLegacyBillingAgreement = {
```
bywood: Not sure about this name. I thought we were moving away from calling it "Legacy"? Also, code-wise, this is a `BillingSession`, but maybe that could be considered a type of `PaymentSession`? Not sure.

packages/paypal-js/types/v6/index.d.ts
gregjopa: I like that suggestion @bywood 💯

packages/paypal-js/types/v6/index.d.ts
EvanReinstein: Good points, I went back and forth on this. I'll update to remove the legacy part.

bywood: Maybe `BillingAgreementPaymentSession`? I don't know.

EvanReinstein: Gah! I went with `BillingAgreementSession`, I can update if we want to align on `xPaymentSession`. I don't see any harm in including it.

packages/paypal-js/README.md
```diff
@@ -293,6 +293,42 @@ loadCustomScript({
...
+        paypal?: PayPalV6Namespace | null;
+    }
+}
+
+const createInstanceOptions: CreateInstanceOptions = {
+    clientToken,
+    components: ["paypal-payments", "venmo-payments"]
+}
+
+const sdkInstance: SdkInstance = await window.paypal.createInstance(createInstanceInputs);
```
gregjopa: Technically this return type should automatically be applied.
```suggestion
const sdkInstance = await window.paypal.createInstance(createInstanceInputs);
```

packages/paypal-js/README.md
```diff
@@ -293,6 +293,42 @@ loadCustomScript({
...
+
+const paypalCheckout = sdkInstance.createPayPalOneTimePaymentSession({
```
gregjopa, line 329: Most merchants will define these callback functions separately since they can get pretty long. Can we update the example to reflect that?
```ts
function onApproveCallback(data: OnApproveDataOneTimePayments) {
  // merchant code for capturing or authorizing the order
}
function onShippingAddressChangeCallback(data: OnShippingAddressChangeData) {
  // merchant code for validating the shipping address
  // and patching the order with the new shipping cost
}
```
The types are needed when the above approach is used. When the functions are used inline with the `createPayPalOneTimePaymentSession` arguments, they should be automatically types correctly for what `data` is.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export type VenmoSessionStartOptions = SessionStartOptions & {
+    eligibilityResponse?: VenmoEligibility | null;
+    buttonSessionID?: string;
+};
+
+export function loadCustomScript(options: {
+    url: string;
+    attributes?: Record<string, string>;
+    PromisePonyfill?: PromiseConstructor;
```
gregjopa: If possible, can we drop this from the v6 version of `loadCustomScript()`? The v5 sdk was designed to work in older browsers that did not support Promises. The v6 sdk browser support strategy requires that Promises and even async/await are supported by the browser natively.
```suggestion
```

EvanReinstein: Sounds good, I'll remove and we can test it out 👍

EvanReinstein: Ended up exporting the method with a simple wrapper. I think that is probably the best bet without changing the original function.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,252 @@
...
+    | "home"
+    | "mini-cart"
+    | "product-details"
+    | "product-listing"
+    | "search-results";
+
+export type CreateInstanceOptions<T extends readonly Components[]> = {
+    clientMetadataId?: string;
+    clientToken: string;
+    components: T;
```
EvanReinstein, line 37: @hamzanasir spent a second to update the `createInstanceOptions` typing on line 16, very similar to what we had in office hours yesterday 👍

packages/paypal-js/types/v6/components/paypal-payments.d.ts
```diff
@@ -0,0 +1,110 @@
...
+    onShippingOptionsChange?: (
+        data: OnShippingOptionsChangeData,
+    ) => Promise<void>;
+    savePayment?: boolean;
+};
+
+export type SavePaymentSessionOptions = BaseSessionOptions & {
+    clientMetadataId?: string;
+    orderId?: never;
+    setupToken?: string;
```
gregjopa:
```suggestion
    vaultSetupToken?: string;
```

packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
```diff
@@ -0,0 +1,75 @@
...
+    | "paypal_pay_later"
+    | "paypal_credit"
+    | "paypal"
+    | "venmo";
+
+export type FindEligibleMethodsOptions = {
+    currencyCode?: string;
+};
+
+export type FundingSource =
```
gregjopa: can we reduce this list to just what we support today with this first round of types? 

- paypal
- paylater
- credit
- venmo

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -0,0 +1,194 @@
...
+        : Record<string, never>) &
+    (T[number] extends "venmo-payments"
+        ? VenmoPaymentsInstance
+        : Record<string, never>) &
+    (T[number] extends "paypal-legacy-billing-agreements"
+        ? PayPalLegacyBillingInstance
+        : Record<string, never>);
+
+export interface BaseInstance {
+    findEligibleMethods: (
```
gregjopa, line 89: Can we add `updateLocale(locale: string)` to this baseInstance?

gregjopa (approved):
Great work with this @EvanReinstein! Happy to see it get merged.

PR #668, opened 2025-09-02
feat: Add `loadCoreSdkScript` script loader for v6 sdk

gregjopa:
This PR adds the `loadCoreSdkScript()` script loader function for v6. Ex:
```ts
import {
  loadCoreSdkScript
} from "@paypal/paypal-js/sdk-v6";

const paypalNamespaceReference = await loadCoreSdkScript({ environment: "sandbox", debug: true });
const sdkInstance = await paypalNamespaceReference.createInstance({ 
  clientToken, 
  components: ["paypal-payments"] 
});
```

The `loadCustomScript()` function was removed and replaced with this new `loadCoreSdkScript()` function for improved readability.

### DIFF
--- a/packages/paypal-js/src/v6/index.ts
+++ b/packages/paypal-js/src/v6/index.ts
@@ -1,4 +1,5 @@
 import { insertScriptElement } from "../utils";
+import type { PayPalV6Namespace } from "../../types/v6/index";
 
 const version = "__VERSION__";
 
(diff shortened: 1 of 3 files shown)


packages/paypal-js/src/v6/index.ts
```diff
@@ -1,13 +1,64 @@
...
+    validateArguments(options);
+    const { environment, debug } = options;
+
+    // resolve with null when running in Node or Deno
+    if (typeof document === "undefined") return Promise.resolve(null);
+
+    const baseURL =
+        environment === "sandbox"
+            ? "https://www.sandbox.paypal.com"
+            : "https://www.paypal.com";
```
bywood: Would it be safer to default to `sandbox`?
```suggestion
    const baseURL =
        environment === "production"
            ? "https://www.paypal.com"
            : "https://www.sandbox.paypal.com";
```

gregjopa: Nice suggestion @bywood. I agree, let's default to sandbox instead. I updated the PR with this change.

PR #681, opened 2025-09-17
Feat: Initial v6 configuration

EvanReinstein:
This PR currently contains the very initial updates to add v6-centric React components.


packages/react-paypal-js/package.json
EvanReinstein, line 1: Updated the `@rollup/plugin-typescript` and `typescript` package versions to match `paypal-js`.

packages/react-paypal-js/tsconfig.json
```diff
@@ -6,7 +6,7 @@
         "forceConsistentCasingInFileNames": true,
         "jsx": "react",
         "module": "esnext",
-        "moduleResolution": "node",
+        "moduleResolution": "bundler",
```
EvanReinstein, line 10: This update is causing some issues when I run `npm run type-declarations`. Will investigate further.

PR #688, opened 2025-09-26
Feat: V6 Context Provider

EvanReinstein:
This PR adds a Provider wrapper that can be used to access SDK instance context. This is an initial set of basic changes that leverage `paypal-js`'s `loadCoreSdkScript` and adds SSR related conditions to prevent/guard against running client-side code.

Testing instructions coming soon 👀 🧪

### DIFF
--- a/packages/paypal-js/src/utils.ts
+++ b/packages/paypal-js/src/utils.ts
@@ -158,3 +158,5 @@ function createScriptElement(
 
     return newScript;
 }
+
+export const isServer = typeof window === "undefined";
--- a/packages/paypal-js/src/v6/index.ts
+++ b/packages/paypal-js/src/v6/index.ts
@@ -8,6 +8,12 @@ const version = "__VERSION__";
 
 function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
     validateArguments(options);
+
+    // SSR safeguard
+    if (typeof document === "undefined") {
+        return Promise.resolve(null);
+    }
+
     const { environment, debug } = options;
 
     const baseURL =
@@ -24,7 +30,7 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
         insertScriptElement({
             url: url.toString(),
             onSuccess: () => {
-                if (!window.paypal) {
+                if (typeof window === "undefined" || !window.paypal) {
                     return reject(
                         "The window.paypal global variable is not available",
                     );
--- a/packages/paypal-js/src/v6/index.ts
+++ b/packages/paypal-js/src/v6/index.ts
@@ -1,4 +1,4 @@
-import { insertScriptElement } from "../utils";
+import { insertScriptElement, isServer } from "../utils";
 import type {
     PayPalV6Namespace,
     LoadCoreSdkScriptOptions,
@@ -30,7 +30,7 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
         insertScriptElement({
             url: url.toString(),
             onSuccess: () => {
-                if (typeof window === "undefined" || !window.paypal) {
+                if (isServer || !window.paypal) {
                     return reject(
                         "The window.paypal global variable is not available",
                     );
--- a/packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
@@ -1,0 +1,117 @@
+import React, { useEffect, useState } from "react";
+import { loadCoreSdkScript } from "@paypal/paypal-js/sdk-v6";
+
+import { InstanceContext } from "../context/InstanceProviderContext";
+
+import type {
+    CreateInstanceOptions,
+    Components,
+    SdkInstance,
+    EligiblePaymentMethodsOutput,
+    LoadCoreSdkScriptOptions,
+} from "../types";
+import type { FC } from "react";
+
+interface PayPalInstanceProviderProps {
+    options: CreateInstanceOptions<readonly [Components, ...Components[]]>;
+    children: React.ReactNode;
+    scriptOptions: LoadCoreSdkScriptOptions;
+}
+
+export const PayPalInstanceProvider: FC<PayPalInstanceProviderProps> = ({
+    options,
+    children,
+    scriptOptions,
+}: {
+    options: CreateInstanceOptions<readonly [Components, ...Components[]]>;
+    children: React.ReactNode;
+    scriptOptions: LoadCoreSdkScriptOptions;
+}) => {
+    const [sdkInstance, setSdkInstance] = useState<SdkInstance<
+        readonly [Components, ...Components[]]
+    > | null>(null);
+    const [eligiblePaymentMethods, setEligiblePaymentMethods] =
+        useState<EligiblePaymentMethodsOutput | null>(null);
+    const [isLoading, setIsLoading] = useState(true);
+    const [error, setError] = useState<Error | null>(null);
+
+    // use a useEffect to load the core sdk script
+    // ensure that we take SSR precautions
+    useEffect(() => {
+        // Skip if SSR
+        if (typeof window === "undefined") {
+            setIsLoading(false);
+            return;
+        }
+
+        const controller = new AbortController();
+
+        const initializeSdk = async () => {
+            try {
+                setIsLoading(true);
+                setError(null);
+
+                const paypalNameSapce = await loadCoreSdkScript(scriptOptions);
+
+                if (controller.signal.aborted) {
+                    return;
+                }
+
+                if (!paypalNameSapce) {
+                    throw new Error("PayPal SDK failed to load");
+                }
+
+                const instance = await paypalNameSapce.createInstance(options);
+
+                if (controller.signal.aborted) {
+                    return;
+                }
+
+                setSdkInstance(instance);
+
+                try {
+                    const eligiblePaymentMethods =
+                        await instance.findEligibleMethods({});
+
+                    if (controller.signal.aborted) {
+                        return;
+                    }
+
+                    setEligiblePaymentMethods(eligiblePaymentMethods);
+                } catch (eligibilityErr) {
+                    if (!controller.signal.aborted) {
+                        console.warn(
+                            "Failed to get eligible payment methods:",
+                            eligibilityErr,
+                        );
+                    }
+                }
+
+                if (!controller.signal.aborted) {
+                    setIsLoading(false);
+                }
+            } catch (err) {
+                if (!controller.signal.aborted) {
+                    const error =
+                        err instanceof Error ? err : new Error(String(err));
+                    setError(error);
+                    setIsLoading(false);
+                }
+            }
+        };
+
+        initializeSdk();
+
+        return () => {
+            controller.abort();
+        };
+    }, [options, scriptOptions]);
+
+    return (
+        <InstanceContext.Provider
+            value={{ sdkInstance, eligiblePaymentMethods, isLoading, error }}
+        >
+            {children}
+        </InstanceContext.Provider>
+    );
+};
--- a/packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 import { loadCoreSdkScript } from "@paypal/paypal-js/sdk-v6";
 
 import { InstanceContext } from "../context/InstanceProviderContext";
+import { isServer } from "../../utils";
 
 import type {
     CreateInstanceOptions,
@@ -10,7 +11,6 @@ import type {
     EligiblePaymentMethodsOutput,
     LoadCoreSdkScriptOptions,
 } from "../types";
-import type { FC } from "react";
 
 interface PayPalInstanceProviderProps {
     options: CreateInstanceOptions<readonly [Components, ...Components[]]>;
@@ -18,7 +18,7 @@ interface PayPalInstanceProviderProps {
     scriptOptions: LoadCoreSdkScriptOptions;
 }
 
-export const PayPalInstanceProvider: FC<PayPalInstanceProviderProps> = ({
+export const PayPalInstanceProvider: React.FC<PayPalInstanceProviderProps> = ({
     options,
     children,
     scriptOptions,
@@ -39,7 +39,7 @@ export const PayPalInstanceProvider: FC<PayPalInstanceProviderProps> = ({
     // ensure that we take SSR precautions
     useEffect(() => {
         // Skip if SSR
-        if (typeof window === "undefined") {
+        if (isServer) {
             setIsLoading(false);
             return;
         }
--- a/packages/paypal-js/types/v6/index.d.ts
+++ b/packages/paypal-js/types/v6/index.d.ts
@@ -190,7 +190,7 @@ export type LoadCoreSdkScriptOptions = {
 
 export function loadCoreSdkScript(
     options: LoadCoreSdkScriptOptions,
-): Promise<PayPalV6Namespace>;
+): Promise<PayPalV6Namespace | null>;
 
 // Components
 export * from "./components/paypal-payments";
--- a/packages/react-paypal-js/src/utils.ts
+++ b/packages/react-paypal-js/src/utils.ts
@@ -95,3 +95,5 @@ export function generateErrorMessage({
 
     return errorMessage;
 }
+
+export const isServer = typeof window === "undefined";
--- a/packages/react-paypal-js/src/v6/context/InstanceProviderContext.tsx
+++ b/packages/react-paypal-js/src/v6/context/InstanceProviderContext.tsx
@@ -1,0 +1,5 @@
+import { createContext } from "react";
+
+import { InstanceContextState } from "../types/InstanceProviderTypes";
+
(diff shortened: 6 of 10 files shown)


packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+    // use a useEffect to load the core sdk script
+    // ensure that we take SSR precautions
+    useEffect(() => {
+        // Skip if SSR
+        if (isServer) {
+            setIsLoading(false);
+            return;
+        }
+
+        const controller = new AbortController();
```
EvanReinstein: This is one strategy used to clean up this component **if** for some reason a user navigates to a page/screen that triggers this Provider but then navigates away. Without some sort of "cancel" indicator the code in the useEffect may continue to run, even if the component has been unmounted, resulting in console errors/warnings.

packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
```diff
@@ -0,0 +1,117 @@
...
+    LoadCoreSdkScriptOptions,
+} from "../types";
+
+interface PayPalInstanceProviderProps {
+    options: CreateInstanceOptions<readonly [Components, ...Components[]]>;
+    children: React.ReactNode;
+    scriptOptions: LoadCoreSdkScriptOptions;
+}
+
+export const PayPalInstanceProvider: React.FC<PayPalInstanceProviderProps> = ({
```
EvanReinstein: In a fit of spontaneity I used "instance" as a key term in the names for the provider and related files, context state, etc. 

I am not attached to this and am willing to make any changes. Originally I was calling this `v6PayPalScriptProvider` but felt that the provider was really providing the SDK instance. Any thoughts here would be appreciated.

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein, line 1: Added a v6-specific utils file. Only the folders/files within the `src/v6/` path get built for V6 and importing utils from the V5 utils file was causing some issues. There is probably a workaround but this was the most straightforward solution.

packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
kand: I think this name makes sense, but what about adding `sdk` to make it clear what PayPal thing is being provided, i.e. `PayPalSdkInstanceProvider`?

packages/paypal-js/src/v6/index.ts
```diff
@@ -8,6 +8,12 @@ const version = "__VERSION__";
 
 function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
     validateArguments(options);
+
+    // SSR safeguard
+    if (typeof document === "undefined") {
```
kand: Makes sense to me, but is this the standard way to check for SSR? I don't have much SSR experience so I'm curious.

Also, what's the difference between this check and `isServer` checking if `typeof window === "undefined"`? Would it make sense to use the same strategy in both places?

EvanReinstein: Shoot I meant to replace this with `isServer`.

I want to standardize on one `isServer` check, which currently is `typeof window === "undefined"`. From looking at a few different packages (stripe, square, adyen-web) there are a few ways to do this, and `typeof window === "undefined"` is just one of those.

packages/react-paypal-js/src/v6/components/PayPalInstanceProvider.tsx
EvanReinstein: I like that @kand I'll update and see what other folks have to say.

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+        if (
+            !isServer &&
+            state.loadingStatus === INSTANCE_LOADING_STATE.INITIAL
+        ) {
+            dispatch({
+                type: INSTANCE_DISPATCH_ACTION.SET_LOADING_STATUS,
+                value: INSTANCE_LOADING_STATE.PENDING,
+            });
+        }
+    }, []); // Run once on mount
```
EvanReinstein: Purposefully using an empty array here in order to run this effect only on mount.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+            controller.abort();
+        };
+    }, [
+        state.loadingStatus,
+        state.sdkInstance,
+        state.createInstanceOptions,
+        state.scriptOptions,
+    ]);
+
+    // Separate effect for eligibility - runs after instance is created
```
EvanReinstein, line 163: Currently going with calling eligibility for all use cases. Originally we'd discussed passing the eligibility response to the provider and then conditionally calling eligibility. I'd like to update this portion of the provider when we implement the eligibility hook.

packages/paypal-js/src/utils.ts
```diff
@@ -158,3 +158,5 @@ function createScriptElement(
 
     return newScript;
 }
+
+export const isServer = typeof window === "undefined";
```
gregjopa: Two minor things:

1. should we make this a function so its callable in different contexts?
2. Can we also check `document`? When doing server-side rendering, I think some environments may set `window` so its safer to check `document`. Might as well check for both being defined.

```suggestion
export function isServer() { 
  return typeof window === "undefined" && typeof document === "undefined";
}
```

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
kand: It looks like github / react is unhappy that `state.loadingStatus` is missing from the dependency array. I think we'll need it here because `state` is technically state that should cause this `useEffect` to update. What about putting a check at the top of the effect with a `ref` that checks if it's been run already?

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+        state.sdkInstance,
+        state.createInstanceOptions,
+        state.scriptOptions,
+    ]);
+
+    // Separate effect for eligibility - runs after instance is created
+    useEffect(() => {
+        if (isServer()) {
+            return;
+        }
```
kand: Given the note on line 101:

> SDK loading effect - only runs on client (useEffect doesn't run during SSR)

Do you need this `isServer` check here?

kand (commented):
A couple small things, but looks great otherwise!

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+ *   createInstanceOptions={{
+ *     components: ["paypal-payments"],
+ *     clientToken: token,
+ *   }}
+ *   scriptOptions={{ environment: "sandbox" }}
+ * >
+ *   {children}
+ * </PayPalSdkInstanceProvider>
+ */
+export const PayPalSdkInstanceProvider: React.FC<
```
hamzanasir: I would actually vote name this more generic. Something like `<PayPalProvider /> `or `<PayPalCheckoutProvider />`. I think that the word `SdkInstance` doesn't hold a lot of meaning for the merchant since the only consumers of this provider will be our hooks/components. Stripe has something similar I believe:

```
<CheckoutProvider stripe={stripePromise} options={{clientSecret: promise}}>
      <CheckoutForm />
</CheckoutProvider>
```

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
hamzanasir: +1

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
EvanReinstein: Good catch!

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+ * - Server: Initializes in INITIAL state, no script loading attempted
+ * - Client: Initializes in PENDING state, loads scripts on mount
+ * - Hydration: Client takes over from server state seamlessly
+ *
+ * Props are automatically deep-compared to prevent unnecessary SDK reloads.
+ * You can pass options directly without manual memoization.
+ *
+ * @example
+ * <PayPalSdkInstanceProvider
+ *   createInstanceOptions={{
```
hamzanasir: Once again I don't think `createInstanceOptions` will ring a bell to anyone who hasn't used the v6 SDK before. I would vote for a more abstracted flattened provider like this one:

```
<PayPalCheckoutProvider
    components: ["paypal-payments"],
    clientToken: token,
    environment: "sandbox"
>
    {children}
</PayPalSdkInstanceProvider>
```

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+        const loadSdk = async () => {
+            try {
+                // Load the core SDK script
+                const paypalNamespace = await loadCoreSdkScript(
+                    state.scriptOptions,
+                );
+
+                if (
+                    controller.signal.aborted ||
```
hamzanasir: How does this abort logic work? If the `useEffect` cleanup function runs which aborts this controller I don't think this hook will run again? Also I believe `AbortController` is used for aborting async network requests which is a bit different from how it's being used here

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
EvanReinstein: Yes, going to have to add a `useRef` to track this. At first I thought that since `loadingStatus !== Initial` after the first mount that a `useRef` wasn't necessary since the condition will never be fulfilled. 

Massive but here...but it is safer and more explicit to leverage `useRef`. Let me push a change

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
EvanReinstein: Point taken!

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
EvanReinstein: Ah you know what, I am implementing wrong. I learned about `AbortController` via reviewing a [package](https://github.com/weareseeed/react-square-web-payments-sdk) that is focused on providing Square payment processing. I thought it would be neat to include this, however I'm reviewing my notes and I think there is more ground work to leveraging the abort signal.

I am going to remove this in favor of just using `isSubscribed`

packages/react-paypal-js/src/v6/components/PayPalSdkInstanceProvider.tsx
EvanReinstein: Ok I considered this and tbh it would be helpful in terms of inline objects passed as props causing rerenders, which the provider then needs to manage.

If other folks don't have strong opinions against this I'll go down this path.

EvanReinstein: @hamzanasir Alrighty, I updated the naming convention across the board. Lmk what you think

packages/paypal-js/src/utils.ts
```diff
@@ -158,3 +158,7 @@ function createScriptElement(
 
     return newScript;
 }
+
+export function isServer(): boolean {
+    return typeof window === "undefined" && typeof document === "undefined";
```
hamzanasir, line 163: Do we need the `document` check?

packages/paypal-js/src/v6/index.ts
```diff
@@ -24,7 +31,7 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
         insertScriptElement({
             url: url.toString(),
             onSuccess: () => {
-                if (!window.paypal) {
+                if (isServerEnv || !window.paypal) {
```
hamzanasir: Would we even get to this line of code if `isServerEnv` is `true`? I would assume the above SSR safeguard does an early return

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -0,0 +1,247 @@
...
+ * SSR Behavior:
+ * - Server: Initializes in INITIAL state, no script loading attempted
+ * - Client: Initializes in PENDING state, loads scripts on mount
+ * - Hydration: Client takes over from server state seamlessly
+ *
+ * Props are automatically deep-compared to prevent unnecessary SDK reloads.
+ * You can pass options directly without manual memoization.
+ *
+ * @example
+ * <PayPalSdkInstanceProvider
```
hamzanasir:
```suggestion
 * <PayPalProvider
```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -0,0 +1,247 @@
...
+                    createInstanceOptions: memoizedCreateOptions,
+                    scriptOptions: memoizedScriptOptions,
+                },
+            });
+        }
+    }, [
+        memoizedCreateOptions,
+        memoizedScriptOptions,
+        state.createInstanceOptions,
+        state.scriptOptions,
```
hamzanasir: I'm a little confused by this logic. Why are we listening to both options for changes?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -0,0 +1,247 @@
...
+    // SDK loading effect - only runs on client (useEffect doesn't run during SSR)
+    useEffect(() => {
+        if (
+            state.loadingStatus !== INSTANCE_LOADING_STATE.PENDING ||
+            state.sdkInstance
+        ) {
+            return;
+        }
+
+        let isSubscribed = true;
```
hamzanasir, line 116: Can you explain what this does exactly? To my knowledge once the hook calls the cleanup function this useEffect will be lost so it shouldn't run anymore. Do we need this `isSubscribed` logic?

hamzanasir (approved):
Great work @EvanReinstein!

packages/paypal-js/src/v6/index.ts
EvanReinstein: Oh shoot, correct. Let me push that change

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Great catch!

packages/paypal-js/src/utils.ts
EvanReinstein: We discussed this offline on 10/3 and agreed to keep for now 👍

packages/paypal-js/src/v6/index.ts
```diff
@@ -8,6 +8,13 @@ const version = "__VERSION__";
 
 function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
     validateArguments(options);
+    const isServerEnv = isServer();
+
+    // SSR safeguard
+    if (isServerEnv) {
+        return Promise.resolve(null);
```
imbrian, line 18: Do we want a `noop` here or a warning? Presumably this would only be called in a `useEffect`, which won't be run until the client. Seems like this being called on the server is probably more indicative of an integration problem.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
imbrian: It's an attempt to prevent `loadSdk` from continuing to run with an update happening in the background. It's async, and its steps are also async. It probably would benefit from a check before `await loadCoreSdkScript` though, since we're already in deferred operation at that point.

I think if you wanted to clean it up, you could use an `AbortController` instead. Haven't looked at what APIs are used in `loadCoreSdkScript` to see if any of them actually take an `AbortSignal`, but you could write your own, if this is going to be a common pattern.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+        const loadEligibility = async () => {
+            try {
+                const eligiblePaymentMethods =
+                    await state.sdkInstance?.findEligibleMethods({});
```
imbrian, line 178: Does there need to be a mechanism for the merchant to override the necessity of this call? Can the merchant provide their own server-side eligibility determination?

EvanReinstein: Yes, that work is forthcoming 👍 and this is more of a placeholder.

packages/paypal-js/src/v6/index.ts
EvanReinstein: Great point on adding a warning!

PR #689, opened 2025-09-29
feat(types): improve v6 types and add tests

gregjopa:
This PR updates the v6 TypeScript types

### DIFF
--- a/packages/paypal-js/types/v6/components/base-component.d.ts
+++ b/packages/paypal-js/types/v6/components/base-component.d.ts
@@ -1,0 +1,49 @@
+export type OnCompleteData = {
+    paymentSessionState: "approved" | "cancelled" | "error";
+};
+
+export type OnErrorData = Error;
+
+export type BasePaymentSessionOptions = {
+    // this onApproveData will be overwritten
+    onApprove: (data: Record<string, unknown>) => Promise<void>;
+    onCancel?: () => void;
+    onComplete?: (data: OnCompleteData) => void;
+    onError?: (data: OnErrorData) => void;
+    testBuyerCountry?: string;
+};
+
+export type PresentationModeOptionsForPopup = {
+    presentationMode: "popup";
+    fullPageOverlay?: { enabled: boolean };
+};
+
+export type PresentationModeOptionsForModal = {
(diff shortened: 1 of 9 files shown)


EvanReinstein (approved):
This lgtm! Thanks for organizing and giving these a once over 👍 💯

packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -0,0 +1,59 @@
+export type OnApproveDataOneTimePayments = {
+    orderId: string;
+    payerId?: string;
+    billingToken?: string;
+};
+
+export type OnCompleteData = {
+    paymentSessionState: "approved" | "cancelled" | "error";
```
bywood: I think this should be:
```suggestion
    paymentSessionState: "approved" | "canceled" | "error";
```

packages/paypal-js/types/v6/tests/paypal-legacy-billing-agreements.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        clientToken: "fakeValue",
+        components: ["paypal-legacy-billing-agreements"],
+    });
+
+    const paymentMethods = await sdkInstance.findEligibleMethods({
+        currencyCode: "USD",
+    });
+
+    if (!paymentMethods.isEligible("paypal")) {
+        return;
```
nbierdeman, line 31: Should this `throw` here and in the other tests?

packages/paypal-js/types/v6/components/base-component.d.ts
gregjopa: Thanks @bywood. I updated it to `canceled`

packages/paypal-js/types/v6/tests/paypal-legacy-billing-agreements.test.ts
gregjopa: Good question @nbierdeman. When a payment method is ineligible, its not necessarily an error.  For example, for non-US traffic, venmo will not be eligible. I think we should not error, and instead just return early and not render the button.

PR #690, opened 2025-10-02
feat: add jsdoc comments to v6 public interface types

nbierdeman:
This PR adds JSDoc comments to v6 public interface types.

### DIFF
--- a/packages/paypal-js/types/v6/index.d.ts
+++ b/packages/paypal-js/types/v6/index.d.ts
@@ -8,14 +8,13 @@ import {
 
 export interface PayPalV6Namespace {
     /**
-     * Creates an SDK instance with the specified components and configuration.
+     * Creates an SDK instance, which is the first step in an SDK integration. This instance serves as the base layer for all SDK components.
      *
      * This is an asynchronous method that initializes the PayPal SDK with the provided
-     * client token and components. The SDK is designed to be lightweight and modular,
-     * allowing you to load only the functionality you need.
+     * client token and components.
      *
      * @param {CreateInstanceOptions} createInstanceOptions - Configuration options for creating the SDK instance
-     * @returns {Promise<SdkInstance<T>>} - A promise that resolves to an SDK instance with methods based on the specified components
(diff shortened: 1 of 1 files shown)


packages/paypal-js/types/v6/index.d.ts
```diff
@@ -7,6 +7,26 @@ import {
 } from "./components/find-eligible-methods";
 
 export interface PayPalV6Namespace {
+    /**
+     * Creates an SDK instance with the specified components and configuration.
```
gregjopa:
```suggestion
     * Creates a SDK instance which is the first step in a SDK integration. This instance serves as the base layer for all SDK components.
```

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -7,6 +7,26 @@ import {
...
+     *
+     * This is an asynchronous method that initializes the PayPal SDK with the provided
+     * client token and components. The SDK is designed to be lightweight and modular,
```
gregjopa: I think we should remove this part about the SDK design. It feels like marketing material.
```suggestion
     * client token and components.
```

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -7,6 +7,26 @@ import {
...
+     * allowing you to load only the functionality you need.
+     *
+     * @param {CreateInstanceOptions} createInstanceOptions - Configuration options for creating the SDK instance
+     * @returns {Promise<SdkInstance<T>>} - A promise that resolves to an SDK instance with methods based on the specified components
```
gregjopa: I don't think there's any context about what `T` is here for the generic to make sense. I wonder if we can replace with
```suggestion
     * @returns {Promise<SdkInstance<Components[]>>} - A promise that resolves to an SDK instance with methods based on the specified components
```

gregjopa (approved):
Awesome work @nbierdeman! 💯 

I left a minor comment about the createInstance description. After that's addressed, I'm happy to see this get merged.

PR #691, opened 2025-10-03
fix(types): export v6 types for callback arguments

gregjopa:


EvanReinstein (approved):
🚀

bywood (approved):

packages/paypal-js/types/v6/tests/paypal-payments.test.ts
```diff
@@ -31,8 +31,19 @@ async function main() {
         return;
     }
 
+    function onApproveCallback({
```
gregjopa, line 34: I updated the onApprove() callback in these examples. Note this this code never actually executes so having the `console.log()` in here is fine. It's only meant to be evaluated by TypeScript as example integration code to make sure our types are working as expected.

packages/react-paypal-js/src/components/braintree/BraintreePayPalButtons.test.tsx
```diff
@@ -372,20 +372,18 @@ describe("render Braintree PayPal button component", () => {
             dataClientToken: CLIENT_TOKEN,
         };
         render(
-            <PayPalScriptProvider options={ options }>
-                <BraintreePayPalButtons
-                    message={{ amount: "100" }}
-                />
-            </PayPalScriptProvider>
+            <PayPalScriptProvider options={options}>
```
gregjopa, line 375: This file got formatted by prettier. No actual changes were made to it. I think its safe to just commit it.

PR #696, opened 2025-10-08
Feat: Provider Script Cleanup

EvanReinstein:
This change enables the React provider to clean up loose script elements 2 ways:
- `loadCoreSdkScript` checks if there is an existing Core script element and determines whether to re-use an already established namespace or remove the script element and update the namespace (when `scriptOptions` change).
- The provider will use the method `cleanupComponentScripts` on mount and unmount to remove Component-level scripts if needed.

There are a few outstanding questions that I have, please see the comments.

### DIFF
--- a/.changeset/brave-ties-bathe.md
+++ b/.changeset/brave-ties-bathe.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+Enhance V6 React Provider to clean up script elements


packages/paypal-js/src/v6/index.ts
```diff
@@ -20,6 +20,32 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+    const currentScript = document.querySelector<HTMLScriptElement>(
+        'script[src*="/web-sdk/v6/core"]',
+    );
+
+    if (currentScript) {
+        const currentUrl = new URL(currentScript.src);
+        const currentEnvironment = currentUrl.hostname.includes("sandbox")
+            ? "sandbox"
+            : "production";
+        const currentDebug = currentUrl.searchParams.get("debug") === "true";
```
EvanReinstein, line 32: A couple of methods to employ here in order to evaluate what the environment is and whether the debug query param exists.

I chose to use `new URL` in order to get access to the debug param more easily. I am open to changing this, and was considering just using RegEx here for this, please let me know what you think.

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -29,3 +29,16 @@ export function useDeepCompareMemoize<T>(value: T): T {
 
     return ref.current;
 }
+
+export function cleanupComponentScripts(): void {
+    const paypalScripts = document.querySelectorAll<HTMLScriptElement>(
+        'script[src*="paypal.com/web-sdk/v6"]',
```
EvanReinstein, line 35: The component script src attribute always include the full semver, i.e. v6.33.2, while the core script is always v6. With that in mind I think this string could be updated to: 'script[src*="paypal.com/web-sdk/v6."]'

This would only select Component-level scripts and would remove the need to check whether or not the script is the Core script. But I was concerned that the script src may change in the future. Let me know if you have an opinion on this

packages/paypal-js/src/v6/index.ts
```diff
@@ -20,6 +20,32 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+
+        const isEnvironmentMatch =
+            environment === currentEnvironment ||
+            (!environment && currentEnvironment === "sandbox");
+        const isDebugMatch = (debug ?? false) === currentDebug;
+
+        if (isEnvironmentMatch && isDebugMatch && window.paypal) {
```
kand, line 39: Do we need to check if `window.paypal.version` matches as well?

packages/react-paypal-js/src/v6/utils.ts
kand: > the core script is always v6

Is that true? If I give a version to one-cent (e.g. https://one-cent-sb.fly.dev/mixed-version?coreVersion=v6.29.2&packageVersion=v6.29.2) then a request to `https://www.sandbox.paypal.com/web-sdk/v6.29.2/core?debug=true` is made, which includes the minor and patch numbers for the `core` script.

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -29,3 +29,16 @@ export function useDeepCompareMemoize<T>(value: T): T {
...
+    );
+
+    paypalScripts.forEach((script) => {
+        // Only remove component scripts
+        if (!script.src.includes("/v6/core")) {
+            script.remove();
```
kand, line 41: Why do we want to remove these separately from the `core` script?

EvanReinstein: The `loadCoreSdkScript` has all the `core` script logic, so opted to manage clean up there rather than on the component.

packages/paypal-js/src/v6/index.ts
EvanReinstein: Are you thinking in case there is a V5 script on the page as well?

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein: Absolutely correct, but our use case is less dynamic since `loadCoreSdkScript` sets [the path](https://github.com/paypal/paypal-js/blob/61aba9fe90afcbd8ca2eef5eaffad50e284510b0/packages/paypal-js/src/v6/index.ts#L13-L17) as `v6`.

packages/paypal-js/src/v6/index.ts
kand: Yea, though I suppose, `window.paypal` will be in a bad state we want to clean up in that case anyways. So not checking version is probably safer, right?

packages/react-paypal-js/src/v6/utils.ts
kand: Hmm, would it be better to group both the core and component cleanup together, i.e. put this logic in `loadCoreSdkScript`? I imagine some future maintainer not realizing core and component cleanup are necessarily connected to each other and accidentally removing one or the other and extra `scripts` hanging out on the page.

EvanReinstein: I see your point and want to ensure that this is as maintainable as possible. Let me see how moving the component clean up to the load script would work. One aspect of this current solution is that the Core script is left in the DOM, and `loadCoreSdkScript` is really just checking to see if we need to re-use the current or insert a new `script` element and remove the old one.

EvanReinstein: Another issue thats make this complex is that the `core` and `component` scripts are dependent on different values. The `core` script should only be replaced if `scriptOptions` updates, while the `component` scripts should be updated when `createInstanceOptions` change AND when `scriptOptions` change.

I wonder if leaving a comment in the code would be enough for future developers. Lmk what you think. Otherwise we can pursue a more aggressive refactor.

PR #697, opened 2025-10-08
usePayLaterOneTimePaymentSession

kand:
This PR implements the pay later one time payment session hook. The proposed usage for this hook would be as follows:

```ts
const { handleClick, handleCancel, handleDestroy } = usePayLaterOneTimePaymentSession({
    onApprove: async (data) => {
      // onApprove logic
    },
    onCancel: () => {
       // onCancel logic
    },
    onError: (error: Error) => {
       // onError logic
    },
    presentationMode,
  });
```

```html
<paypal-pay-later-button
    onClick={handleClick}
    id="paylater-button"
></paypal-pay-later-button>
```

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.tsx
+++ b/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.tsx
@@ -23,7 +23,7 @@ describe("usePayLaterOneTimePaymentSession", () => {
         );
     });
 
-    it("should create a pay later payment session when the hook is called", () => {
+    test("should create a pay later payment session when the hook is called", () => {
         const mockOrderId = "123";
         const mockOnApprove = jest.fn();
         const mockOnCancel = jest.fn();
@@ -64,7 +64,7 @@ describe("usePayLaterOneTimePaymentSession", () => {
         });
     });
 
-    it("should error if there is no sdkInstance when called", () => {
+    test("should error if there is no sdkInstance when called", () => {
         const mockOrderId = "123";
 
         (usePayPal as jest.Mock).mockReturnValue({ sdkInstance: null });
@@ -81,7 +81,7 @@ describe("usePayLaterOneTimePaymentSession", () => {
         expect(error).toEqual(new Error("no sdk instance available"));
     });
 
-    it("should provide a click handler that calls session start", () => {
+    test("should provide a click handler that calls session start", async () => {
         const mockStart = jest.fn();
         const mockSession: OneTimePaymentSession = {
             start: mockStart,
@@ -109,7 +109,7 @@ describe("usePayLaterOneTimePaymentSession", () => {
             }),
         );
 
-        act(() => {
+        await act(async () => {
             handleClick();
         });
 
@@ -119,7 +119,7 @@ describe("usePayLaterOneTimePaymentSession", () => {
         });
     });
 
-    it("should call the createOrder callback, if it was provided, on start inside the click handler", () => {
+    test("should call the createOrder callback, if it was provided, on start inside the click handler", async () => {
         const mockStart = jest.fn();
         const mockSession: OneTimePaymentSession = {
             start: mockStart,
@@ -149,8 +149,12 @@ describe("usePayLaterOneTimePaymentSession", () => {
             }),
         );
 
-        act(() => {
+        await act(async () => {
             handleClick();
+        });
+
+        expect(mockCreatePayLaterOneTimePaymentSession).toHaveBeenCalledWith({
+            orderId: undefined,
         });
 
         expect(mockStart).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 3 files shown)


packages/react-paypal-js/src/v6/hooks/useCreatePayLaterOneTimePaymentSession.ts
```diff
@@ -0,0 +1,45 @@
...
+            updateSession(newSession);
+        } else {
+            // TODO what if sdk instance is not available? Error?
+            throw new Error("no sdk instance available");
+        }
+
+        return () => {
+            sessionRef.current?.destroy();
+            sessionRef.current = null;
+            updateSession(null);
```
EvanReinstein: It is unclear to me if we need to update the session to `null` here.

In the case that the component is unmounted and this clean up runs, then we're updating state within an unmounted component and will trigger a React warning. I read in [this blog post](https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/) that the warning was removed in React 18, but `react-paypal-js` is on React 17. Found the actual React Working Group post on it [here](https://github.com/reactwg/react-18/discussions/82).

In the case that one of the dependencies has updated and the useEffect is going to rerun, I believe that the previous session will be replaced with the new session before the button re-renders.

EvanReinstein: We discussed [on slack here](https://paypal.slack.com/archives/C090XES8HQ8/p1760019766962459) to remove this, but I wanted to keep the comment for posterity's sake

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (sdkInstance) {
+            const newSession =
+                sdkInstance.createPayLaterOneTimePaymentSession(
+                    memoizedOptions,
+                );
+            sessionRef.current = newSession;
+            updateSession(newSession);
+        } else {
+            // TODO what if sdk instance is not available? Error?
+            throw new Error("no sdk instance available");
```
EvanReinstein: Good question about this. I think there are a few reasons that `sdkInstance` would be null, like if the hook is being used outside of the Provider for example. I wonder if discussing the error handling strategy overall is a good idea.

kand: Let's discuss here: https://github.com/paypal/paypal-js/pull/702/files#r2433831373

packages/react-paypal-js/package.json
```diff
@@ -35,6 +35,7 @@
         "lint": "eslint .",
         "prepack": "rimraf dist && npm run build && npm run type-declarations",
         "test": "jest --env=jsdom",
+        "test:watch": "jest --env=jsdom --watch",
```
EvanReinstein, line 38: ❤️

hamzanasir: Ahh nice improvement!

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.tsx
```diff
@@ -0,0 +1,381 @@
...
+describe("usePayLaterOneTimePaymentSession", () => {
+    beforeEach(() => {
+        // mocking this for each test rather than a module so it can be easily unmocked
+        // where needed
+        (useProxyProps as jest.Mock).mockImplementation(
+            (callbacks) => callbacks,
+        );
+    });
+
+    it("should create a pay later payment session when the hook is called", () => {
```
hamzanasir: I expanded this test for venmo to also check if the callbacks when called trigger the one's passed in via props: https://github.com/paypal/paypal-js/pull/702/files#diff-d8c8d43bcb5e32ddfe9cbe9b1ecfb4afadaeed0feddca7fcf6dc12423835d92dR104-R111

Is this something that would be useful here too?

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.tsx
```diff
@@ -0,0 +1,381 @@
...
+    useProxyProps: jest.fn(),
+}));
+
+describe("usePayLaterOneTimePaymentSession", () => {
+    beforeEach(() => {
+        // mocking this for each test rather than a module so it can be easily unmocked
+        // where needed
+        (useProxyProps as jest.Mock).mockImplementation(
+            (callbacks) => callbacks,
+        );
```
hamzanasir, line 23: Ahh this is interesting, so this code just directly returns a reference to the function instead of going through `useProxyProps`. That's why you're able to check if the `createSession` call is called with the exact function that was passed into the hook. What if we didn't do this mock though since the hook is supposed to not return the original reference to the function and instead return a wrapped version of these callbacks?

kand: We chatted about this a bit yesterday and decided it's all good as is.

PR #698, opened 2025-10-09
feat: typedoc

nbierdeman:
This PR adds a scalable approach for generating public documentation for multiple packages within `paypal-js` using `typedoc`.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -4,54 +4,9 @@
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>PayPal JS SDK Documentation</title>
+        <link rel="stylesheet" href="paypal-fonts.css" />
         <style>
-            @font-face {
-                font-family: "PayPal Plain";
-                font-weight: 400;
-                font-display: swap;
-                font-style: normal;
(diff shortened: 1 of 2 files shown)


nbierdeman:
> Not a fan of `tabWidth: 4`, but I guess it's been like that for years. 🙂

@bywood, we could easily change it in a separate PR! Do you like `tabWidth: 2`?

docs/index.html
```diff
@@ -0,0 +1,154 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>PayPal JS SDK Documentation</title>
+        <style>
```
gregjopa, line 8: Should this be using the paypal-fonts.css file instead of inlining these styles?

nbierdeman: Yes, good catch. Done! https://github.com/paypal/paypal-js/pull/698/commits/f4dc24d0d3d9d659eea8ac0946e35dd5265ef16a

bywood:
@nbierdeman I'll survive with `tabWidth: 4`. When all our other repos use `tabWidth: 2`, it just seems odd when you come across it. 🙂

PR #695, opened 2025-10-07
Feat: React PayPal JS V6

EvanReinstein:

PR #699, opened 2025-10-10
Feat: usePayPalOneTimePaymentSession hook

EvanReinstein:
This PR implements the PayPal one time payment session hook, the proposed usage for this hook would be as follows:

```ts
const { handleClick, handleCancel, handleDestroy } = usePayPalOneTimePaymentSession({
    onApprove: async (data) => {
      // onApprove logic
    },
    onCancel: () => {
       // onCancel logic
    },
    onError: (error: Error) => {
       // onError logic
    },
    presentationMode,
  });
```

```jsx
<paypal-button
    onClick={handleClick}
    type="pay"
    id="paypal-button"
></paypal-button>
```

**Note:** there is a follow up task to update this hook to handle `redirect` and `direct-app-switch` presentation modes.


packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,575 @@
...
+            const props: UsePayPalOneTimePaymentSessionProps = {
+                presentationMode: "popup",
+                orderId: "test-order-id",
+                onApprove: jest.fn(),
+                onCancel: jest.fn(),
+                onError: jest.fn(),
+            };
+
+            try {
+                renderHook(() => usePayPalOneTimePaymentSession(props));
```
kand: There's a slightly cleaner way to do this by using the `error` returned from `renderHook`:

https://github.com/paypal/paypal-js/pull/695/files#diff-644d32d52c22e6a85441a3b2784c835fad6e22b94b6a698280b8d27bfe41d86aR77-R86

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,575 @@
...
+                mockSdkInstance.createPayPalOneTimePaymentSession.mock
+                    .calls[0][0];
+
+            expect(createSessionCall).toHaveProperty("onApprove");
+            expect(createSessionCall).toHaveProperty("onCancel");
+            expect(createSessionCall).toHaveProperty("onError");
+
+            expect(createSessionCall.onApprove).not.toBe(onApprove);
+            expect(createSessionCall.onCancel).not.toBe(onCancel);
+            expect(createSessionCall.onError).not.toBe(onError);
```
kand: @hamzanasir and I discussed this setup that it's not a sufficient unit test. I see the venmo tests are still setup this way, so maybe we could discuss again tomorrow?

EvanReinstein: Absolutely. I can't recall the outcome of that conversation so am definitely up to huddle on this.

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.test.ts
EvanReinstein: Lmk what you think of the update 👋

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.test.ts
kand: Ah it was on Monday when you were out. We can leave it for now and refactor it in the main feature branch. But in summary, this test would allow the following implementation of `usePayPalOneTimePaymentSession`:

```ts
const usePayPalOneTimePaymentSession = () => {
  const { sdkInstance } = usePayPal();

  sdkInstance.createPayPalOneTimePaymentSession({
    onApprove: "",
    onCancel: true,
    onError: 5,
  });  
})
```

Technically, the other tests would fail, but I think this test can be more robust.

You have the `useProxyProps` actual implementation available here, so you could refactor this into this test to make sure updating callbacks doesn't cause a re-render:

https://github.com/paypal/paypal-js/pull/695/files#diff-644d32d52c22e6a85441a3b2784c835fad6e22b94b6a698280b8d27bfe41d86aR277

All the other tests are using `useProxyProps` mock, so they're pretty much checking that the proxied function ends up in the create session call.

Open to huddling about this though if you'd like!

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,505 @@
...
+
+            jest.clearAllMocks();
+
+            rerender({ onApprove: newOnApprove });
+
+            expect(mockPayPalSession.destroy).not.toHaveBeenCalled();
+            expect(
+                mockSdkInstance.createPayPalOneTimePaymentSession,
+            ).not.toHaveBeenCalled();
+        });
```
EvanReinstein, line 275: @kand I think that this test gets at the components not re-rendering when a callback changes.

PR #701, opened 2025-10-14
Feat: Add check for existing core script

EvanReinstein:
This updates the `loadCoreSdkScript` method to return `window.paypal` if it detects that there is already a script loaded. This change is to prevent multiple Core script loads from occurring when the `PayPalProvider` re-renders and provider options have changed.

A follow up to this PR https://github.com/paypal/paypal-js/pull/696

### DIFF
--- a/.changeset/heavy-badgers-care.md
+++ b/.changeset/heavy-badgers-care.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+Adds an early resolve to loadCoreSdkScript if a v6 core script already exists.


packages/paypal-js/src/v6/index.ts
```diff
@@ -18,6 +18,14 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
         return Promise.resolve(null);
     }
 
+    const currentScript = document.querySelector<HTMLScriptElement>(
+        'script[src*="/web-sdk/v6/core"]',
+    );
+
+    if (window.paypal && currentScript) {
```
EvanReinstein: There is a question about the best condition to use here:
```
// would it be better to check:
// 1. window.__paypal_sdk__.v6
// 2. if window.paypal?.version is a v6 version or v5 version, and resolve the v6 namespace accordingly
// 3. window.paypal with the assumption that V5 provider is using data-namespace attribute and v6 has window.paypal all to themselves.
```

gregjopa: Can we also validate the version number?
```suggestion
    if (window.paypal?.version.startsWith("6") && currentScript) {
```

EvanReinstein: Would this end up being something similar to this
```
if (window.paypal?.version.startsWith('6') && currentScript) {
        return Promise.resolve(window.paypal as unknown as PayPalV6Namespace);
    } else if (window.paypal?.version.startsWith('5') && currentScript) {
        return Promise.resolve(window.paypal.v6 as unknown as PayPalV6Namespace);
}
```

And include an update to the `PayPalNamespace` to add the `v6` key:
```
export interface PayPalNamespace {
    Buttons?: (
        options?: PayPalButtonsComponentOptions,
    ) => PayPalButtonsComponent;
    ...
    version: string;
    v6?: PayPalV6Namespace
}
```

gregjopa: No, I recommend not having anything related to v5 in this. This `loadCoreSdkScript` is for v6 only.

PR #702, opened 2025-10-15
feat: Use Venmo One Time Payment Session Hook 

hamzanasir:
This PR implements the venmo one time payment session hook, the proposed usage for this hook would be as follows:

```ts
const { handleClick, handleCancel, handleDestroy } = useVenmoOneTimePaymentSession({
    onApprove: async (data) => {
      // onApprove logic
    },
    onCancel: () => {
       // onCancel logic
    },
    onError: (error: Error) => {
       // onError logic
    },
    presentationMode,
  });
```

```jsx
<venmo-button
    onClick={handleClick}
    type="pay"
    id="venmo-button"
></venmo-button>
```

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
@@ -2,25 +2,30 @@ import { renderHook, act } from "@testing-library/react-hooks";
 
 import { useVenmoOneTimePaymentSession } from "./useVenmoOneTimePaymentSession";
 import { usePayPal } from "./usePayPal";
+import { useProxyProps } from "../utils";
 import {
     INSTANCE_LOADING_STATE,
     type UseVenmoOneTimePaymentSessionProps,
     type VenmoOneTimePaymentSession,
 } from "../types";
 
-// Mock the usePayPal hook
 jest.mock("./usePayPal");
 
-const mockUsePayPal = usePayPal as jest.MockedFunction<typeof usePayPal>;
+jest.mock("../utils", () => ({
+    useProxyProps: jest.fn(),
+}));
 
-// Mock VenmoOneTimePaymentSession
+const mockUsePayPal = usePayPal as jest.MockedFunction<typeof usePayPal>;
+const mockUseProxyProps = useProxyProps as jest.MockedFunction<
+    typeof useProxyProps
+>;
+
 const createMockVenmoSession = (): VenmoOneTimePaymentSession => ({
     start: jest.fn().mockResolvedValue(undefined),
     cancel: jest.fn(),
     destroy: jest.fn(),
 });
 
-// Mock SDK instance
 const createMockSdkInstance = (venmoSession = createMockVenmoSession()) => ({
     createVenmoOneTimePaymentSession: jest.fn().mockReturnValue(venmoSession),
 });
@@ -30,6 +35,8 @@ describe("useVenmoOneTimePaymentSession", () => {
     let mockSdkInstance: ReturnType<typeof createMockSdkInstance>;
 
     beforeEach(() => {
+        mockUseProxyProps.mockImplementation((callbacks) => callbacks);
+
         mockVenmoSession = createMockVenmoSession();
         mockSdkInstance = createMockSdkInstance(mockVenmoSession);
 
@@ -68,7 +75,6 @@ describe("useVenmoOneTimePaymentSession", () => {
             try {
                 renderHook(() => useVenmoOneTimePaymentSession(props));
             } catch (error) {
-                // The hook throws when SDK instance is null, which is expected behavior
                 expect(error).toEqual(new Error("no sdk instance available"));
             }
         });
@@ -96,9 +102,9 @@ describe("useVenmoOneTimePaymentSession", () => {
                 mockSdkInstance.createVenmoOneTimePaymentSession,
             ).toHaveBeenCalledWith({
                 orderId: "test-order-id",
-                onApprove: expect.any(Function),
-                onCancel: expect.any(Function),
-                onError: expect.any(Function),
+                onApprove,
+                onCancel,
+                onError,
             });
 
             const mockData = { orderId: "test-order-id" };
@@ -137,12 +143,11 @@ describe("useVenmoOneTimePaymentSession", () => {
                 mockSdkInstance.createVenmoOneTimePaymentSession,
             ).toHaveBeenCalledWith({
                 orderId: undefined,
-                onApprove: expect.any(Function),
-                onCancel: expect.any(Function),
-                onError: expect.any(Function),
+                onApprove,
+                onCancel,
+                onError,
             });
 
-            // Test that the proxied callbacks actually call the original functions
             const mockData = { test: "data" };
             createSessionCall.onApprove(mockData);
             createSessionCall.onCancel();
@@ -174,19 +179,21 @@ describe("useVenmoOneTimePaymentSession", () => {
         });
 
         test("should recreate session when orderId changes", () => {
+            const onApprove = jest.fn();
+            const onCancel = jest.fn();
+            const onError = jest.fn();
             const { rerender } = renderHook(
                 ({ orderId }) =>
                     useVenmoOneTimePaymentSession({
                         presentationMode: "popup",
                         orderId,
-                        onApprove: jest.fn(),
-                        onCancel: jest.fn(),
-                        onError: jest.fn(),
+                        onApprove,
+                        onCancel,
+                        onError,
                     }),
                 { initialProps: { orderId: "test-order-id-1" } },
             );
 
-            // Clear mock calls from initial render
             jest.clearAllMocks();
 
             rerender({ orderId: "test-order-id-2" });
@@ -196,9 +203,9 @@ describe("useVenmoOneTimePaymentSession", () => {
                 mockSdkInstance.createVenmoOneTimePaymentSession,
             ).toHaveBeenCalledWith({
                 orderId: "test-order-id-2",
-                onApprove: expect.any(Function),
-                onCancel: expect.any(Function),
-                onError: expect.any(Function),
+                onApprove,
+                onCancel,
+                onError,
             });
         });
 
@@ -215,10 +222,8 @@ describe("useVenmoOneTimePaymentSession", () => {
                 useVenmoOneTimePaymentSession(props),
             );
 
-            // Clear mock calls from initial render
             jest.clearAllMocks();
 
-            // Change SDK instance
             const newMockSession = createMockVenmoSession();
             const newMockSdkInstance = createMockSdkInstance(newMockSession);
 
@@ -240,6 +245,10 @@ describe("useVenmoOneTimePaymentSession", () => {
         });
 
         test("should NOT recreate session when only callbacks change", () => {
+            mockUseProxyProps.mockImplementation(
+                jest.requireActual("../utils").useProxyProps,
+            );
+
             const initialOnApprove = jest.fn();
             const newOnApprove = jest.fn();
 
@@ -308,10 +317,9 @@ describe("useVenmoOneTimePaymentSession", () => {
                 await result.current.handleClick();
             });
 
-            // The hook calls createOrder() and passes the result as second argument to start()
             expect(mockVenmoSession.start).toHaveBeenCalledWith(
                 { presentationMode: "popup" },
-                expect.any(Promise), // createOrder() returns a Promise
+                expect.any(Promise),
             );
             expect(mockCreateOrder).toHaveBeenCalled();
         });
@@ -329,7 +337,6 @@ describe("useVenmoOneTimePaymentSession", () => {
                 useVenmoOneTimePaymentSession(props),
             );
 
-            // Unmount to destroy session
             unmount();
 
             await expect(
@@ -363,7 +370,6 @@ describe("useVenmoOneTimePaymentSession", () => {
                     presentationMode: mode,
                 });
 
-                // Clear for next iteration
                 jest.clearAllMocks();
                 mockVenmoSession = createMockVenmoSession();
                 mockSdkInstance = createMockSdkInstance(mockVenmoSession);
@@ -413,7 +419,6 @@ describe("useVenmoOneTimePaymentSession", () => {
                 useVenmoOneTimePaymentSession(props),
             );
 
-            // Unmount to destroy session
             unmount();
 
             expect(() => {
@@ -458,7 +463,6 @@ describe("useVenmoOneTimePaymentSession", () => {
                 useVenmoOneTimePaymentSession(props),
             );
 
-            // Unmount to destroy session
             unmount();
 
             expect(() => {
@@ -481,12 +485,10 @@ describe("useVenmoOneTimePaymentSession", () => {
                 useVenmoOneTimePaymentSession(props),
             );
 
-            // Destroy the session manually
             act(() => {
                 result.current.handleDestroy();
             });
 
-            // Trying to click after manual destroy should throw
             await expect(
                 act(async () => {
                     await result.current.handleClick();
@@ -496,6 +498,12 @@ describe("useVenmoOneTimePaymentSession", () => {
     });
 
     describe("callback proxying", () => {
+        beforeEach(() => {
+            mockUseProxyProps.mockImplementation(
+                jest.requireActual("../utils").useProxyProps,
+            );
+        });
+
         test("should proxy callbacks correctly through useProxyProps", () => {
             const onApprove = jest.fn();
             const onCancel = jest.fn();
@@ -553,7 +561,6 @@ describe("useVenmoOneTimePaymentSession", () => {
                 handleDestroy: result.current.handleDestroy,
             };
 
-            // Functions should be stable across re-renders
             expect(firstRender.handleClick).toBe(secondRender.handleClick);
             expect(firstRender.handleCancel).toBe(secondRender.handleCancel);
             expect(firstRender.handleDestroy).toBe(secondRender.handleDestroy);
--- a/packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
+++ b/packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
@@ -4,6 +4,7 @@ import type {
     VenmoOneTimePaymentSessionPromise,
 } from "@paypal/paypal-js/sdk-v6";
 
+// TODO: Add all startOptions to this hook
 export type UseVenmoOneTimePaymentSessionProps =
     | (Omit<VenmoOneTimePaymentSessionOptions, "orderId"> & {
           createOrder: () => VenmoOneTimePaymentSessionPromise;


packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -29,3 +29,33 @@ export function useDeepCompareMemoize<T>(value: T): T {
 
     return ref.current;
 }
+
+export function useProxyProps<T extends Record<PropertyKey, unknown>>(
```
imbrian, line 33: How do we want to deal with duplicate code like this?

The original implementation has tests too: https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/hooks/useProxyProps.test.ts

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
```diff
@@ -0,0 +1,78 @@
...
+    orderId,
+    ...callbacks
+}: UseVenmoOneTimePaymentSessionProps): UseVenmoOneTimePaymentSessionReturn {
+    const { sdkInstance } = usePayPal();
+    const sessionRef = useRef<VenmoOneTimePaymentSession | null>(null);
+    const proxyCallbacks = useProxyProps(callbacks);
+
+    useEffect(() => {
+        if (!sdkInstance) {
+            throw new Error("no sdk instance available");
```
imbrian, line 30: How do we want to teach error handling? I think a hook throwing versus returning an error state forces a specific integration pattern, because it dictates where a merchant needs to place error boundaries.

An error state allows merchant to conditionally handle gracefully in the consuming component versus in a higher error boundary. If that's what we want, then we need our examples to show wrapping the components that do this setup in an error boundary (and we should probably provide this error boundary component too).

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export function useVenmoOneTimePaymentSession({
+    presentationMode,
+    createOrder,
+    orderId,
+    ...callbacks
+}: UseVenmoOneTimePaymentSessionProps): UseVenmoOneTimePaymentSessionReturn {
+    const { sdkInstance } = usePayPal();
+    const sessionRef = useRef<VenmoOneTimePaymentSession | null>(null);
+    const proxyCallbacks = useProxyProps(callbacks);
```
imbrian, line 21: Yeah, I see what you're doing. Leveraging the stable ref in `useProxyProps` to satisfy the useeffect dependency array below, while allowing the underlying functions to change. It's the same use case as v5.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        return () => {
+            if (sessionRef.current) {
+                sessionRef.current.destroy();
+                sessionRef.current = null;
+            }
+        };
+    }, [sdkInstance, orderId, proxyCallbacks]);
+
+    const handleClick = useCallback(async () => {
```
kand, line 42: Do we need to implement the click handler for the merchant or is that something they can do? Same for the other handlers. Is the concern just that they may not use `start`, `cancel`, and `destroy` correctly?

If we do go with this pattern, how will we add additional `start` arguments to the top level of arguments for the session hook? I.e. other session types require more `start` options, how should those be passed in? I assume we'd want a similar interface for all of these session hooks if possible.

packages/react-paypal-js/src/v6/utils.ts
kand: Is there a limitation to importing code from the old react implementation?

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
kand: Discussed this point with the team and we're in agreement that this is the way to go.

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
kand, line 1: Following the type file convention, should this be `useVenmoOneTimePaymentSession.d.ts`? Is there a need to have these types separate from the `useVenmoOneTypePaymentSession.ts` file?

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
```diff
@@ -0,0 +1,78 @@
...
+            orderId,
+            ...proxyCallbacks,
+        });
+        sessionRef.current = newSession;
+
+        return () => {
+            if (sessionRef.current) {
+                sessionRef.current.destroy();
+                sessionRef.current = null;
+            }
```
kand: Would it work to shorten these lines to the following?

```suggestion
            sessionRef.current?.destroy();
            sessionRef.current = null;
```

kand: Could this entire `return` even just be `return handleDestroy`? Might be clearer that this is the exact same behavior.

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
```diff
@@ -0,0 +1,23 @@
+import type {
+    VenmoOneTimePaymentSessionOptions,
+    VenmoPresentationModeOptions,
+    VenmoOneTimePaymentSessionPromise,
+} from "@paypal/paypal-js/sdk-v6";
+
+export type UseVenmoOneTimePaymentSessionProps =
```
kand, line 8: I'll test this out in my PR as well, but I'm worried using `type` here instead of some kind of `interface` is going to leave merchants with some pretty ugly type errors.

packages/react-paypal-js/src/v6/utils.ts
hamzanasir: @kand had an idea to maybe create a shared utils folder between the v6 implementation and v5 implementation. We could pursue something like that 🤔

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
hamzanasir: Ahh that's a good callout @imbrian! I think the v5 implementation also relies on throwing an error right? If that's the case we may want to follow the same pattern here and teach using ErrorBoundary

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
hamzanasir: That's good to hear @imbrian! It's an awesome hook to use 🙂

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
hamzanasir: Hmm I believe `useVenmoOneTypePaymentSession.ts` doesn't exist. Are you suggestion we rename this file? The idea was to have the "hook" related types in a different file, but definitely open to moving it to the `venmo-payments.d.ts` file as well if that's what you mean 🙂

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
hamzanasir: Ahh good point. Yeah I'm not opposed to using `interface` if that works better 🤔

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
kand: Ah sorry, wrong words 😅 . I meant to say:

> Is there a need to have these types separate from the `useVenmoOneTimePaymentSession.ts` file?

As in, could we move the types to `packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts`, or is there a reason the types need to be in separate file?

Putting them in `venmo-payments.d.ts` also makes sense to me.

kand: Ooh, okay, I assume they're not in the file because we have a `types` package. If that's the case, this setup makes sense. Nit pick: could we follow the file naming convention and lowercase to `venmoOneTimePaymentSessionTypes`?

Edit: Sorry, disregard. I'm getting confused by the 2 different v6-related `type` directories. Nothing to change here.

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein: Currently the V6 specific tsconfig instructs rollup to not allow imports outside of the v6 directory. In a discussion yesterday we talked about including a barrel file to export this methods for v6 usage. We can also update the v6-specific tsconfig to allow importing this hook specifically or more hooks generally.

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
```diff
@@ -0,0 +1,23 @@
...
+    | (Omit<VenmoOneTimePaymentSessionOptions, "orderId"> & {
+          createOrder: () => VenmoOneTimePaymentSessionPromise;
```
kand, line 10: Not a part of this PR, but out of curiosity, do we need a different type for `VenmoOneTimePaymentSessionPromise`? It appears to be equivalent to `BasePaymentSessionPromise`.

Edit: [Slack thread here](https://paypal.slack.com/archives/C090XES8HQ8/p1760716693716589)

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein: Ok, discussed this with @gregjopa and he suggested that we go with a WET approach in order to make V5 "very deletable" in the future.

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
kand: Alright, well, experimenting with this today, there doesn't really seem to be an improvement using `interface` here. So what you have is all good!

kand: A bit unfortunate because the type error message is a bit confusing and verbose 🫤

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
kand: I like the idea of returning some kind of error state rather than throwing an error. Requiring merchants to use an `<ErrorBoundary>` in a specific place (i.e. outside the component using the hook) feels unnecessarily restrictive.

packages/react-paypal-js/src/v6/utils.ts
kand: Alright, sounds good. Did you mention yesterday we are sharing other stuff from v5 currently though? Do we need to duplicate that code?

kand: I moved this util into the base branch so we can all share it while we work on our hooks.

https://github.com/paypal/paypal-js/pull/695/files#r2440686418

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
hamzanasir: Yes that's a great suggestion @kand ! Thank you!

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,537 @@
...
+                onError: jest.fn(),
+            };
+
+            renderHook(() => useVenmoOneTimePaymentSession(props));
+
+            expect(
+                mockSdkInstance.createVenmoOneTimePaymentSession,
+            ).toHaveBeenCalledWith({
+                orderId: "test-order-id",
+                onApprove: expect.any(Function),
```
kand: Is `expect.any(Function)` leaving a mock function in place? All good, but I've usually seen `jest.fn()` in place of this or an empty arrow function `() => {}`.

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein: > Did you mention yesterday we are sharing other stuff from v5 currently though? Do we need to duplicate that code?

@kand In a prior iteration of the `loadCoreSdkScript` function, it was a nearly identical method to the `loadScript` function that V5 leverages. So it was basically copy-pasted over for V6 to use. I mentioned this as an example of the WET approach we're currently taking.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,537 @@
...
+                .mockImplementation();
+
+            try {
+                renderHook(() => useVenmoOneTimePaymentSession(props));
+            } catch (error) {
+                // The hook throws when SDK instance is null, which is expected behavior
+                expect((error as Error).message).toContain(
+                    "no sdk instance available",
+                );
+            }
```
kand, line 79: There's a slightly cleaner way to write this using the `result.error` returned by `renderHook`. See my test here:

https://github.com/paypal/paypal-js/pull/697/files#diff-7739eba6c04a62ff31d3d1edb1d736a8d486a9d9405688b749b095da209eee18R67-R82

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,537 @@
...
+                orderId: "test-order-id",
+                onApprove: jest.fn(),
+                onCancel: jest.fn(),
+                onError: jest.fn(),
+            };
+
+            // Mock console.error to suppress React's error logging
+            const consoleSpy = jest
+                .spyOn(console, "error")
+                .mockImplementation();
```
kand: A bit of a nit, but what errors are being logged? IMO we should expose these because they might be actual errors to fix.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
hamzanasir: I believe this is the check to see if `createVenmoOneTimePaymentSession` was called with the right things, if we have `jest.fn()` in place of this then it will return a new mock function which will fail the equality check I believe 🤔

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
hamzanasir: Ahh good call! Updated!

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
hamzanasir: Eagle eye as always @kand ! I think this is just AI hallucinated code 😅 Removed!

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,537 @@
...
+            );
+
+            // Unmount to destroy session
+            unmount();
+
+            await expect(
+                act(async () => {
+                    await result.current.handleClick();
+                }),
+            ).rejects.toThrow("Venmo session not available");
```
kand, line 346:
```suggestion
            expect(
                act(() => {
                   result.current.handleClick();
                }),
            ).rejects.toThrow("Venmo session not available");
```

I'm fairly certain `async` / `await` isn't needed here. My tests didn't end up needing it. Though, I am a bit confused by the sparse documentation on `act()`. In this example, `async` / `await` _is_ used, but not the inner `await`; i.e. in this case, remove the `await` in `await result.current.handleClick()`.

https://legacy.reactjs.org/docs/testing-recipes.html#data-fetching

hamzanasir: Hmm for some reason my tests seem to fail if I don't use async / await. Using async / await makes sense to me since handleClick is an async function 🤔

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
kand: Ah, sorry, I saw this usage in a few places and glossed over this being used in `expect`. There are several locations though, where I think the AI got confused and is using this where you'd typically see `jest.fn()`. I'll highlight one of those locations in another comment.

hamzanasir: Oh are you talking about passing props https://github.com/paypal/paypal-js/blob/8b7af4edcd4ba4c21272ccf7a8dbb735bed03e4d/packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts#L441? Yeah I think you're right that needs to change to use `jest.fn()`

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,528 @@
...
+    });
+
+    describe("session lifecycle", () => {
+        test("should destroy session on unmount", () => {
+            const props: UseVenmoOneTimePaymentSessionProps = {
+                presentationMode: "popup",
+                orderId: "test-order-id",
+                onApprove: expect.any(Function),
+                onCancel: expect.any(Function),
+                onError: expect.any(Function),
```
kand: Here's an example where we should be using `jest.fn()` rather than `expect.any(Function)`. There are several more of these.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,528 @@
...
+            await expect(
+                act(async () => {
+                    await result.current.handleClick();
+                }),
+            ).rejects.toThrow("Venmo session not available");
+        });
+    });
+
+    describe("callback proxying", () => {
+        test("should proxy callbacks correctly through useProxyProps", () => {
```
kand, line 507: In my tests, I mocked out `useProxyProps`:

https://github.com/paypal/paypal-js/pull/697/files#diff-7739eba6c04a62ff31d3d1edb1d736a8d486a9d9405688b749b095da209eee18R19-R23

Which I think is slightly more robust than the unit test here because this test isn't checking that the functions went through `useProxyProps`; for example, `createSessionCall.onApprove` could be `true` here and the test would pass. What do you think?

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
hamzanasir: Good call @kand !

packages/react-paypal-js/src/v6/types/VenmoOneTimePaymentSessionTypes.d.ts
EvanReinstein, line 8: As this type is defined, I'm not sure if the hook can accept `fullPageOverlay` as an argument. I was taking another look at this while looking into the `redirect` and `direct-app-switch` presentation modes for PayPal.

hamzanasir: This is a great callout @EvanReinstein! I think a follow up ticket would be great to get parity on all the different types of start options we can pass 🙂

PR #711, opened 2025-10-25
Refactor paypal provider

hamzanasir:


packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -95,70 +96,102 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+        if (!state.sdkInstance) {
             return;
         }
 
-        let isSubscribed = true;
+        dispatch({
+            type: INSTANCE_DISPATCH_ACTION.RESET_STATE,
+            value: INSTANCE_LOADING_STATE.SDK_LOADED,
+        });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
kand, line 120: I think we should turn what's missing here into `refs` rather than ignore the deps array.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -95,70 +96,102 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-                    return;
+                if (!window.paypal?.version.startsWith("6")) {
+                    throw new Error("Failed to load PayPal SDK");
+                }
+
+                dispatch({
+                    type: INSTANCE_DISPATCH_ACTION.SET_LOADING_STATUS,
+                    value: INSTANCE_LOADING_STATE.SDK_LOADED,
+                });
+            } catch (error) {
```
kand, line 141: This setup looks good 👍

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Would that mean keeping the `sdkInstance` in a ref AND in state? Would that open us up to issues if the instances aren't the same/managed?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -95,70 +96,102 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
         dispatch({
             type: INSTANCE_DISPATCH_ACTION.RESET_STATE,
+            value: INSTANCE_LOADING_STATE.PENDING,
         });
         // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, [memoizedCreateOptions, memoizedScriptOptions]); // Only reset when options change, not the SDK instance
+    }, [memoizedScriptOptions]); // Only reset when script options change
 
-    // SDK loading effect - only runs on client (useEffect doesn't run during SSR)
     useEffect(() => {
```
kand, line 111: I think this `useEffect` might conflict with the `useEffect` on line 99. Also, having this `dispatch` separate from the `createInstance` call feels incorrect, as in, I think we should have this `dispatch` occur in the same effect as the create call.

PR #713, opened 2025-10-28
chore: v6 docs updates

nbierdeman:
This PR updates the v6 `tsdoc` comments for documentation purposes. It also bumps the version of `@playwright/test` to overcome failures in CI.

### DIFF
--- a/docs/readmes/web-sdk-v6-reference.md
+++ b/docs/readmes/web-sdk-v6-reference.md
@@ -37,10 +37,10 @@ const sdkInstance = await window.paypal.createInstance({
 });
 
 function onApproveCallback(data: OnApproveDataOneTimePayments) {
-    /* ... */
+    // ...
 }
 function onShippingAddressChangeCallback(data: OnShippingAddressChangeData) {
-    /* ... */
+    // ...
 }
 
 const paypalCheckout = sdkInstance.createPayPalOneTimePaymentSession({
--- a/packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
+++ b/packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
@@ -54,7 +54,7 @@ export type FindEligibleMethodsGetDetails<T extends FundingSource> =
  *     onApprove: (data) => console.log('Payment approved:', data),
  *   });
  *   // set up button
- *   ...
+ *   // ...
  * }
  *
  * if (paymentMethods.isEligible("paylater")) {
@@ -69,7 +69,7 @@ export type FindEligibleMethodsGetDetails<T extends FundingSource> =
  *   const paylaterButton = document.querySelector("#paylater-button");
  *   paylaterButton.productCode = productCode;
  *   paylaterButton.countryCode = countryCode;
- *   ...
+ *   // ...
  * }
  *
  * if (paymentMethods.isEligible("credit")) {
@@ -83,7 +83,7 @@ export type FindEligibleMethodsGetDetails<T extends FundingSource> =
  *   // set up button
  *   const paypalCreditButton = document.querySelector("#paypal-credit-button");
  *   paypalCreditButton.countryCode = countryCode;
- *   ...
+ *   // ...
  * }
  * ```
  */


docs/readmes/web-sdk-v6-reference.md
```diff
@@ -1,3 +1,53 @@
...
+## Importing and Using v6 Types
+
+**paypal-js** includes comprehensive TypeScript type definitions to help you build type-safe integrations. Here's how to import and use the v6 types in your TypeScript projects.
+
+### Installation
+
+Install **paypal-js** as a dependency:
+
+```bash
+npm install @paypal/paypal-js
```
bywood: Should we include `--save-dev`? Or leave that up to the merchant?

docs/readmes/web-sdk-v6-reference.md
```diff
@@ -1,3 +1,53 @@
...
+    PayPalV6Namespace,
+    OnApproveDataOneTimePayments,
+    OnShippingAddressChangeData,
+} from "@paypal/paypal-js/sdk-v6";
+
+declare global {
+    interface Window {
+        paypal: PayPalV6Namespace;
+    }
+}
```
bywood, line 32: Hmm. Unrelated to this PR, but I didn't realize the merchant had to declare this. Is this not something that we can declare so the merchant doesn't have to?

docs/readmes/web-sdk-v6-reference.md
```diff
@@ -1,3 +1,53 @@
...
+
+const sdkInstance = await window.paypal.createInstance({
+    clientToken: "INSERT_YOUR_CLIENT_TOKEN_HERE",
+    components: ["paypal-payments", "venmo-payments"],
+});
+
+function onApproveCallback(data: OnApproveDataOneTimePayments) {}
+function onShippingAddressChangeCallback(data: OnShippingAddressChangeData) {}
```
bywood: Maybe just to indicate that this isn't showing everything...
```suggestion
function onApproveCallback(data: OnApproveDataOneTimePayments) { /* ... */ }
function onShippingAddressChangeCallback(data: OnShippingAddressChangeData) { /* ... */ }
```

packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
```diff
@@ -36,8 +36,71 @@ export type FindEligibleMethodsGetDetails<T extends FundingSource> =
...
+ * const paymentMethods = await sdkInstance.findEligibleMethods({
+ *   currencyCode: "USD",
+ * });
+ *
+ * if (paymentMethods.isEligible("paypal")) {
+ *   const paypalPaymentSession = sdkInstance.createPayPalOneTimePaymentSession({
+ *     onApprove: (data) => console.log('Payment approved:', data),
+ *   });
+ *   // set up button
+ *   ...
```
bywood: How does it render these `...`s, since they aren't valid syntax? Should we comment these out?
```suggestion
 *   // ...
```

docs/readmes/web-sdk-v6-reference.md
EvanReinstein: @bywood I'm curious for your 2 cents on this. Currently `paypal-js` does declare this, just only for v5 see [here](https://github.com/paypal/paypal-js/blob/acc3ba2fd0a67ab7bf70c445907531b2101e2959/packages/paypal-js/types/index.d.ts#L58-L62). I don't necessarily think that including v6 in the declaration is as easy as updating the `window.paypal` type to `paypal?: PayPalNamespace | PayPalV6Namespace | null;` but maybe I'm wrong?

packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
nbierdeman: I think it looks alright without turning it into a comment:
Screenshot 2025-10-29 at 4 25 08 PM

nbierdeman: What do you think?

nbierdeman: Meh, I just commented them out. 😁

docs/readmes/web-sdk-v6-reference.md
gregjopa: > paypal?: PayPalNamespace | PayPalV6Namespace | null;

I'd prefer the namespace `window.paypal` to be strongly typed. I worry that the OR operator ends up allowing both `window.paypal.Buttons` and `window.paypal.createInstance` to exist at the same time.

nbierdeman: I think the union operator means that `paypal` could only be one of those types at a time, as opposed to an intersection operator, which would allow for both at the same time. I just opened a draft PR that I'll test: https://github.com/paypal/paypal-js/pull/718.

PR #714, opened 2025-10-29
Refactor PayPalProvider

EvanReinstein:
The provider had a few todos/code smells that were not taken up during the first pass. This updates the following:
- Flattens `scriptOptions` prop to its 2 keys.
- Removes deep comparison of all props and targets the `components` array specfically.
- Removes usage of `dequal` package from dependencies and replaces with a custom `components` specific `isEqual` variation.
- Splits out load core script and create instance functionality.
- Removes/Replaces reset state `useEffect`
- Updates tests to reflect new logic/provider intention.

Takes some of the ideas that were put together in [this PR](https://github.com/paypal/paypal-js/pull/711)

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
@@ -10,7 +10,7 @@ import {
     INSTANCE_LOADING_STATE,
     INSTANCE_DISPATCH_ACTION,
 } from "../types/PayPalProviderEnums";
-import { useDeepCompareMemoize } from "../utils";
+import { useCompareMemoize } from "../utils";
 
 import type {
     CreateInstanceOptions,
@@ -41,7 +41,7 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
     environment,
     debug,
 }) => {
-    const memoizedComponents = useDeepCompareMemoize(components);
+    const memoizedComponents = useCompareMemoize(components);
 
     const [paypalNamespace, setPaypalNamespace] =
         useState<PayPalV6Namespace | null>(null);
--- a/packages/react-paypal-js/src/v6/utils.ts
+++ b/packages/react-paypal-js/src/v6/utils.ts
@@ -7,41 +7,40 @@ export function isServer(): boolean {
 }
 
 /**
- * Performs a shallow equality check on two components arrays used by PayPalProvider.
+ * Performs a shallow equality check on two arrays.
  *
- * This function is specifically designed to compare the `components` prop arrays passed to
- * PayPalProvider.
+ * This function compares two arrays element-by-element using strict equality (===).
+ * It's primarily used to compare the `components` prop arrays passed to PayPalProvider
+ * to prevent unnecessary re-initialization of the PayPal SDK.
  *
- * This optimization prevents unnecessary re-initialization of the PayPal SDK when the components
- * array reference changes but the actual component names remain the same. This is particularly
- * important because re-initializing the SDK is an expensive operation that involves loading scripts
- * and setting up PayPal integrations.
+ * This optimization is important because re-initializing the SDK is an expensive operation
+ * that involves loading scripts and setting up PayPal integrations.
  *
- * @param arr1 - First components array to compare
- * @param arr2 - Second components array to compare
- * @returns `true` if both arrays are null/undefined, or if they contain the same components in the same order
+ * @param arr1 - First array to compare
+ * @param arr2 - Second array to compare
+ * @returns `true` if both arrays are null/undefined, or if they contain the same elements in the same order
  *
  * @example
- * // Returns true - both arrays have the same components in the same order
- * isEqualComponentsArray(
+ * // Returns true - both arrays have the same elements in the same order
+ * shallowEqualArray(
  *   ["paypal-payments", "venmo-payments"],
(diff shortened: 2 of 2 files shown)


packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -55,83 +38,85 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-            !hasHandledInitialHydration.current
-        ) {
-            hasHandledInitialHydration.current = true;
-            dispatch({
-                type: INSTANCE_DISPATCH_ACTION.SET_LOADING_STATUS,
-                value: INSTANCE_LOADING_STATE.PENDING,
-            });
-        }
-    }, [state.loadingStatus]); // Run when loadingStatus changes, but only act once
+        let isSubscribed = true;
```
EvanReinstein, line 54: Initially this `useEffect` did not leverage the `isSubscribed` strategy, when running the unit tests I was seeing warnings related to not preventing state updates on unmounted components.

Without the `isSubscribed` strategy, technically this `useEffect` could run `setPaypalNamespace` on an unmounted component. 

I know that there is some debate about whether we need the `isSubscribed` strategy, I fall on the side of we should prevent state updates or dispatches to unmounted components and avoid the subsequent console warnings (courtesy of React), until we can upgrade the React version to 18/19. Check out [this](https://github.com/reactwg/react-18/discussions/82) React Working Group post on removing this warning from React 18.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -55,83 +38,85 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-
-                if (!isSubscribed || !paypalNamespace) {
-                    return;
-                }
-
                 // Create SDK instance
-                const instance = await paypalNamespace.createInstance(
-                    memoizedCreateOptions,
-                );
+                const instance = await paypalNamespace.createInstance({
```
kand, line 108: Do we need a `!isSubscribed` check before this line? Is it possible for the component to have unmounted between calling the parent `useEffect` and the start of the async function, before `createInstance` has been called?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -55,83 +38,85 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-            dispatch({
-                type: INSTANCE_DISPATCH_ACTION.SET_LOADING_STATUS,
-                value: INSTANCE_LOADING_STATE.PENDING,
-            });
-        }
-    }, [state.loadingStatus]); // Run when loadingStatus changes, but only act once
+        let isSubscribed = true;
+
+        const loadSdk = async () => {
+            try {
```
kand, line 57: Similarly, here, do we need an `!isSubscribed` check at the top of this `async` function?

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -1,29 +1,53 @@
 import { useRef } from "react";
-import { dequal } from "dequal";
+
+import type { Components } from "./types";
 
 export function isServer(): boolean {
     return typeof window === "undefined" && typeof document === "undefined";
 }
 
+function isEqualComponentsArray(
```
kand: Would it be helpful to leave a jsdoc comment for this function?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -55,83 +38,85 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-    const memoizedScriptOptions = useDeepCompareMemoize(scriptOptions);
-
-    // Track if we've already handled the initial hydration
-    const hasHandledInitialHydration = useRef(false);
+    const memoizedComponents = useDeepCompareMemoize(components);
 
-    const [state, dispatch] = useReducer(instanceReducer, {
-        sdkInstance: null,
-        eligiblePaymentMethods: null,
-        loadingStatus: INSTANCE_LOADING_STATE.INITIAL,
```
kand, line 79: Does the `INITIAL` value need to be removed from [the enum](https://github.com/paypal/paypal-js/pull/695/files#diff-883a9ac9b45d9d9fb732a559923008cf1b16e042b5aebaf1d79a9d770055838cR2)?

EvanReinstein: Great point, thank you!

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: I don't think we need one. Everything in the `useEffect` runs synchronously and must complete before the clean up method can be called b/c of an unmounting. So `createInstance` will always be called before the clean up can be called, but the `async` operation may complete after both the `useEffect` completes AND the component has unmounted.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Same as above, I don't think we need to update here.

kand (approved):

PR #718, opened 2025-10-30
feat: add `version` to `PayPalV6Namespace`

nbierdeman:
This PR adds `version` to `PayPalV6Namespace`.

### DIFF
--- a/packages/paypal-js/types/v6/tests/paypal-legacy-billing-agreements.test.ts
+++ b/packages/paypal-js/types/v6/tests/paypal-legacy-billing-agreements.test.ts
@@ -25,12 +25,6 @@ async function main() {
         throw new Error("PayPal v6 namespace missing version property");
     }
 
-    // verify version has correct format
-    const versionRegex = /^\d+\.\d+\.\d+$/;
-    if (!versionRegex.test(paypal.version)) {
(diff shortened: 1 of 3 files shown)


packages/paypal-js/types/v6/tests/paypal-legacy-billing-agreements.test.ts
```diff
@@ -21,6 +21,16 @@ async function main() {
         throw new Error("Invalid paypal object for v6");
     }
 
+    if (!paypal?.version) {
+        throw new Error("PayPal v6 namespace missing version property");
+    }
+
+    // verify version has correct format
+    const versionRegex = /^\d+\.\d+\.\d+$/;
+    if (!versionRegex.test(paypal.version)) {
```
gregjopa: Right now these tests are only evaluated by the typescript complier to see if there are any type errors with common integration code. The JS code is not actually executed so a runtime check like this won't actually provide value. I think we should remove this.
```suggestion
```

nbierdeman: Good catch! I made that mistake earlier today, but then forgot again when working on this. I just removed the regex tests.

gregjopa (approved):

PR #719, opened 2025-10-30
feat: update loadCoreSdkScript to handle dataNamespace option

EvanReinstein:
Adds `dataNamespace` as an argument to `loadCoreSdkScript` and namespace validation.

### DIFF
--- a/.changeset/shiny-wasps-start.md
+++ b/.changeset/shiny-wasps-start.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
+---
+
+Update loadCoreSdkScript function to handle data-namespace attribute.


packages/paypal-js/src/v6/index.ts
```diff
@@ -20,16 +21,27 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
         url.searchParams.append("debug", "true");
     }
 
+    if (dataNamespace) {
+        attributes["data-namespace"] = dataNamespace;
+    }
```
EvanReinstein, line 26: V5 leverages a helper called `processOptions` to return an `attributes` object, which we could use if we needed later on. As it stands `data-namespace` is the only data attribute we want to handle.

kand (approved):

packages/paypal-js/src/v6/index.ts
```diff
@@ -20,16 +21,27 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+
     return new Promise<PayPalV6Namespace>((resolve, reject) => {
         insertScriptElement({
             url: url.toString(),
+            attributes,
             onSuccess: () => {
-                if (!window.paypal) {
+                const namespace = dataNamespace ?? "paypal";
+                const paypalSDK = Reflect.get(
```
gregjopa, line 34: What is the benefit of using `Reflect.get()` here vs `window[namespace]`? This is my first time seeing [Reflect.get](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/get)

gregjopa (approved):
Nice work @EvanReinstein! 💯

packages/paypal-js/src/v6/index.ts
nbierdeman: `Reflect.get()` allows you to have a separate `target` and `receiver` when accessing properties, but that isn't the case here, so not sure. [Here's](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/get#using_reflect.get) an example.

EvanReinstein: Great question and to be perfectly transparent it is the first time I'm using it as well. I went through a few iterations to essentially do `window[namespace]` but was encountering type errors. Reflect is a type-safe way to get objects on the window dynamically, although this usage is not the premier use case for Reflect.

gregjopa: Sorry to be a pain, but I'd like to remove the usage of `Reflect.get()` here. It changes runtime behavior: https://stackoverflow.com/questions/44639309/javascript-difference-between-reflect-get-and-objfoo. I'm happy to help with typing `window[namespace]`.

PR #721, opened 2025-10-30
v6 React error handling

kand:


packages/react-paypal-js/src/v6/context/PayPalProviderContext.tsx
```diff
@@ -46,4 +46,7 @@ export function instanceReducer(
     }
 }
 
-export const PayPalContext = createContext<PayPalContextState | null>(null);
+export const PayPalContext = createContext<PayPalContextValue>({
```
kand: This wasn't quite the change I was looking for, but I think these types are a little bit cleaner now. `PayPalContextValue` inherits from `PayPalState` and the context will never be `null`.

packages/react-paypal-js/.eslintrc.json
```diff
@@ -27,6 +27,7 @@
         }
     },
     "rules": {
+        "jsdoc/no-undefined-types": 1,
```
kand, line 30: This allows importing types only for use in `jsdoc` without the linter reporting an unused variable error.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -10,25 +10,22 @@ import {
...
     PayPalV6Namespace,
     SdkInstance,
 } from "../types";
-import type { InstanceAction } from "../context/PayPalProviderContext";
 
-export interface PayPalContextState {
-    sdkInstance: SdkInstance<readonly [Components, ...Components[]]> | null;
-    eligiblePaymentMethods: EligiblePaymentMethodsOutput | null;
-    error: Error | null;
-    dispatch: React.Dispatch<InstanceAction>;
```
kand, line 29: `dispatch` was removed from `state` since it's only necessary for `PayPalProvider` internals. Everywhere else, `PayPalContextState` has been replaced with `PayPalState`.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -30,17 +31,26 @@ export function usePayLaterOneTimePaymentSession({
...
     const sessionRef = useRef<OneTimePaymentSession | null>(null); // handle cleanup
     const proxyCallbacks = useProxyProps(callbacks);
+    const [error, setError] = useState<Error | null>(null);
 
     const handleDestroy = useCallback(() => {
         sessionRef.current?.destroy();
         sessionRef.current = null;
     }, []);
 
+    // Separate error reporting effect to avoid infinite loops with proxyCallbacks
```
kand, line 44: IMO a bit of a bummer that we need this separate effect, but I'm not sure another way we can prevent an infinite re-render loop with `proxyProps` and `setError` if a merchant passes in a new function for every render of one of these hooks.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -57,8 +67,13 @@ export function usePayLaterOneTimePaymentSession({
     }, []);
 
     const handleClick = useCallback(async () => {
+        if (!isMountedRef.current) {
+            return;
```
kand, line 71: Since we can't call `setError` if the component is unmounted, I decided here to instead do nothing. I don't know that we need an error here since there's no impact. We could maybe `console.warn` the merchant here that an event is occurring on an unmounted component, but I dunno, feels a bit unnecessary.

kand (commented):
Info for reviewers.

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
```diff
@@ -563,25 +545,12 @@ describe("Auto-memoization", () => {
     });
 });
 
-describe("usePayPal", () => {
```
kand, line 566: I moved this to its own test suite for `usePayPal`.

PR #727, opened 2025-11-05
feat(types): add v6 types for paypal-guest-payments

gregjopa:


packages/paypal-js/types/v6/components/paypal-guest-payments.d.ts
```diff
@@ -0,0 +1,44 @@
...
+    PresentationModeOptionsForAuto,
+} from "./base-component";
+
+export type PayPalGuestOneTimePaymentSessionOptions =
+    BasePaymentSessionOptions & {
+        orderId?: string;
+        onApprove: (data: OnApproveDataOneTimePayments) => Promise<void>;
+    };
+
+export type PayPalGuestPresentationModeOptions = PresentationModeOptionsForAuto;
```
EvanReinstein: Would it be the right move to update this to something like 
```
export type PayPalGuestPresentationModeOptions = PresentationModeOptionsForAuto & {
  targetElement: string | EventTarget;
};
```

EvanReinstein (approved):

packages/paypal-js/types/v6/components/paypal-guest-payments.d.ts
gregjopa: ah yes. This is used for the auto open on page load behavior, right? I see it used here: https://github.com/paypal-examples/v6-web-sdk-sample-integration/blob/main/client/components/paypalGuestPayments/html/src/onload/app.js#L10

I'll add it as an optional type.

packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -15,7 +15,19 @@ export type OnCompleteData = {
     paymentSessionState: "approved" | "canceled" | "error";
 };
 
-export type OnErrorData = Error;
+export type PayPalError = Error & {
```
gregjopa: Happy to update these to use `interface` if that makes more sense.

gregjopa: @kand @EvanReinstein I switched to `interface` for PayPalError and PayPalWarning

nbierdeman (approved):
Nice work, @gregjopa! 🚢

PR #732, opened 2025-11-06
Feat: Enhance useXPaymentSession hooks to include all Presentation Mode Options

EvanReinstein:
Adds the following as props to the `useXPaymentSession` V6 React Hooks:
- `autoRedirect`
- `fullPageOverlay`

Some background on this update. Most of the session hooks leverage a discriminating union `PayPalPresentationModeOptions`
```
type PayPalPresentationModeOptions =
    | PresentationModeOptionsForAuto
    | PresentationModeOptionsForPopup
    | PresentationModeOptionsForModal
    | PresentationModeOptionsForPaymentHandler
    | PresentationModeOptionsForRedirect;
```
As a stand alone type this union does a perfect job of catching type errors, such as mixing `presentationMode = "modal` and `fullPageOverlay = {enabled: true}`. 🎉 

However, when it is incorporated into the types used for the `useXPaymentSession` props, TS can't guarantee which presentation mode will be used so it only allows properties that exists on all parts of the union, functionally resulting in `autoRedirect`, `fullPageOverlay` and others being typed as `any`. This is a result of intersecting with a union, which can create some headaches.

This led to this change:
- Update the PresentationModeOptions props with every key and type them as `undefined`. This solves the problem in the `useXPaymentSession` use-case and clearly indicates which keys should be used with the appropriate presentation mode.
- The `never` keyword is reserved for a scenario where we want to indicate that the presence of one property in a type means that another property should never exist.

### DIFF
--- a/packages/paypal-js/types/v6/components/base-component.d.ts
+++ b/packages/paypal-js/types/v6/components/base-component.d.ts
@@ -44,35 +44,30 @@ export type PresentationModeOptionsForPopup = {
     presentationMode: "popup";
     fullPageOverlay?: { enabled: boolean };
     autoRedirect?: undefined;
-    loadingScreen?: { label: string };
 };
 
 export type PresentationModeOptionsForModal = {
     presentationMode: "modal";
     fullPageOverlay?: undefined;
     autoRedirect?: undefined;
-    loadingScreen?: { label: string };
 };
 
 export type PresentationModeOptionsForRedirect = {
     presentationMode: "redirect";
     autoRedirect?: { enabled: boolean };
     fullPageOverlay?: undefined;
-    loadingScreen?: undefined;
 };
 
 export type PresentationModeOptionsForPaymentHandler = {
     presentationMode: "payment-handler";
     fullPageOverlay?: undefined;
     autoRedirect?: undefined;
-    loadingScreen?: undefined;
 };
 
 export type PresentationModeOptionsForAuto = {
     presentationMode: "auto";
     fullPageOverlay?: { enabled: boolean };
     autoRedirect?: undefined;
-    loadingScreen?: undefined;
 };
 
 export type CreateOrderPromise = Promise<{ orderId: string }>;


packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -43,31 +43,36 @@ export type BasePaymentSessionOptions = {
 export type PresentationModeOptionsForPopup = {
     presentationMode: "popup";
     fullPageOverlay?: { enabled: boolean };
-    autoRedirect?: never;
+    autoRedirect?: undefined;
```
hamzanasir, line 46: Should be the same type as defined in the SDK itself?

```
autoRedirect?: {
    enabled?: boolean;
  };
```

packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -43,31 +43,36 @@ export type BasePaymentSessionOptions = {
...
+    loadingScreen?: { label: string };
```
hamzanasir: I believe we can remove this option for now. Currently we only have one loading label to choose from and it's already set as the default experience even if this option isn't passed in to `.start()`

packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -43,31 +43,36 @@ export type BasePaymentSessionOptions = {
...
+    fullPageOverlay?: undefined;
+    autoRedirect?: undefined;
+    loadingScreen?: { label: string };
 };
 
 export type PresentationModeOptionsForRedirect = {
     presentationMode: "redirect";
     autoRedirect?: { enabled: boolean };
-    fullPageOverlay?: never;
+    fullPageOverlay?: undefined;
```
hamzanasir, line 58: I believe fullPageOverlay also has the nested enabled structure like autoRedirect 🤔

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
```diff
@@ -10,20 +10,23 @@ import {
...
-      });
+      })
+) &
+    PayPalPresentationModeOptions;
 
 export function usePayPalOneTimePaymentSession({
     presentationMode,
+    fullPageOverlay,
+    autoRedirect,
+    loadingScreen,
```
hamzanasir: What would these props be passed as? Normal boolean's or using the same input as the SDK's start with the nested object like { enabled: boolean }? So basically this:

```ts
const { handleClick } = useVenmoOneTimePaymentSession({
  fullPageOverlay: true,
  autoRedirect: false,
})
```

Or

```ts
const { handleClick } = useVenmoOneTimePaymentSession({
  fullPageOverlay: { enabled: true},
  autoRedirect: { enabled: false },
})
```

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
```diff
@@ -77,9 +80,12 @@ export function useVenmoOneTimePaymentSession({
             return;
         }
 
-        const startOptions: VenmoPresentationModeOptions = {
+        const startOptions = {
             presentationMode,
-        };
+            fullPageOverlay,
+            autoRedirect,
+            loadingScreen,
```
hamzanasir: I think we need to add these to the dependency array for this hook 🤔

EvanReinstein: So weird that my linter wasn't highlighting this for me earlier but is now.

packages/paypal-js/types/v6/components/base-component.d.ts
EvanReinstein: Great insight. I'll go ahead and remove this.

packages/paypal-js/types/v6/components/base-component.d.ts
EvanReinstein: In the case of Popup we want to specifically say that `autoRedirect` property should be `undefined`. This helps the discriminating union `PayPalPresentationModeOptions` to providing typing for all properties referenced when it is used to compose the session props.

Typescript can't guarantee properties that don't exist consistently within a discriminating union, when that union is used in an intersection, which is how we are using it in `PayPalPresentationModeOptions`.

packages/paypal-js/types/v6/components/base-component.d.ts
EvanReinstein: Correct! See my other comment related to `autoRedirect: undefined` above.

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
EvanReinstein: Good question. Currently this is typed as the latter in order to more closely align with the SDK.

packages/paypal-js/types/v6/components/base-component.d.ts
hamzanasir: Ahh that's right! Thank you for the explanation @EvanReinstein!

packages/paypal-js/types/v6/components/base-component.d.ts
hamzanasir: Thanks!

packages/paypal-js/types/v6/components/base-component.d.ts
hamzanasir: Perfect sorry for reading this incorrectly!

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
hamzanasir: Makes sense! Thanks!

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
hamzanasir: Ahh yeah editer linters are wonky sometimes. Thanks!

PR #735, opened 2025-11-12
v6 React: unify error handling

kand:
Unify error handling in v6 react to `console.log` anytime there's an error.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/useError.ts
+++ b/packages/react-paypal-js/src/v6/hooks/useError.ts
@@ -3,7 +3,7 @@ import { useCallback, useState } from "react";
 /**
  * Sets the `error` returned by {@link useError}. Also, calls `console.error` with the given {@link Error}.
  */
-type TypeSetError = (error: Error) => void;
+type TypeSetError = (error: unknown) => void;
 
 /**
  * Centralized hook for handling {@link Error}s in a consistent manner.
@@ -19,7 +19,7 @@ export function useError(
         (newError) => {
             setErrorInternal(newError);
 
-            if (newError instanceof Error && !noConsoleErrors) {
+            if (!noConsoleErrors) {
                 console.error(newError);
             }
         },


packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.ts
```diff
@@ -1,5 +1,6 @@
 import { act, renderHook } from "@testing-library/react-hooks";
 
+import { expectSetError } from "./useSetErrorTestUtil";
```
kand: Kinda wonky, but this import must happen before other imports that use `useSetError` so it can be properly mocked and spied on for testing.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -48,7 +49,7 @@ export function usePayLaterOneTimePaymentSession({
         if (!sdkInstance) {
             setError(new Error("no sdk instance available"));
         }
-    }, [sdkInstance]);
+    }, [sdkInstance, setError]);
```
kand, line 52: `setError` is technically a stable ref, so it has no effect as a part of the dependency array and makes the linter happy.

packages/react-paypal-js/src/v6/hooks/useError.ts
```diff
@@ -0,0 +1,30 @@
...
+    noConsoleErrors = false,
+): [Error | null, TypeSetError] {
+    const [error, setErrorInternal] = useState<Error | null>(null);
+
+    const setError = useCallback(
+        (newError) => {
+            setErrorInternal(newError);
+
+            if (newError instanceof Error && !noConsoleErrors) {
+                console.error(newError);
```
kand, line 23: We could presumably add other stuff here as well if we wanted to, like additional callbacks for merchants.

packages/react-paypal-js/src/v6/hooks/useError.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+type TypeSetError = (error: Error) => void;
+
+/**
+ * Centralized hook for handling {@link Error}s in a consistent manner.
+ *
+ * @param {Boolean} noConsoleErrors - set to `true` to prevent `setError` calls from logging to `console.error`.
+ */
+export function useSetError(
+    noConsoleErrors = false,
```
kand, line 14: I'm not totally into having this flag available, but it's hard to prevent `console.error` calls in test otherwise without entirely mocking out `console.error` which I think is worse than this setup.

packages/react-paypal-js/src/v6/hooks/useSetErrorTestUtil.ts
```diff
@@ -0,0 +1,19 @@
+import { useSetError } from "./useSetError";
+
+// Any module importing this test helper will spy on useSetError and prevent it from calling console.error. Note,
+// this module must be imported in a test file *BEFORE* other imports that use the useSetError hook. Otherwise,
+// the module mock will not happen soon enough.
+jest.mock("./useSetError", () => ({
```
kand: This is the `mock` that requires this test utility module to be imported before anything being tested imports `useSetError`.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -51,6 +52,8 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
     const [state, dispatch] = useReducer(instanceReducer, initialState);
     // Ref to hold script options to avoid re-running effect
     const loadCoreScriptOptions = useRef(scriptOptions);
+    // Using the error hook here so it can participate in side-effects provided bhy the hook
+    const [, setError] = useSetError();
```
kand: I dunno if I'm a huge fan of this. I could take it or leave it. Alternatively, I could just add the `console.error` calls inline in this file. The one benefit of using `useSetError` is all the errors are gonna go through it. What do you all think?

packages/react-paypal-js/src/v6/hooks/useError.ts
EvanReinstein: Maybe leave a comment explaining the reasoning behind having this flag.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Hmm yea interesting conundrum here. Maybe leave a comment indicating that `useSetError` is explicitly being used here to trigger `console.error` rather than returning the `error` object.

packages/react-paypal-js/src/v6/hooks/useError.ts
hamzanasir: Why do we need to test for the fact that it shouldn't console.log the error? Can we just not remove this flag and the `should not console.log the error if noConsoleErrors is true` test? Just trying to understand this setup a little bit more 🙂

packages/react-paypal-js/src/v6/hooks/useSetError.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Sets the `error` returned by {@link useSetError}. Also, calls {@link console.error} with the given {@link Error}.
+ */
+type TypeSetError = (error: Error) => void;
+
+/**
+ * Centralized hook for handling {@link Error}s in a consistent manner.
+ *
+ * @param {Boolean} noConsoleErrors - set to `true` to prevent `setError` calls from logging to `console.error`.
+ */
+export function useSetError(
```
hamzanasir: Very very nitpicky but can we name this something generic like `useError()`? Cool with `useSetError()` as well the only thing is that it's not clearly obvious that you can also get the error using this hook.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
hamzanasir: I'm cool with keeping it to be consistent with our error strategy! Who knows we might actually have to check `error` in this component later

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.ts
```diff
@@ -244,7 +249,11 @@ describe("usePayLaterOneTimePaymentSession", () => {
             await result.current.handleClick();
         });
 
-        expect(result.current.error).toBeNull();
+        const { error } = result.current;
+
+        expectSetError(error);
```
hamzanasir: Why would `setError()` be called if there's no error?

EvanReinstein (approved):
This looks great @kand ! Thanks for taking this on!

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
kand: I updated the comment, let me know if ya'll are good with it.

> is explicitly being used here to trigger console.error rather than returning the error object.

@EvanReinstein I made the comment slightly more generic because I imagine at some point we might have more side-effects than just `console.log`.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.ts
kand: That's a good question 😅 let me double check what I was trying to do here.

packages/react-paypal-js/src/v6/hooks/useSetError.ts
kand: Oh yea, `useError` would be a better name! I'll update. The return type should hint to the return type as well.

packages/react-paypal-js/src/v6/hooks/useSetError.test.ts
```diff
@@ -0,0 +1,42 @@
...
+        act(() => {
+            result.current[1](mockError);
+        });
+
+        expect(mockConsoleError).toHaveBeenCalledTimes(1);
+        expect(mockConsoleError).toHaveBeenCalledWith(mockError);
+        expect(result.current[0]).toBe(mockError);
+    });
+
+    test("should not console.log the error if noConsoleErrors is true", () => {
```
kand:
```suggestion
    test("should not console.error the error if noConsoleErrors is true", () => {
```

packages/react-paypal-js/src/v6/hooks/useError.ts
kand: > Maybe leave a comment explaining the reasoning behind having this flag.

There's a comment in the function's JS doc, but would it be helpful to have another comment above this argument?

> Why do we need to test for the fact that it shouldn't console.log the error?

I could remove that test, but I have it just to completely test the function. The flag is available to prevent this specific function from logging `console.error` without needing to entirely mock out `console.error`, in case any other code is doing a `console.error` that should show up in test results. If there are thoughts on a better way to set this up, I'm open to ideas cause I don't think this is ideal.

packages/react-paypal-js/src/v6/hooks/useSetError.ts
kand: Done

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.test.ts
kand: I renamed the utility function to hopefully better clarify its purpose. The expectation is for the `error` value returned by the hook. So, in this case, there hasn't been an error, so the `error` returned by the hook is `null`.

Let me know if that makes sense!

PR #736, opened 2025-11-13
Add npm run dev watch script

hamzanasir:
This PR adds an `npm run dev` script which watches `src` files for changes and rebuilds accordingly


kand (approved):
Awesome! I was just thinking about this the other day, thanks for putting it together!

packages/react-paypal-js/package.json
```diff
@@ -31,6 +31,7 @@
     },
     "scripts": {
         "build": "rollup --config --bundleConfigAsCjs",
+        "dev": "rollup --config --bundleConfigAsCjs --watch",
```
kand: Should this also run `prepack` so types are updated? Or does this update types?

hamzanasir: I believe this isn't needed for development since it might just be needed for the release to npm 🤔

hamzanasir: Actually good callout @kand just pushed a fix for this!

PR #738, opened 2025-11-14
Feat: Add V6 Find Eligible Methods Hook

EvanReinstein:


packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -25,6 +30,8 @@ type PayPalProviderProps = CreateInstanceOptions<
     readonly [Components, ...Components[]]
 > &
     LoadCoreSdkScriptOptions & {
+        eligibleMethodsResponse?: FindEligiblePaymentMethodsResponse;
+        eligibleMethodsPayload?: FindEligiblePaymentMethodsRequestPayload; // Add this
```
kand: Does the `// Add this` comment need to be addressed?

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -0,0 +1,187 @@
...
+    national_number?: string;
+};
+
+type PaymentFlow =
+    | "ONE_TIME_PAYMENT"
+    | "RECURRING_PAYMENT"
+    | "VAULT_WITHOUT_PAYMENT"
+    | "VAULT_WITH_PAYMENT";
+
+export type FindEligiblePaymentMethodsRequestPayload = {
```
kand, line 24: Are there any required properties for this request? Or properties we want to make required from the merchant perspective?

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -0,0 +1,220 @@
...
+    environment,
+}: FindEligiblePaymentMethodsOptions): {
+    eligibleMethods: FindEligiblePaymentMethodsResponse | null;
+    isLoading: boolean;
+    error: Error | null;
+} {
+    const [eligibleMethods, setEligibleMethods] =
+        useState<FindEligiblePaymentMethodsResponse | null>(null);
+    const [isLoading, setIsLoading] = useState(true);
+    const [error, setError] = useState<Error | null>(null);
```
kand: Is the plan to swap this out for `useSetError` when https://github.com/paypal/paypal-js/pull/735 is merged?

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -139,3 +139,44 @@ export function toError(error: unknown): Error {
...
+
+    if (keys1.length !== keys2.length) {
+        return false;
+    }
+
+    for (const key of keys1) {
+        if (!keys2.includes(key)) {
+            return false;
+        }
+        if (!deepEqual(obj1[key], obj2[key])) {
```
kand: Is there any reason to protect against a max call stack error? I'm not sure if there are possible exploits here, but I think we can assume some reasonable recursion depth, like 10, right? I.e. no object we use this on would be deeper than 10 levels. I dunno if 10 makes sense, but what do you think about having some boundary?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Good catch, no more work to be done here 👍

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
EvanReinstein: No, there aren't any required properties. I'm not sure if we want to make any of these required, which would alter the API contract.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
EvanReinstein: Yessir!

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein: Yea that is a good idea. This function is very generic right now and I wouldn't mind keeping it that way, however if we want to narrow its focus to evaluating only the eligible methods payload or the response then we'd at least know more of what to expect in terms of recursion depth. What do you think?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -56,6 +65,22 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+    });
+
+    useEffect(() => {
+        if (!isLoading && eligibleMethods) {
+            dispatch({
+                type: INSTANCE_DISPATCH_ACTION.SET_ELIGIBILITY,
+                value: eligibleMethods,
+            });
+        }
+    }, [isLoading, eligibleMethods]);
```
nityasp, line 82: I think we have to also check for eligibility data changes too. For example, if the user changes the shipping address. 

```suggestion
    useEffect(() => {
        if (!isLoading && eligibleMethods && eligibleMethods !== state.eligiblePaymentMethods) {
            dispatch({
                type: INSTANCE_DISPATCH_ACTION.SET_ELIGIBILITY,
                value: eligibleMethods,
            });
        }
    }, [isLoading, eligibleMethods, state.eligiblePaymentMethods]);
```

packages/react-paypal-js/src/v6/utils.ts
kand: I'm cool with keeping it generic, but could we add a max recursion depth parameter that defaults to something reasonable?

EvanReinstein: Yea absolutely!

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Yes, we 100% need to ensure that we react appropriately to an updated/changed `eligibleMethodsPayload` which I assume is where a change would occur if shipping address changes.

Although we don't do anything in the provider, we do monitor the payload in the eligible methods hook itself. So if the provider receives a new payload this is what happens:

1. provider gets new `eligibleMethodsPayload` and passes to the hook.
2. hook uses `deepMemoizeCompare` in order to determine if the payload is new, in this case it is 👍 
3. new payload triggers the hook's `useEffect` that will re-query for eligibility.
4. the new `eligibleMethods` are passed back up to the provider and triggers the `useEffect` to dispatch and update the context.

Let me know if you see any problems with that breakdown, happy to huddle on this!

nityasp: Thanks @EvanReinstein for clarifying this over the huddle.

PR #739, opened 2025-11-17
Enhacing PayPal and PayLater hooks to handle redirect and appswitch presentation modes

nityasp:
This PR is for enhacing PayPal and PayLater react hooks to handle redirect and appswitch presentation modes.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
@@ -61,24 +61,31 @@ export function usePayLaterOneTimePaymentSession({
         });
         sessionRef.current = newSession;
 
-        const handleReturnFromPayPal = async () => {
-            try {
-                if (!newSession) {
-                    return;
-                }
-                const isResumeFlow = newSession.hasReturned?.();
-                if (isResumeFlow) {
-                    await newSession.resume?.();
-                }
-            } catch (err) {
-                setError(err as Error);
-            }
-        };
+        // check for resume flow in redirect-based presentation modes
+        const isRedirectMode =
+            presentationMode === "redirect" ||
+            presentationMode === "direct-app-switch";
 
-        handleReturnFromPayPal();
+        if (isRedirectMode) {
+            const handleReturnFromPayPal = async () => {
+                try {
+                    if (!newSession) {
+                        return;
+                    }
+                    const isResumeFlow = newSession.hasReturned?.();
+                    if (isResumeFlow) {
+                        await newSession.resume?.();
+                    }
+                } catch (err) {
+                    setError(err as Error);
+                }
+            };
+
+            handleReturnFromPayPal();
+        }
 
         return handleDestroy;
-    }, [sdkInstance, orderId, proxyCallbacks, handleDestroy]);
+    }, [sdkInstance, orderId, proxyCallbacks, handleDestroy, presentationMode]);
 
     const handleCancel = useCallback(() => {
         sessionRef.current?.cancel();
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
@@ -65,24 +65,31 @@ export function usePayPalOneTimePaymentSession({
 
         sessionRef.current = newSession;
 
-        const handleReturnFromPayPal = async () => {
-            try {
-                if (!newSession) {
-                    return;
-                }
-                const isResumeFlow = newSession.hasReturned?.();
-                if (isResumeFlow) {
-                    await newSession.resume?.();
-                }
-            } catch (err) {
-                setError(err as Error);
-            }
-        };
+        // Only check for resume flow in redirect-based presentation modes
+        const shouldCheckResume =
+            presentationMode === "redirect" ||
+            presentationMode === "direct-app-switch";
 
-        handleReturnFromPayPal();
+        if (shouldCheckResume) {
+            const handleReturnFromPayPal = async () => {
+                try {
+                    if (!newSession) {
+                        return;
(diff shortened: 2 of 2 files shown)


packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -61,6 +61,19 @@ export function usePayLaterOneTimePaymentSession({
         });
         sessionRef.current = newSession;
 
+        const handleReturnFromPayPal = async () => {
+            const isResumeFlow = newSession.hasReturned?.();
```
kand: Should this entire function body be wrapped in a `try` / `catch`? `hasReturned` seems pretty safe, but just to be sure we catch any possible error our SDK would throw, it might be safer to wrap everything.

packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -55,7 +55,7 @@ export type PresentationModeOptionsForModal = {
 export type PresentationModeOptionsForRedirect = {
     presentationMode: "redirect";
     autoRedirect?: { enabled: boolean };
-    fullPageOverlay?: undefined;
```
EvanReinstein, line 58: Does this presentation mode accept/leverage `fullPageOverlay`?  Technically I think that one can pass `fullPageOverlay` to `.start` but the SDK won't honor it in this use case, which is why `PresentationModeOptionsForRedirect` originally was just:
```
type PresentationModeOptionsForRedirect = {
  presentationMode: "redirect";
  autoRedirect?: { enabled: boolean };
}
```

We want to use these types to prevent "incorrect" combinations, even if the SDK will still accept the argument but not use it behind the scenes.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -85,9 +98,14 @@ export function usePayLaterOneTimePaymentSession({
...
-            await sessionRef.current.start(startOptions, createOrder());
+            const result = await sessionRef.current.start(
+                startOptions,
+                createOrder(),
+            );
+            return result;
         } else {
-            await sessionRef.current.start(startOptions);
+            const result = await sessionRef.current.start(startOptions);
+            return result;
```
kand: We have this setup in every session hook, but I believe this could be shortened:

```suggestion
const result = await sessionRef.current.start(
    startOptions,
    createOrder?.()
);
return result;
```

cc @EvanReinstein should we do this with other session hooks as well?

packages/paypal-js/types/v6/components/base-component.d.ts
EvanReinstein: Oh, I see `fullPageOverlay` being leveraged in the `RedirectPresentationMode` file.  Ignore me!

EvanReinstein: I guess I'm still wondering why `fullPageOverlay` was left out in the first place 🤔

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
EvanReinstein: Oh this is a tidy enhancement, I like it!

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -61,6 +61,22 @@ export function usePayLaterOneTimePaymentSession({
...
+        const handleReturnFromPayPal = async () => {
+            if (!newSession) {
+                return;
+            }
+            const isResumeFlow = newSession.hasReturned?.();
+            if (isResumeFlow) {
+                try {
+                    await newSession.resume?.();
+                } catch (err) {
+                    setError(err as Error);
```
hamzanasir: Hmm I don't think we should set an error if `hasReturned` returns `false`. That could mean that they're just initializing checkout for the first time

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
nityasp: @kand, I wrapped this in try/catch now.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
nityasp: Yup I cleaned up this.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
hamzanasir: Actually nevermind this is try catching the error!

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -61,6 +61,19 @@ export function usePayLaterOneTimePaymentSession({
...
+                }
+            }
+        };
+
+        handleReturnFromPayPal();
```
hamzanasir: Should we only do this when the presentation mode is redirect/direct app switch? Seems like an unnecessary code path for all other presentation modes 🤔

nityasp: Good callout @hamzanasir . I applied the check for presentation mode now.

hamzanasir: Thank you!

EvanReinstein (approved):
This looks good to me, thanks for taking this up @nityasp 🚀

PR #743, opened 2025-11-21
BCDC hook implementation

nityasp:
React hook for guest payment session (BCDC).

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
@@ -34,6 +34,22 @@ export type UsePayPalGuestPaymentSessionProps = (
 ) &
     PayPalGuestPresentationModeHookOptions;
 
+/**
+ * `usePayPalGuestPaymentSession` is used to interface with a guest checkout session. Guest checkout
+ * sessions require a `<paypal-basic-card-button>` to target for displaying the guest checkout form.
+ *
+ * @example
+ * const { buttonRef, error, handleClick } = usePayPalGuestPaymentSession({ ...arguments });
+ *
+ * return (
+ *   <paypal-basic-card-button
+ *     onClick={handleClick}
+ *     ref={buttonRef}
+ *     {...otherProps}>
+ *   </paypal-basic-card-button>
+ * )
+ *
+ */
 export function usePayPalGuestPaymentSession({
     fullPageOverlay,
     createOrder,


packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,199 @@
...
+    | (PayPalGuestOneTimePaymentSessionOptions & {
+          createOrder?: never;
+          orderId: string;
+      })
+) & {
+    buyerCountry?: string;
+    targetElement?: string | HTMLElement;
+    presentationMode?: "auto";
+    fullPageOverlay?: boolean;
+    autoRedirect?: boolean;
```
EvanReinstein: If these are the guest payment presentation mode `.start` options I think it is fair to do a `paypal-js` change to consolidate this as a type.

Ideally any change that is in `paypal-js` should be a separate PR.

packages/react-paypal-js/src/v6/types/index.ts
```diff
@@ -3,8 +3,10 @@ export * from "./PayPalProviderEnums";
 export type * from "@paypal/paypal-js/sdk-v6";
 
 export interface BasePaymentSessionReturn {
+    buttonRef?: { current: HTMLElement | null };
     error: Error | null;
     handleClick: () => Promise<{ redirectURL?: string } | void>;
     handleCancel: () => void;
     handleDestroy: () => void;
+    isProcessing?: boolean;
```
EvanReinstein: If the properties `buttonRef` and `isProcessing` are unique to BCDC we could do something like 
```
interface PayPalGuestPaymentReturn extends BasePaymentSessionReturn & {
  buttonRef?: { current: HTMLElement | null };
  isProcessing?: boolean;
}
```

and it can be defined in the hook's file 👍

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,199 @@
...
+        ) => {
+            if (!callback) {
+                return callback;
+            }
+
+            return async (...args: Parameters<T>) => {
+                try {
+                    await callback(...args);
+                } finally {
+                    resetState();
```
kand: Under what conditions does this "reset button state"? Reading this, I think this will only reset state if the callback throws an error.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,199 @@
...
+    const isMountedRef = useIsMountedRef();
+    const sessionRef = useRef<PayPalGuestOneTimePaymentSession | null>(null);
+    const buttonRef = useRef<HTMLElement>(null);
+    const proxyCallbacks = useProxyProps(callbacks);
+    const [error, setError] = useError();
+    const [isProcessing, setIsProcessing] = useState(false);
+    const isSessionActiveRef = useRef(false);
+
+    // Track whether shipping callbacks are present to trigger session recreation
+    const hasShippingCallbacks = Boolean(
```
kand: [`useProxyProps`](https://github.com/paypal/paypal-js/blob/940ee299836b45bcf71b55a5946b28ad7aec8747/packages/react-paypal-js/src/hooks/useProxyProps.ts#L9) prevents changes to `callbacks` triggering a component re-render. If I'm understanding `hasShippingCallbacks`, it appears to be reversing the `useProxyProps` functionality. Would the same effect be achieved by unpacking the shipping callbacks from the argument on line 39, then having those be dependencies to the session creation `useEffect`?

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,199 @@
...
+    targetElement,
+    ...callbacks
+}: UsePayPalGuestPaymentSessionProps): BasePaymentSessionReturn {
+    const { sdkInstance } = usePayPal();
+    const isMountedRef = useIsMountedRef();
+    const sessionRef = useRef<PayPalGuestOneTimePaymentSession | null>(null);
+    const buttonRef = useRef<HTMLElement>(null);
+    const proxyCallbacks = useProxyProps(callbacks);
+    const [error, setError] = useError();
+    const [isProcessing, setIsProcessing] = useState(false);
```
kand: I don't see the `isProcessing` value used anywhere. What would this be used for?

kand (commented):
A good start! A few comments now, but I'll continue the review on Monday!

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
kand: BCDC has `inline`, `popup`, and `modal` presentation modes, so any options those can use need to be available. I think `fullPageOverlay` is valid because it's used by `popup`. I'm not sure about `autoRedirect` though, I'm only seeing that in relation to redirect presentation modes. Could you verify if this is an option BCDC needs?

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: Initialy I used `isProcessing` manually disable the button for testing purposes by adding it as a attribute value. Since SDK takes care of disabling the button when the form is open, we no longer need this. Thanks for pointing out. I'll remove it.

packages/react-paypal-js/src/v6/types/index.ts
nityasp: Right. `buttonRef` is specific to BCDC. I'll keep it in hook itself.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: I was using this "reset button state" to manually reset button after payment flow is completed. We dont need this since SDK is handling the button re-enabling here. https://github.paypal.com/PayPal-R/core-web-sdk/blob/8cc18742de053cf3a8014f13a086d2299526fa96/packages/web-sdk-payments/src/CheckoutSession/PayPalGuestCheckoutSession.ts#L239

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: Great point!! That seems to be a much cleaner way to destructure the shipping callbacks from props and passing then to session creation conditionally. Later use them as dependencies to trigger session creation.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: Removed the`autoRedirect` option.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,159 @@
...
+    targetElement?: string | HTMLElement;
+    presentationMode?: "auto";
+    fullPageOverlay?: boolean;
+    onShippingAddressChange?: (
+        data: OnShippingAddressChangeData,
+    ) => Promise<void>;
+    onShippingOptionsChange?: (
+        data: OnShippingOptionsChangeData,
+    ) => Promise<void>;
+};
```
EvanReinstein: ~~I think that `presentationMode` should include `inline`, `popup`, and `modal`.~~ I'm also seeing a `PayPalGuestPresentationModeOptions` [defined in](https://github.com/paypal/paypal-js/blob/71f597c2e362ed4e666882485f54294a2fa429bf/packages/paypal-js/types/v6/components/paypal-guest-payments.d.ts#L16-L19) `paypal-js` that looks like:
```
export type PayPalGuestPresentationModeOptions =
  PresentationModeOptionsForAuto & {
      targetElement?: string | EventTarget;
};
```

I'm wondering if we should update that type to include some of the properties that are not currently in `PayPalGuestPresentationModeOptions` but are used in the hook, such as `onShippingOptionsChange`/`onShippingAddressChange` and `targetElement` or `buyerCountry`.

EvanReinstein: Huddled on this with @kand and I'll open a PR to update `PayPalGuestPresentationModeOptions` with some of the other relevant properties 👍 

We also discussed hardcoding the `presentationMode` and removing it as an argument for the hook.

EvanReinstein:
```suggestion
type PayPalGuestPresentationModeHookOptions = Omit<PayPalGuestPresentationModeOptions, "presentationMode">;

export type UsePayPalGuestPaymentSessionProps = (
    | (Omit<PayPalGuestOneTimePaymentSessionOptions, "orderId"> & {
          createOrder: () => PayPalGuestOneTimePaymentSessionPromise;
          orderId?: never;
      })
    | (PayPalGuestOneTimePaymentSessionOptions & {
          createOrder?: never;
          orderId: string;
      })
) & PayPalGuestPresentationModeHookOptions;
```

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,159 @@
...
+        if (isSessionActiveRef.current) {
+            return;
+        }
+
+        isSessionActiveRef.current = true;
+
+        try {
+            const target = targetElement || buttonRef.current;
+            const startOptions: PayPalGuestPresentationModeOptions = {
+                presentationMode,
```
EvanReinstein: Hard code the `presentationMode` in order to prevent passing values other than `auto`

```suggestion
                presentationMode: "auto",
```

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
EvanReinstein: I opened a PR to update the presentation mode options type [here](https://github.com/paypal/paypal-js/pull/749) 👋

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+        sessionRef.current?.destroy();
+        sessionRef.current = null;
+        isSessionActiveRef.current = false;
+    }, []);
+
+    useEffect(() => {
+        if (!sdkInstance) {
+            setError(new Error("no sdk instance available"));
+        }
+    }, [sdkInstance, setError]);
```
EvanReinstein: @kand Is making an update to other session hooks to be more mindful about when to set an error.
```suggestion
    useEffect(() => {
        if (sdkInstance) {
            setError(null);
        } else if (loadingStatus !== INSTANCE_LOADING_STATE.PENDING) {
            setError(new Error("no sdk instance available"));
        }
    }, [sdkInstance, setError, loadingStatus]);
```

kand: Ah good catch @EvanReinstein, thanks!

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.test.ts
```diff
@@ -0,0 +1,670 @@
...
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    describe("initialization", () => {
+        test("should error if there is no sdkInstance when called", () => {
+            mockUsePayPal.mockReturnValue({
+                sdkInstance: null,
+                loadingStatus: INSTANCE_LOADING_STATE.PENDING,
```
kand:
```suggestion
                loadingStatus: INSTANCE_LOADING_STATE.REJECTED,
```

I made this change in [my updates PR](https://github.com/paypal/paypal-js/pull/752/files#diff-fc36b6cad982cce103caac04e16d3bd69b31206544765a75a332ead22dab07acR59) since it tests only the case where `sdkInstance` is `null` while the `loadingStatus` check is `true`.

There are a couple other updates to this test as well you could copy over from my PR, including the name of the test and checking that the create session call is not made!

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.test.ts
```diff
@@ -0,0 +1,670 @@
...
+
+            const { result } = renderHook(() =>
+                usePayPalGuestPaymentSession(props),
+            );
+
+            act(() => {
+                result.current.handleCancel();
+            });
+
+            expect(mockPayPalSession.cancel).not.toHaveBeenCalled();
```
kand: What's the reason the session's `cancel` isn't called?

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.test.ts
```diff
@@ -0,0 +1,670 @@
...
+
+            const { result, unmount } = renderHook(() =>
+                usePayPalGuestPaymentSession(props),
+            );
+
+            unmount();
+
+            expect(() => {
+                act(() => {
+                    result.current.handleCancel();
```
kand, line 627: There currently isn't any logic using the session in the `handleCancel` callback, or that would throw an error. So, should this test be refactored? Or is there logic missing from `handleCancel`?

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+            return;
+        }
+
+        isSessionActiveRef.current = true;
+
+        try {
+            const target = targetElement || buttonRef.current;
+            const startOptions: PayPalGuestPresentationModeOptions = {
+                presentationMode: "auto",
+                ...(fullPageOverlay !== undefined && { fullPageOverlay }),
```
kand: Does `undefined` need to be checked specifically, or could this just be a truthy check? Also, could this be rewritten as a ternary like the line following it?

```suggestion
                ...(fullPageOverlay ? { fullPageOverlay } : {}),
```

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+                ? createOrder()
+                : undefined;
+            await sessionRef.current.start(
+                startOptions,
+                checkoutOptionsPromise,
+            );
+        } catch (err) {
+            if (isMountedRef.current) {
+                isSessionActiveRef.current = false;
+                setError(err instanceof Error ? err : new Error(String(err)));
```
kand: I believe this function should take `unknown`, [see it's type here](https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/v6/hooks/useError.ts#L6). Were you seeing a typescript error when this wasn't cast?

```suggestion
                setError(err);
```

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+        try {
+            const target = targetElement || buttonRef.current;
+            const startOptions: PayPalGuestPresentationModeOptions = {
+                presentationMode: "auto",
+                ...(fullPageOverlay !== undefined && { fullPageOverlay }),
+                ...(target ? { targetElement: target as EventTarget } : {}),
+            };
+            const checkoutOptionsPromise = createOrder
+                ? createOrder()
+                : undefined;
```
kand: Using optional chaining, I believe this could be shortened to:

```suggestion
            const checkoutOptionsPromise = createOrder?.();
```

This could be moved down to line 123 as well, same as the PayLater session is doing:

https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts#L120

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+
+import type {
+    PayPalGuestOneTimePaymentSession,
+    PayPalGuestOneTimePaymentSessionOptions,
+    PayPalGuestOneTimePaymentSessionPromise,
+    PayPalGuestPresentationModeOptions,
+} from "@paypal/paypal-js/sdk-v6";
+import type { BasePaymentSessionReturn } from "../types";
+
+export interface PayPalGuestPaymentSessionReturn
```
kand: Does this interface need to be exported? I don't see it used anywhere outside this file.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+    extends BasePaymentSessionReturn {
+    buttonRef: { current: HTMLElement | null };
+}
+type PayPalGuestPresentationModeHookOptions = Omit<
+    PayPalGuestPresentationModeOptions,
+    "presentationMode"
+>;
+
+export type UsePayPalGuestPaymentSessionProps = (
```
kand, line 25: Does this type need to be exported? If it's only exported for the test file, I think all the `const prop` constructions could just be inlined directly into their respective `usePayPalGuestPaymentSession` calls.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: we don't to check for undefined since fullPageOverlay is an optional parameter. I think we don't need a ternary option here. Simply passing it as is.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: To be consistent with other payment session hooks, I think we should keep it. Other hooks are exporting their props types.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: removed the export

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
kand: Ah good call. I think it's unnecessary to export this type, but it's fine if we're matching the other sessions. cc @EvanReinstein a bit of a nitpick maybe these can be removed in another PR.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,123 @@
...
+import type {
+    PayPalGuestOneTimePaymentSession,
+    PayPalGuestOneTimePaymentSessionOptions,
+    PayPalGuestOneTimePaymentSessionPromise,
+    PayPalGuestPresentationModeOptions,
+} from "@paypal/paypal-js/sdk-v6";
+import type { BasePaymentSessionReturn } from "../types";
+
+interface PayPalGuestPaymentSessionReturn
+    extends Omit<BasePaymentSessionReturn, "handleCancel"> {
```
kand: I may be misunderstanding the discussion, but don't we still want to return a `handleCancel` callback that calls the session `cancel` function?

nityasp: No my mistake, I removed it since earlier it was just reseting the session state. Looking at the other hooks, i think we still need `handleCancel` callback to actually cancel the payment session in the SDK.

EvanReinstein (approved):
Tested this out locally and it looks good to me 👍  Great work on this @nityasp 🥇

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.test.ts
nityasp: This is fixed now.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.test.ts
nityasp: I see the similar kind of test in all the payment session hooks. Its more of testing the optional chaining and doesn't throw when the session is unavailable.

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -0,0 +1,127 @@
...
+          orderId?: never;
+      })
+    | (PayPalGuestOneTimePaymentSessionOptions & {
+          createOrder?: never;
+          orderId: string;
+      })
+) &
+    PayPalGuestPresentationModeHookOptions;
+
+export function usePayPalGuestPaymentSession({
```
kand, line 53: @EvanReinstein I think we should document each session hook. This one in particular would be good to show how `buttonRef` would be used. Something like this:

```suggestion
/**
 * `usePayPalGuestPaymentSession` is used to interface with a guest checkout session. Guest checkout
 * sessions require a `<paypal-basic-card-button>` to target for displaying the guest checkout form.
 *
 * @example
 * const { buttonRef, error, handleClick } = usePayPalGuestPaymentSession({ ...arguments });
 * 
 * return (
 *   <paypal-basic-card-button
 *     onClick={handleClick}
 *     ref={buttonRef}
 *     {...otherProps}>
 *   </papal-basic-card-button>
 * )
 *
 */
export function usePayPalGuestPaymentSession({
```

nityasp: Added the comment.

nityasp: Added the comment.

PR #746, opened 2025-11-24
PayPalOneTimePaymentButton

kand:
Add `<PayPalOneTimePaymentButton>` which provides a standard button component to wrap the `usePayPalOneTimePaymentSession` hook and wire it up with a `<paypal-button>`.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
@@ -27,10 +27,12 @@ import type {
 import type { PayPalState } from "../context/PayPalProviderContext";
 import type { usePayPal } from "../hooks/usePayPal";
 
-type PayPalProviderProps = CreateInstanceOptions<
-    readonly [Components, ...Components[]]
+type PayPalProviderProps = Omit<
+    CreateInstanceOptions<readonly [Components, ...Components[]]>,
+    "components"
 > &
     LoadCoreSdkScriptOptions & {
+        components?: Components[];
         eligibleMethodsResponse?: FindEligiblePaymentMethodsResponse;
         eligibleMethodsPayload?: FindEligiblePaymentMethodsRequestPayload;
         children: React.ReactNode;
@@ -52,7 +54,7 @@ type PayPalProviderProps = CreateInstanceOptions<
 export const PayPalProvider: React.FC<PayPalProviderProps> = ({
     clientMetadataId,
     clientToken,
-    components,
+    components = ["paypal-payments"],
     locale,
     pageType,
     partnerAttributionId,
--- a/packages/react-paypal-js/src/v6/utils.ts
+++ b/packages/react-paypal-js/src/v6/utils.ts
@@ -7,7 +7,7 @@ import type { Components } from "./types";
  * or `document` are not available; `false` otherwise.
  */
 export function isServer(): boolean {
-    return typeof window === "undefined" && typeof document === "undefined";
+    return typeof window === "undefined";
 }
 
 /**
--- a/.changeset/blue-cars-peel.md
+++ b/.changeset/blue-cars-peel.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
+---
+
+Updates the paypal guest payments presentation mode options type with more properties.
--- a/.changeset/rude-melons-wear.md
+++ b/.changeset/rude-melons-wear.md
@@ -1,0 +1,8 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+- Default `PayPalProvider` `components` to `["paypal-payments"]`.
+- Update session hooks to check `loadingStatus` before returning an error for no `sdkInstance`.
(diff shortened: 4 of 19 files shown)


packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+interface IntrinsicButtonProps
+    extends React.HTMLAttributes<HTMLButtonElement>,
+        ButtonProps {}
+
+declare module "react" {
+    namespace JSX {
+        interface IntrinsicElements {
+            "paypal-button": IntrinsicButtonProps;
```
kand, line 14: This allows the linter to handle our custom component in JSX appropriately.

EvanReinstein: I think we'll need to re-use this declaration for every feature that leverages some web-component. For example `paypal-message` or `paypal-basic-card-container` are some non-button elements that we'll need to define in the future. With that in mind should we rename this file to something like `sdkWebComponents`? I'm open to other suggestions 💯

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -2,6 +2,10 @@ import { useRef } from "react";
 
 import type { Components } from "./types";
 
+/**
+ * @returns {boolean} `true` if code is running in a server environment where `window`
+ * or `document` are not available; `false` otherwise.
+ */
 export function isServer(): boolean {
     return typeof window === "undefined" && typeof document === "undefined";
 }
```
EvanReinstein, line 11: You're going to update this in a different PR, correct? I'm referring to the removal of the document check.

packages/react-paypal-js/src/v6/components/PayPalOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        autoRedirect?: never;
+    };
+
+/**
+ * `PayPalOneTimePaymentButton` is a button that provides a standard PayPal payment flow.
+ *
+ * `PayPalOneTimePaymentButtonProps` combines the arguments for {@link UsePayPalOneTimePaymentSessionProps}
+ * and {@link ButtonProps}.
+ *
+ * Note, `autoRedirect` is not allowed because if given a `presentationMode` of `"redirect"` the button
```
nbierdeman, line 20: Nice explanation. 👍

package-lock.json
EvanReinstein, line 1: Looks like some `package-lock.json` changes snuck through.

packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
kand: Ah right, that makes sense, I'll update the file name.

packages/react-paypal-js/src/v6/hooks/useError.ts
```diff
@@ -19,7 +19,7 @@ export function useError(
         (newError) => {
             setErrorInternal(newError);
 
-            if (!noConsoleErrors) {
+            if (!noConsoleErrors && newError) {
```
kand, line 22: I added this bit because `useError` was logging `null` in some cases.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -38,7 +38,16 @@ type PayPalProviderProps = CreateInstanceOptions<
 
 /**
  * {@link PayPalProvider} creates the SDK script, component scripts, runs eligibility, then
- * provides these in {@link PayPalContext} to child components via the {@link usePayPal} hook.
+ * provides these in context to child components via the {@link usePayPal} hook.
```
kand, line 43: I removed `PayPalContext` from the description since we're not going to export it any longer. I also added an example to get folks started with using the provider.

packages/react-paypal-js/src/v6/utils.ts
kand: Yea, I'll do that in my small fixes PR.

packages/react-paypal-js/src/v6/components/PayPalOneTimePaymentButton.test.tsx
```diff
@@ -0,0 +1,164 @@
...
+            <PayPalOneTimePaymentButton
+                onApprove={() => Promise.resolve()}
+                orderId="123"
+                presentationMode="auto"
+            />,
+        );
+        expect(container.querySelector("paypal-button")).toBeInTheDocument();
+    });
+
+    it("should return null when on server side", () => {
```
AleGastelum, line 41: Would it be worth splitting server/client side tests into two different files just like `PayPalProvider` does with test.tsx/ssr.test.tsx?

kand: If there end up being more SSR tests in the future I think that would make sense. For now, I don't think there's enough tests to split them out.

PR #747, opened 2025-12-01
Feat: v6 PayPal Messaging types

EvanReinstein:


packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+
+// ============================================================================
+// Public API Types
+// ============================================================================
+
+/**
+ * Session object for managing PayPal Messages.
+ */
+export type PayPalMessagesSession = {
+    fetchContent: (options?: UserFetchContentOptions) => MessageContent;
```
jeremy-herman: fetchContent returns a promise: `Promise<MessageContent | null>`
```suggestion
    fetchContent: (options?: UserFetchContentOptions) => Promise<MessageContent | null>;
```

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+ */
+type FetchContentOptions = PayPalMessagesOptions & {
+    amount?: string;
+    cache: Cache;
+    logoPosition: LogoPosition;
+    logoType: LogoType;
+    offerTypes?: OfferTypes;
+    onContentReady?: OnReady;
+    onTemplateReady?: OnReady;
+    requestType?: RequestType; // Used in lower environments to receive request from mock (client-tier) or samples (mid-tier)
```
jeremy-herman: We can remove `requestType` and the corresponding `RequestType` definition

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+ * User-facing options for fetching message content.
+ */
+type UserFetchContentOptions = Omit<
+    Partial<FetchContentOptions>,
+    "offerTypes" | "cache"
+> & {
+    cache?: UserCache;
+    offerTypes?: UserOfferTypes;
+    onReady?: OnReady;
+};
```
jeremy-herman: These "user input" types for the merchant will be interesting because there are some options that we don't want to make publicly known and keep them hidden. For example, the `cache` option is not something we document for merchant use. Options we expose: https://docs.paypal.ai/payments/methods/pay-later/get-started#fetch-content-options

Should we consider a type like this that acts as a suggestion interface that allows for added options if the merchant knows them. Notice I removed `offerTypes` and `cache`, but the addition of `[key: string]: unknown;` would mean the merchant can add any additional options and still satisfy the type.
```typescript
export type FetchContentOptions = PayPalMessagesOptions & {
  amount?: string;
  logoPosition: LogoPosition;
  logoType: LogoType;
  onContentReady?: OnReady;
  onReady?: OnReady;
  onTemplateReady?: OnReady;
  textColor?: TextColor;
  [key: string]: unknown;
};
```

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+
+/**
+ * User-facing options for Learn More presentations.
+ */
+export type UserLearnMoreOptions = Omit<
+    Partial<LearnMoreOptions>,
+    "clickUrl"
+> & {
+    clickUrl?: string | URL;
+};
```
jeremy-herman: Similar to the above comment for UserFetchContentOptions. Here's a suggested type for the LearnMore options based on our public docs: https://docs.paypal.ai/payments/methods/pay-later/get-started#learn-more-options

```typescript
export type LearnMoreOptions = PayPalMessagesOptions & {
  amount?: string;
  presentationMode: PresentationMode;
  onShow?: (data?: object) => void;
  onClose?: (data?: object) => void;
  onApply?: (data?: object) => void;
  onCalculate?: (data?: object) => void;
  [key: string]: unknown;
};
```

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+    | "PAYPAL_CASHBACK_MASTERCARD"
+    | "PAYPAL_CREDIT_NO_INTEREST"
+    | "PAYPAL_DEBIT_CARD"
+    | "PAY_LATER_LONG_TERM"
+    | "PAY_LATER_PAY_IN_1"
+    | "PAY_LATER_SHORT_TERM";
+
+type OfferTypes = OfferType[];
+
+export type UserOfferTypes = OfferTypes | string;
```
jeremy-herman: We should remove these. We probably won't reference offer types at all in the other types, and even if we did it would be type "string".
```suggestion
```

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+// ============================================================================
+// Caching Types
+// ============================================================================
+
+type Cache = {
+    local: boolean;
+    origin: boolean;
+};
+
+type UserCache = boolean | Partial<Cache>;
```
jeremy-herman: Similar to the offer types comment above
```suggestion
```

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+
+// ============================================================================
+// Fetch Content Types
+// ============================================================================
+
+type RequestType = "MOCK" | "SAMPLE";
```
jeremy-herman: Lets remove this
```suggestion
```

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,252 @@
...
+    embeddable?: boolean;
+    partnerAttributionId?: string;
+    buyerCountryCode?: string;
+    // values needed for analytics captured at time of fetch
+    renderDuration?: number;
+    pageType?: string;
+    attributes?: string[];
+    // Flag to indicate if content is template content or final content
+    isTemplateContent: boolean;
+};
```
jeremy-herman: I don't think we need to expose the explicit message content type and could probably replace uses of it with as generic an object as possible because a merchant should never be manipulating this object. It is only ever passed to the UI (`<paypal-message>`) from the Data Layer (`PayPalMessages`) and the merchant wouldn't ever be defining this themselves.

packages/paypal-js/types/v6/components/paypal-messages.d.ts
EvanReinstein: Will remove 👍

packages/paypal-js/types/v6/components/paypal-messages.d.ts
EvanReinstein: Sounds good, we have considered a pattern like this in other places but haven't implemented. this use case sounds good to me.

packages/paypal-js/types/v6/components/paypal-messages.d.ts
EvanReinstein: Sounds good, I'll update accordingly. 👍

packages/paypal-js/types/v6/components/paypal-messages.d.ts
EvanReinstein: Thank you for sharing the new docs 👍

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -0,0 +1,151 @@
...
+
+// ============================================================================
+// Public API Types
+// ============================================================================
+
+/**
+ * Session object for managing PayPal Messages.
+ */
+export type PayPalMessagesSession = {
+    fetchContent: (options?: FetchContentOptions) => MessageContent;
```
jeremy-herman: Fetch content would return a promise `Promise<MessageContent | null>`

EvanReinstein: Oh thanks for re-iterating this, forgot to update 👍

PR #749, opened 2025-12-01
Feat: Update Guest Payments Session presentation mode options

EvanReinstein:


packages/paypal-js/types/v6/components/paypal-guest-payments.d.ts
```diff
@@ -16,6 +20,13 @@ export type PayPalGuestOneTimePaymentSessionOptions =
 export type PayPalGuestPresentationModeOptions =
     PresentationModeOptionsForAuto & {
```
kand, line 21: Where we going to omit `presentationMode` here?

EvanReinstein: Oh, I thought we discussed using `Omit` in the BCDC hook itself and leaving this type unchanged. Something like [this](https://github.com/paypal/paypal-js/pull/743/files#r2578409044), since we want to remove `presentationMode` from the hook's arguments but still be able to correctly type the option we pass to `.start`.

EvanReinstein: Discussed with @kand and we are good on this implementation as is currently.

packages/paypal-js/types/v6/components/paypal-messages.d.ts
```diff
@@ -47,7 +47,7 @@ export type FetchContentOptions = PayPalMessagesOptions & {
 /**
  * Options for configuring Learn More presentations.
  */
-type LearnMoreOptions = PayPalMessagesOptions & {
+export type LearnMoreOptions = PayPalMessagesOptions & {
```
EvanReinstein, line 50: An update needed for the `paypal-message` hook.

PR #750, opened 2025-12-01
Feat: V6 Card Fields types

AleGastelum:
This PR adds types for V6 Card Fields

### DIFF
--- a/packages/paypal-js/types/v6/components/card-fields.d.ts
+++ b/packages/paypal-js/types/v6/components/card-fields.d.ts
@@ -96,7 +96,7 @@ export type EventPayload = {
 };
 
 export type CardFieldsEventsOptions = {
-    [key in MerchantMessagingEvents]: (state: EventPayload) => void;
+    [key in MerchantMessagingEvents]: (eventPayload: EventPayload) => void;
 };
 
 export type OrderAmount = {
@@ -162,8 +162,8 @@ type BaseCardFieldsSession = {
      * 
      * @example
      * ```typescript
-     * cardFieldsInstance.on("focus", (EventPayload) => {
-          console.log("Focus event triggered: ", EventPayload);
+     * cardFieldsInstance.on("focus", (eventPayload) => {
+          console.log("Focus event triggered: ", eventPayload);
         });
      * ```
      */


packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+
+export type StateType = "canceled" | "failed" | "succeeded";
+
+/**
+ * Card Fields Sessions
+ */
+
+type BaseCardFieldsSession = {
+    /**
+     * This method allows to create and configure individual Card Field components.
```
kand:
```suggestion
     * Use this method to create and configure individual Card Field components.
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+     * 
+     * @param config - Configuration options for creating individual Card Field components and customizing different base aspects such as type, placeholder, styling, and accessibility attributes.
+     * @returns An instance of the created Card Field component that can be appended to the DOM
+     * 
+     * @example
+     * ```typescript
+     * const numberField = cardFieldsInstance.createCardFieldsComponent({
+          type: "number",
+          placeholder: "Enter a number:",
```
kand: A bit of a nitpick, but the colon would be strange in `placeholder` text, right? I usually have seen colons used for label text.

```suggestion
          placeholder: "Enter a number",
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+        });
+        document
+          .querySelector("#paypal-card-fields-number-container")
+          .appendChild(numberField);
+     * ```
+     */
+    createCardFieldsComponent: (config: CardFieldOptions) => HTMLElement;
+    /**
+     * This method allows to register event listeners and set callbacks for them.
```
kand:
```suggestion
     * Use this method to register event listeners and set callbacks for them.
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+     * 
+     * @param eventName - Name of the event to listen for
```
kand:
```suggestion
     * @param {MerchantMessagingEvents} eventName - Name of the event to listen for
```

I'm not 100% sure this is the right markup, so double check, but it's nice to have the type in here cause it can be used to jump to the type in the intellisense popup.

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+          console.log("Focus event triggered: ", eventState);
+        });
+     * ```
+     */
+    on: (
+        eventName: MerchantMessagingEvents,
+        callback: CardFieldsEventsOptions[MerchantMessagingEvents],
+    ) => Promise<void>;
+    /**
+     * This method allows to update the Card Fields session with new options such as amount and cobranded eligibility.
```
kand:
```suggestion
     * Use this method to update the Card Fields session with new options such as amount and cobranded eligibility.
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+     * @param options - Configuration options to update the Card Fields session with new options
```
kand:
```suggestion
     * @param {UpdateOptions} options - Configuration options to update the Card Fields session with new options
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+        liabilityShift?: string;
+    };
+    state: StateType;
+};
+
+export type OneTimePaymentSubmitOptions = [orderId: string, data: ExtraFields];
+
+export type CardFieldsOneTimePaymentSession = BaseCardFieldsSession & {
+    /**
+     * This method allows to submit a one-time payment using Card Fields.
```
kand:
```suggestion
     * Use this method to submit a one-time payment using Card Fields.
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+};
+
+export type SavePaymentSubmitOptions = [
+    vaultSetupToken: string,
+    data: ExtraFields,
+];
+
+export type CardFieldsSavePaymentSession = BaseCardFieldsSession & {
+    /**
+     * This method allows to submit and save a payment method using Card Fields.
```
kand:
```suggestion
     * Use this method to submit and save a payment method using Card Fields.
```

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,300 @@
...
+     * ```
+     */
+    submit: (
+        ...args: SavePaymentSubmitOptions
+    ) => Promise<SavePaymentFlowResponse>;
+};
+
+/**
+ * Main Card Fields instance interface
+ */
```
kand: Is this comment necessary? I feel like this will be skimmed over / missed by future maintainers or users of this code.

AleGastelum: I honestly wasn't sure if it was worth adding it, so getting feedback on it is good. I can remove it, do you think it would be good to remove these ones too?
- [this](https://github.com/paypal/paypal-js/blob/89e81bf8ad4ed6399191b3b04e9a107090f7332f/packages/paypal-js/types/v6/components/card-fields.d.ts#L1-L3)
- [this](https://github.com/paypal/paypal-js/blob/89e81bf8ad4ed6399191b3b04e9a107090f7332f/packages/paypal-js/types/v6/components/card-fields.d.ts#L136-L138)

packages/paypal-js/types/v6/components/card-fields.d.ts
AleGastelum: Thanks, this is a nice improvement to the comments. Will check how to correctly markup this.

AleGastelum: Seems is not possible to add the type right after the `@param`. What do you think of something like this instead?

```suggestion
     * @param eventName - Name of the event to listen for. See {@link MerchantMessagingEvents} for a list of available events.
```

kand: Ah, you know what, this must only be for if you _don't_ have a type system in place:

https://jsdoc.app/tags-param

If I test this out in VSCode it seems to know the type of the `@param`s without having this in place.

I think you're good without adding the type to the doc string. Sorry for the confusion!

packages/paypal-js/types/v6/components/card-fields.d.ts
kand: Yea I think those would be good to remove as well. I assume most developers will end up in this file by clicking through to a specific type definition, and won't be scrolling around this file.

AleGastelum: I removed all those comments.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -75,6 +77,7 @@ export type CreateInstanceOptions<T extends readonly Components[]> = {
  * **Conditionally includes methods based on components:**
  * - `"paypal-payments"` - Adds PayPalPaymentsInstance methods
  * - `"venmo-payments"` - Adds VenmoPaymentsInstance methods
+ * - `"card-fields"` - Adds CardFieldsInstance methods
```
EvanReinstein, line 80: Thank you for your attention to detail here!

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -0,0 +1,294 @@
...
+     * ```typescript
+     * cardFieldsInstance.on("focus", (eventPayload) => {
+          console.log("Focus event triggered: ", eventPayload);
+        });
+     * ```
+     */
+    on: (
+        eventName: MerchantMessagingEvents,
+        callback: CardFieldsEventsOptions[MerchantMessagingEvents],
+    ) => Promise<void>;
```
EvanReinstein, line 173: Generally speaking I'm curious how these event listeners get cleaned up. Does the SDK take care of this?

mchoun (approved):
lgtm

packages/paypal-js/types/v6/components/card-fields.d.ts
AleGastelum: Yes, the Card Fields SDK is planned to take care of the listener clean up.

PR #751, opened 2025-12-03
Feat: Add v6 paypal-message component hook

EvanReinstein:
This PR adds a React hook to facilitate the basic `paypal-message` session creation and also provide access to 2 methods `fetchContent` and `createLearnMore`, which are used by merchants to enhance/customize the PayLater messaging.

### DIFF
--- a/.changeset/huge-lizards-dance.md
+++ b/.changeset/huge-lizards-dance.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
+---
+
+Updates loadCoreSdkScript to not log when called server-side.
--- a/.changeset/many-eyes-spend.md
+++ b/.changeset/many-eyes-spend.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+Adding PayLater UI Component
--- a/packages/paypal-js/src/v6/index.ts
+++ b/packages/paypal-js/src/v6/index.ts
@@ -10,11 +10,7 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
     validateArguments(options);
     const isServerEnv = isServer();
 
-    // SSR safeguard - warn about incorrect usage
     if (isServerEnv) {
-        console.warn(
-            "PayPal JS: loadCoreSdkScript() was called on the server. This function should only be called on the client side. Please ensure you're not calling this during server-side rendering.",
-        );
         return Promise.resolve(null);
     }
 
--- a/packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.test.tsx
@@ -1,0 +1,207 @@
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+
+import { PayLaterOneTimePaymentButton } from "./PayLaterOneTimePaymentButton";
+import { usePayLaterOneTimePaymentSession } from "../hooks/usePayLaterOneTimePaymentSession";
+import { usePayPal } from "../hooks/usePayPal";
+import { isServer } from "../utils";
+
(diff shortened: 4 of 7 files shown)


packages/react-paypal-js/src/v6/hooks/usePayPalMessages.ts
```diff
@@ -0,0 +1,108 @@
...
+    error: Error | null;
+    handleCreateLearnMore: (
+        options?: LearnMoreOptions,
+    ) => LearnMore | undefined;
+    handleFetchContent: (
+        options: FetchContentOptions,
+    ) => Promise<MessageContent | void>;
+};
+
+export function usePayPalMessages({
```
kand, line 41: Would it be helpful to add some documentation how to use this hook?

kand (approved):

packages/react-paypal-js/src/v6/hooks/usePayPalMessages.ts
```diff
@@ -0,0 +1,108 @@
...
+    useEffect(() => {
+        if (!sdkInstance) {
+            setError(new Error("no sdk instance available"));
+        }
+    }, [sdkInstance, setError]);
+
+    useEffect(() => {
+        if (!sdkInstance) {
+            return;
+        }
```
nityasp, line 62: I think we should check for the loadingStatus to avoid setting the error when SDK is still loading. Also to be consistent with error handling across the hooks.
```suggestion
     useEffect(() => {
            if (sdkInstance) {
                setError(null);
            } else if (loadingStatus !== INSTANCE_LOADING_STATE.PENDING) {
                setError(new Error("no sdk instance available"));
            }
        }, [sdkInstance, setError, loadingStatus]);

```

packages/react-paypal-js/src/v6/hooks/usePayPalMessages.ts
```diff
@@ -0,0 +1,108 @@
...
+
+        const newSession = sdkInstance.createPayPalMessages({
+            buyerCountry,
+            currencyCode,
+            shopperSessionId,
+        });
+
+        sessionRef.current = newSession;
```
nityasp: Maybe it is good to clear the sessionRef when component unmounts or when dependencies change to prevent any memory leaks. Something like this

```suggestion
        sessionRef.current = newSession;
        return () => {
            sessionRef.current = null;
        };
```

packages/react-paypal-js/src/v6/hooks/usePayPalMessages.ts
EvanReinstein: I added the useEffect above. I think that resolves this comment 👍

packages/react-paypal-js/src/v6/hooks/usePayPalMessages.ts
EvanReinstein: See below for the return with `sessionRef` clean up 👍

packages/react-paypal-js/src/v6/hooks/usePayPalMessages.ts
```diff
@@ -0,0 +1,123 @@
...
+ * });
+ */
+export function usePayPalMessages({
+    buyerCountry,
+    currencyCode,
+    shopperSessionId,
+}: PayPalMessagesOptions): PayPalMessagesReturn {
+    const { sdkInstance, loadingStatus } = usePayPal();
+    const isMountedRef = useIsMountedRef();
+    const [session, setSession] = useState<PayPalMessagesSession | null>(null); //useRef
```
nityasp: Are you considering to use `useRef` later? Does this component needs to re-render when the session changes and that why are we using `useState` instead of `useRef`?

EvanReinstein: Ah, thank you for calling this out. That comment was a reminder to the prior logic, and yes `useState` is preferred here so that the hook's `session` can be updated, otherwise the callbacks do not get the updated state.

I'm going to remove the comment.

nityasp (approved):
LGTM.

PR #752, opened 2025-12-03
Small updates for react

kand:


packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -43,7 +45,7 @@ type PayPalProviderProps = CreateInstanceOptions<
 export const PayPalProvider: React.FC<PayPalProviderProps> = ({
     clientMetadataId,
     clientToken,
-    components,
+    components = ["paypal-payments"],
```
kand, line 48: Most merchants will be using `paypal-payments` so it's a safe default.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -46,10 +47,12 @@ export function usePayLaterOneTimePaymentSession({
 
     // Separate error reporting effect to avoid infinite loops with proxyCallbacks
     useEffect(() => {
-        if (!sdkInstance) {
+        if (sdkInstance) {
```
kand, line 50: This change prevents an error when there's no `sdkInstance` because the instance is still being loaded. I put this change in all of the session hooks.

packages/react-paypal-js/src/v6/index.ts
```diff
@@ -1,6 +1,5 @@
 export * from "./types";
 export { PayPalProvider } from "./components/PayPalProvider";
-export { PayPalContext } from "./context/PayPalProviderContext";
```
kand, line 3: This export is removed since merchants won't need to use it directly.

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -3,7 +3,7 @@ import { useRef } from "react";
 import type { Components } from "./types";
 
 export function isServer(): boolean {
-    return typeof window === "undefined" && typeof document === "undefined";
+    return typeof window === "undefined";
```
kand, line 6: Checking `window` should be sufficient for SSR.

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
```diff
@@ -220,6 +220,30 @@ describe("PayPalProvider", () => {
...
+        test("should use default components array when not provided", async () => {
+            const mockCreateInstance = jest
+                .fn()
+                .mockResolvedValue(createMockSdkInstance());
+
+            (loadCoreSdkScript as jest.Mock).mockResolvedValue({
+                createInstance: mockCreateInstance,
+            });
+
+            // @ts-expect-error components is required, testing defaulting behavior
```
AleGastelum: nit: Could a similar omit like the one used [here](https://github.com/paypal/paypal-js/blob/a73fc6d99fbf0ae4f29f21682443f0abd7ce6f56/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx#L30-L35) be added [here](https://github.com/paypal/paypal-js/blob/a73fc6d99fbf0ae4f29f21682443f0abd7ce6f56/packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx#L54-L57) to prevent using the `@ts-expect-error` comment?

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
AleGastelum: Sweet change, will make sure to add this as well to the new CardFieldsProvider component that checks the `sdkInstance` instance.

packages/react-paypal-js/src/v6/utils.ts
AleGastelum: Out of curiosity, why is the document check not needed?

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -83,6 +83,70 @@ describe("useVenmoOneTimePaymentSession", () => {
             expect(error).toEqual(new Error("no sdk instance available"));
         });
 
+        test("should not error if there is no sdkInstance but loading is still pending", () => {
```
AleGastelum, line 89: Right before this test there is this other one `should not create session when no SDK instance is available`

I might be missing something, but they seem to be set up fairly similarly (both have `sdkInstance:null` and `loadingStatus` as `pending`), but one is expecting and error while the other is not. Is that correct? I see this commit checks passed, so not very sure where the difference is

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
```diff
@@ -83,6 +83,70 @@ describe("useVenmoOneTimePaymentSession", () => {
...
+            const {
+                result: {
+                    current: { error },
+                },
+            } = renderHook(() => useVenmoOneTimePaymentSession(props));
+
+            expect(error).toBeNull();
+        });
+
+        test.only("should clear any sdkInstance related errors if the sdkInstance becomes available", () => {
```
AleGastelum: Is the `only` here intentional?

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
```diff
@@ -48,10 +49,12 @@ export function usePayPalOneTimePaymentSession({
...
     // Separate error reporting effect to avoid infinite loops with proxyCallbacks
     useEffect(() => {
-        if (!sdkInstance) {
+        if (sdkInstance) {
+            setError(null);
+        } else if (loadingStatus !== INSTANCE_LOADING_STATE.PENDING) {
             setError(new Error("no sdk instance available"));
         }
-    }, [sdkInstance, setError]);
+    }, [sdkInstance, setError, loadingStatus]);
```
AleGastelum, line 57: How this conditions are set up would make the error state persist `new Error("no sdk instance available")` value until a new instance of `sdkInstance` is loaded successfuly. Is this desired behavior? or should the error state be handled by the `useEffect` clean up? something like this:

```
useEffect(() => {
    if (!sdkInstance && loadingStatus !== INSTANCE_LOADING_STATE.PENDING) {
    }
    
    return () => {
        setError(null);
    }
}, [sdkInstance, setError, loadingStatus])
```

packages/react-paypal-js/src/v6/utils.ts
kand: I can't find evidence of a server environment that would have a `document` without a `window`, so this check should be a superset of the check it replaced. The original issue I had with this check, that started a discussion around this, was the `&&` would cause `isServer` to return `false` if `window` was undefined and `document` was defined.

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
kand: I cleaned up the `renderProvider` type to:

1. Use the props from the actual component.
2. Not need these expect error comments.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
kand: Ah good catch. I differentiated these tests a bit more. Let me know if it makes more sense now.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
kand: Nope! 🤦  Good catch!

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
kand: If there's no `sdkInstance` and the `loadingStatus` is not `"pending"`, then there should be an `error` returned by the hook. That `error` should exist until there's an `sdkInstance`. So, I think the way I have the logic is equivalent to your suggested changes but IMO a bit clearer.

> How this conditions are set up would make the error state persist new Error("no sdk instance available") value until a new instance of sdkInstance is loaded successfuly.

Yes this is what we want, I believe; unless, is there a non-error state I'm missing when there's no `sdkInstance` and the `loadingStatus` is not `"pending"`?

AleGastelum: This is the case I was thinking when making my suggestion:
1. `sdkInstance` fails to load, loadingStatus becomes `REJECTED`, hook error is set.
2. `sdkInstance` attempts to load again, loadingStatus becomes `PENDING` again
 - Current implementation would persist error (and only remove it if `sdkInstance` resolves successfully)
 - My suggestion would set the error to null (and only set an error again in case of a second `REJECTED`)

Based on your response it seems current implementation achieves the desired experience.

packages/react-paypal-js/src/v6/utils.ts
AleGastelum: Got it, thank you for providing context on this

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.test.ts
AleGastelum: Checked them and look good!

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
kand: Ah right, yes, I think we want the `error` to persist in this case. Though maybe there's more discussion to be had here at some point.

PR #754, opened 2025-12-05
CardFieldsProvider and context created

AleGastelum:
Adds the following:

- Card Fields Provider
- Card Fields Context
- useCardFields hook for interacting with the provider's context
- Test files

The Card Fields Provider creates, manages, and shares session with child components that will need access to it.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/CardFieldsProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/CardFieldsProvider.tsx
@@ -40,9 +40,15 @@ type CardFieldsProviderProps = {
  * {@link CardFieldsProvider} creates the appropriate Card Fields session based on the `sessionType` prop value, and then provides it to child components that require it.
  *
  * @example
- * <CardFieldsProvider sessionType={"one-time-payment"}>
- *  <CheckoutForm />
- * </CardFieldsProvider>
+ * <PayPalProvider
+ *  components={["card-fields"]}
+ *  clientToken={clientToken}
+ *  pageType="checkout"
+ * >
+ *   <CardFieldsProvider sessionType={"one-time-payment"}>
+ *    <CheckoutForm />
+ *   </CardFieldsProvider>
+ * </PayPalProvider>
  */
 export const CardFieldsProvider = ({
     children,


packages/react-paypal-js/src/v6/components/CardFieldsProvider.tsx
```diff
@@ -0,0 +1,97 @@
...
+            return;
+        }
+
+        // Error case: Loading finished but no sdkInstance
+        if (!sdkInstance) {
+            const errorMsg = toError("no sdk instance available");
+            setError(errorMsg);
+            setCardFieldsError(errorMsg);
+            return;
+        }
```
EvanReinstein, line 83: I'm wondering if we want to handle the error case in a separate useEffect? The hooks follow [this](https://github.com/paypal/paypal-js/blob/1e87dd348134c5fff77c296c04c18462a9fedcfd/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts#L50-L57) pattern.

The early return AND the error handling together effectively do the same thing as the hook's strategy, this is merely a suggestion.

AleGastelum: I think it might be good to follow the pattern other hooks/components are following. Thanks for mentioning it, I will make the update

AleGastelum: Took another look at this—the separate useEffect pattern in [this](https://github.com/paypal/paypal-js/blob/1e87dd348134c5fff77c296c04c18462a9fedcfd/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts#L50-L57) file seems to be addressing an infinite looping issue we don't have here. I'm leaning towards keeping it in one useEffect for simplicity and maintainability, but wanted to get your thoughts. Is there another benefit to splitting it that I'm not seeing?

EvanReinstein: I think you're correct in the difference between this use case and the hooks'. The only benefit is having a separate effect that will handle errors unrelated to creating the session, but as you said it may impact simplicity and maintainability.

packages/react-paypal-js/src/v6/index.ts
```diff
@@ -2,6 +2,7 @@ export * from "./types";
 export { PayPalOneTimePaymentButton } from "./components/PayPalOneTimePaymentButton";
 export { PayPalProvider } from "./components/PayPalProvider";
 export { usePayPal } from "./hooks/usePayPal";
+export { CardFieldsProvider } from "./components/CardFieldsProvider";
```
nityasp: Any specific reason why `useCardFields` hook is not exported? Currently we are exporting all our hooks. Is there any different intended usage pattern that we want to access CardFields session?

AleGastelum: Good question. `useCardFields` hook will be internally used by other Card Field components and hooks (that will be added in following PRs) to manage session usage. Those components will then make sure to export anything the merchant needs from the session. That is why only the provider is needed by the merchant and not the hook.

packages/react-paypal-js/src/v6/hooks/useCardFields.ts
```diff
@@ -0,0 +1,46 @@
...
+import type {
+    CardFieldsSessionState,
+    CardFieldsStatusState,
+} from "../context/CardFieldsProviderContext";
+
+/**
+ * Returns {@link CardFieldsStatusContext} provided by a parent {@link CardFieldsProvider}
+ *
+ * @returns {CardFieldsStatusContext}
+ */
```
nityasp, line 18: Looks like the comment here should say that the function returns `CardFieldsStatusState` since we are returning the state not the context. 
```
/**
 * Returns {@link CardFieldsStatusState} provided by a parent {@link CardFieldsProvider}
 *
 * @returns {CardFieldsStatusState}
 */

File: packages/react-paypal-js/src/v6/hooks/useCardFields.ts [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    return context;
+}
+
+/**
+ * Returns {@link CardFieldsSessionContext} provided by a parent {@link CardFieldsProvider}
+ *
+ * @returns {CardFieldsSessionContext}
+ */
```
nityasp, line 35: Same the comment should say that the function returns `CardFieldsSessionState` 
```
/**
 * Returns {@link CardFieldsSessionState} provided by a parent {@link CardFieldsProvider}
 *
 * @returns {CardFieldsSessionState}
 */

File: packages/react-paypal-js/src/v6/hooks/useCardFields.ts (hunk 3 above)
Comment (AleGastelum) [reply packages/react-paypal-js/src/v6/hooks/useCardFields.ts:18]: Aaah good catch! Thank you for noticing it, will make the change

File: packages/react-paypal-js/src/v6/types/index.ts [hunk 5]
```diff
@@ -9,3 +9,5 @@ export interface BasePaymentSessionReturn {
     handleCancel: () => void;
     handleDestroy: () => void;
 }
+
+export type CardFieldsSessionType = "one-time-payment" | "save-payment";
```
EvanReinstein: I think this type is pretty specific to the card fields provider and should be located in that file 👍

packages/react-paypal-js/src/v6/components/CardFieldsProvider.tsx
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/**
+ * {@link CardFieldsProvider} creates the appropriate Card Fields session based on the `sessionType` prop value, and then provides it to child components that require it.
+ *
+ * @example
+ * <CardFieldsProvider sessionType={"one-time-payment"}>
+ *  <CheckoutForm />
+ * </CardFieldsProvider>
+ */
```
EvanReinstein, line 52: Would it be worth including the `<PayPalProvider/>` dependency in the example as well?

packages/react-paypal-js/src/v6/components/CardFieldsProvider.ssr.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    const html = renderToString(
+        <CardFieldsProvider sessionType={sessionType}>
+            <TestComponent>{children}</TestComponent>
+        </CardFieldsProvider>,
+    );
+
+    return { html, cardFieldsSessionState, cardFieldsStatusState };
+}
+
+function setupSSRTestComponent() {
```
kand, line 61: What makes this "SSR" specifically?

packages/react-paypal-js/src/v6/components/CardFieldsProvider.ssr.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    CardFieldsSessionState,
+    CardFieldsStatusState,
+} from "../context/CardFieldsProviderContext";
+import type { CardFieldsSessionType } from "../types";
+
+jest.mock("../hooks/usePayPal");
+
+jest.mock("../utils", () => ({
+    ...jest.requireActual("../utils"),
+    isServer: () => true,
```
kand, line 24: Also, I don't see the provider using `isServer` anywhere, should it be?

packages/react-paypal-js/src/v6/types/index.ts
AleGastelum: Moved the type to the card fields provider file!

packages/react-paypal-js/src/v6/components/CardFieldsProvider.ssr.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    });
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    describe("Server-Side Rendering", () => {
+        test("should verify isServer mock is working", () => {
+            expect(isServer()).toBe(true);
+        });
```
kand, line 104: This is a strange test, did this come from AI? I don't think this test suite should be checking this.

packages/react-paypal-js/src/v6/components/CardFieldsProvider.ssr.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+        test("should not attempt DOM access during server rendering", () => {
+            // In Node environment, document should be undefined
+            expect(typeof document).toBe("undefined");
```
kand, line 108: IMO, if we're meaning to test SSR, we should use and mock `isServer` for these tests rather than checking if the document is `undefined` directly.

packages/react-paypal-js/src/v6/components/CardFieldsProvider.tsx
AleGastelum: Updated the example to include `<PayPalProvider/>` in it.

packages/react-paypal-js/src/v6/components/CardFieldsProvider.ssr.test.tsx
kand, line 1: These tests are weird to me since we don't have SSR logic in the provider. Some of these tests seem incorrect, others are testing functionality I'm not sure we should be testing for SSR specifically (e.g. `Context isolation`), and some I don't think we're actually specifically coding for (e.g. `should maintain state consistency across multiple server renders`).

IMO we should add these when we have SSR logic, and we should be more specific / careful about what we're testing.

AleGastelum: Great callout! This tests were implemented following the tests `PayPalProvider` had for SSR. This subject was brought up with the team. The SSR tests files for [Card fields](https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/v6/components/CardFieldsProvider.ssr.test.tsx) and [PayPal](https://github.com/paypal/paypal-js/blob/main/packages/react-paypal-js/src/v6/components/PayPalProvider.ssr.test.tsx) providers will be removed in a new PR. SSR tests will be added as needed.

AleGastelum: Removed [here](https://github.com/paypal/paypal-js/pull/759)

PR #755, opened 2025-12-05
feat: Venmo One Time Payment button

EvanReinstein:
Adds the Venmo one time payment button. Also cleans up props in the Venmo session hook and some other small miscellaneous changes.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/useEligibleMethods.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/useEligibleMethods.test.ts
@@ -1,5 +1,6 @@
 import { renderHook } from "@testing-library/react-hooks";
 
+import { expectCurrentErrorValue } from "./useErrorTestUtil";
 import {
     useEligibleMethods,
     fetchEligibleMethods,
@@ -366,7 +367,7 @@ describe("useEligibleMethods", () => {
                     environment: "sandbox",
                 }),
             );
-
+            expectCurrentErrorValue(result.current.error);
             expect(result.current.isLoading).toBe(false);
             expect(result.current.error).toEqual(
                 new Error(


packages/react-paypal-js/src/v6/components/VenmoOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const { error, handleClick } = useVenmoOneTimePaymentSession(hookProps);
+    const isServerSide = isServer();
+
+    useEffect(() => {
+        if (error) {
+            console.error(error);
+        }
+    }, [error]);
+
+    return isServerSide ? null : (
```
AleGastelum: This might be a nit, but [this](https://github.com/paypal/paypal-js/blob/1e87dd348134c5fff77c296c04c18462a9fedcfd/packages/react-paypal-js/src/v6/components/PayPalOneTimePaymentButton.tsx#L47-L49) other button returns a `<div />` when on server. Is it fine for both buttons to return different things? or would it be worth it to standardize what button components return on server?

EvanReinstein: Oh! thank you for calling this out, I actually thought it returned `null` instead! I'll update this 👍

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.test.ts
```diff
@@ -366,7 +367,7 @@ describe("useEligibleMethods", () => {
                     environment: "sandbox",
                 }),
             );
-
+            expectCurrentErrorValue(result.current.error);
```
EvanReinstein, line 370: When running tests I was seeing 2 `console.error` logs that were appropriate considering that the test in question was confirming the expected error behavior. But they were false-flag noisey so I added this useful helper which silences those logs. This helper has been [implemented](https://github.com/paypal/paypal-js/pull/735/files) a few times for methods that use `useError`.

PR #757, opened 2025-12-08
Fix: react-paypal-js build warnings

EvanReinstein:
Running `npm run prepack` or `npm run build` in `react-paypal-js` completes successfully, however there are several warnings:
- spread array operator warning.
- output to filesystem is defaulting to true warning (see below comment for explanation).

The spread array operator warning stems from using v5 of `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser`. Currently, `paypal-js` uses v6 of both of these packages, so this change upgrades to match `paypal-js`. There were a few lint issues in V5 SDK-related files that have been fixed or ignored.

### DIFF
--- a/.changeset/curly-turkeys-camp.md
+++ b/.changeset/curly-turkeys-camp.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+Fixes a build warning log by upgrading typescript-eslint packages from v5 to v6.


package-lock.json
```diff
@@ -41336,250 +40947,6 @@
...
-                "@eslint-community/regexpp": "^4.4.0",
-                "@typescript-eslint/scope-manager": "5.62.0",
-                "@typescript-eslint/type-utils": "5.62.0",
-                "@typescript-eslint/utils": "5.62.0",
-                "debug": "^4.3.4",
-                "graphemer": "^1.4.0",
-                "ignore": "^5.2.0",
-                "natural-compare-lite": "^1.4.0",
-                "semver": "^7.3.7",
-                "tsutils": "^3.21.0"
```
EvanReinstein, line 41355: This dependency, `tsutils`, is the culprit behind the spread operator warning log. Basically this package depends on `tslib v1.4.1` while we wanted `tslib v2.6.2`. So `react-paypal-js` was using an older `tslib` version by proxy.

The updated `typescript-eslint` packages do not leverage `tsutils` and therefore do not have the older dependency.

packages/react-paypal-js/src/stories/payPalHostedFields/PayPalHostedFields.stories.tsx
```diff
@@ -243,7 +243,7 @@ export default {
     ],
 };
 
-export const Default: FC<StoryProps> = ({ amount, styles, style }) => {
+export const Default: FC<StoryProps> = ({ styles, style }) => {
```
EvanReinstein, line 246: Removed unused variable `amount`.

packages/react-paypal-js/src/types/scriptProviderTypes.ts
```diff
@@ -5,6 +5,7 @@ import type { ReactNode, Dispatch } from "react";
 import type { PayPalScriptOptions } from "@paypal/paypal-js";
 
 export interface ReactPayPalScriptOptions extends PayPalScriptOptions {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Allow any for dynamic SDK options
     [key: string]: any;
```
EvanReinstein, line 9: Considered changing this to `[key: string]: unknown` but then opted for the `eslint-disable-next-line` to avoid breaking any integrations.

packages/react-paypal-js/rollup.config.js
```diff
@@ -21,6 +21,7 @@ export default [
         plugins: [
             typescript({
                 tsconfig: "./tsconfig.lib.json",
+                outputToFilesystem: true,
```
EvanReinstein, line 24: The default value for `outputToFilesystem` is true and if you haven't explicitly set the value here it also triggers a warning. So technically we were already writing output to a `.tsbuildinfo` file in the `dist`. Updating this to false will remove that buildinfo file.

PR #760, opened 2025-12-11
PayLater UI component

nityasp:


packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { usePayLaterOneTimePaymentSession } from "../hooks/usePayLaterOneTimePaymentSession";
+import { isServer } from "../utils";
+
+import type { PayLaterButtonProps } from "../types";
+import type { PayLaterOneTimePaymentSessionProps } from "../hooks/usePayLaterOneTimePaymentSession";
+
+type PayLaterOneTimePaymentButtonProps = PayLaterOneTimePaymentSessionProps &
+    PayLaterButtonProps & {
+        autoRedirect?: never;
```
kand: What do you think about leaving a comment for why this is not available? If I saw this type as a merchant I might be confused why this isn't allowed for this specific button.

kand (approved):

packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
```diff
@@ -4,15 +4,26 @@ export interface ButtonProps {
     disabled?: boolean;
 }
 
+export interface PayLaterButtonProps {
+    countryCode?: string;
+    productCode?: string;
+    disabled?: boolean;
+}
```
EvanReinstein: I see in the SDK code that `productCode` and `countryCode` are optional attributes for the web component. But digging a little deeper I see that if they are null/undefined then the Button's label is not rendered. What happens when these values are not passed in here?

nityasp: I see that the button is completely hidden in the SDK and not clickable when `countryCd`/'productCd` is undefined.
https://github.paypal.com/PayPal-R/core-web-sdk/blob/1447ba4ae4a8c2af19f772e6b1b194b683762209/packages/web-sdk-ui/src/buttons/PayLaterButton/PayLaterButton.ts#L137

So in the React component, if the merchant doesn't pass these props, the button will render but will be invisible.

Do we want to make these props as required in the React component to keep the button visible and clickable when any of the props are null/undefined?

EvanReinstein: I don't think we want to deviate from the SDK's logic too drastically. I'm bringing this up mostly to remind us that there is a documentation gap re: PayLater integration

EvanReinstein (approved):
This LGTM. I am going to make a followup task based on some of the eligibility convos happening on slack.

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ...hookProps
+}: PayLaterOneTimePaymentButtonProps): JSX.Element | null => {
+    const { error, handleClick } = usePayLaterOneTimePaymentSession(hookProps);
+    const { eligiblePaymentMethods } = usePayPal();
+    const isServerSide = isServer();
+
+    const payLaterDetails =
+        eligiblePaymentMethods?.eligible_methods?.paypal_pay_later;
+    const countryCode = countryCodeProp ?? payLaterDetails?.country_code;
+    const productCode = productCodeProp ?? payLaterDetails?.product_code;
```
EvanReinstein: This is great, I think we can remove the `countryCode` and `productCode` props entirely though. The way the `PayPalProvider` is set up there should always be an `eligiblePaymentMethods` object (either server-side provided or queried via hook in the provider). So we should always use the eligibility response values, rather than allowing values to be passed as props.

nityasp: ah good to know. I thought we want to provide merchants option to manually pass the props.

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+    }, [error]);
+
+    return isServerSide ? (
+        <div />
+    ) : (
+        <paypal-pay-later-button
+            onClick={handleClick}
+            countryCode={countryCode}
+            productCode={productCode}
+            disabled={disabled || error !== null ? true : undefined}
```
jtanya17:
```suggestion
            disabled={disabled || !!error || undefined}
```

packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
EvanReinstein: Update: this was discussed with the team and for now we will implement an eligibility dependent solution, where the `productCode` and `countryCode` are directly sourced from the paypal context. 🚀

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.tsx
nityasp: Thanks this is much cleaner. I made the update now.

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+            orderId: "123",
+            presentationMode: "auto" as const,
+        };
+        render(<PayLaterOneTimePaymentButton {...hookProps} />);
+        expect(mockUsePayLaterOneTimePaymentSession).toHaveBeenCalledWith(
+            hookProps,
+        );
+    });
+
+    it("should log error to console when an error from the hook is present", () => {
```
kand: We don't need to test this since this functionality is internal to `usePayLaterOneTimePaymentSession`.

packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
```diff
@@ -4,15 +4,34 @@ export interface ButtonProps {
...
+    /**
+     * Required for the button to render. Automatically populated from the eligibility API response
+     * (`eligible_methods.paypal_pay_later.country_code`) by `PayLaterOneTimePaymentButton`.
+     */
+    countryCode?: string;
+    /**
+     * Required for the button to render. Automatically populated from the eligibility API response
+     * (`eligible_methods.paypal_pay_later.product_code`) by `PayLaterOneTimePaymentButton`.
+     */
+    productCode?: string;
```
AleGastelum: Is it necessary to include this two props in the interface? This comment explains that the code will take the values from the eligibility hook response, so it seems that the merchant providing the props or not makes no difference. Are the values of this optional props intended to be used somewhere? Otherwise it might be confusing for the merchant to have props that don't impact the component when used.

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.test.tsx
```diff
@@ -0,0 +1,227 @@
...
+                    presentationMode="auto"
+                />,
+            );
+
+            const button = container.querySelector("paypal-pay-later-button");
+            expect(button).not.toHaveAttribute("countrycode");
+            expect(button).not.toHaveAttribute("productcode");
+        });
+
+        it("should handle when PayLater is not in the eligibility response", () => {
```
kand, line 187: Do we need to also disable the button in this case?

jtanya17 (approved):

packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
nityasp: You are right. The button component no longer need these props. These props are required for the SDK's web component. I'll update the comment.

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.test.tsx
nityasp: I don't think we have to disable the button here because we are verifying that the props aren't set, which causes the SDK to hide the button. We don't need to explicitly test for disabled in this case.

packages/react-paypal-js/src/v6/types/sdkWebComponents.d.ts
AleGastelum: More than updating the comment, should those two props be removed from the interface? If the merchant won't use them seems counterintuitive to have them there.

nityasp: I agree to your point. It could mislead the merchant to pass those values which get overwritten later by the eligibility data. I removed those props now.

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.test.tsx
kand: Ah gotcha okay that makes sense!

packages/react-paypal-js/src/v6/components/PayLaterOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+            console.error(error);
+        }
+    }, [error]);
+
+    return isServerSide ? (
+        <div />
+    ) : (
+        <paypal-pay-later-button
+            onClick={handleClick}
+            countryCode={countryCode}
```
kand, line 58: There's a type issue with this line I think: https://github.com/paypal/paypal-js/actions/runs/20284227056/job/58253958459#step:9:43

PR #762, opened 2025-12-11
use card fields one time payment hook

AleGastelum:
This PR creates the card fields `usePayPalCardFieldsOneTimePaymentSession` hook.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
@@ -129,7 +129,7 @@ export const PayPalCardFieldsProvider = ({
 
     const statusContextValue: CardFieldsStatusState = useMemo(
         () => ({
-            cardFieldsError,
+            error: cardFieldsError,
         }),
         [cardFieldsError],
     );
--- a/packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.test.tsx
@@ -112,8 +112,8 @@ describe("PayPalCardFieldsProvider", () => {
             const { result } = renderCardFieldsProvider();
 
             expect(result.current.session.cardFieldsSession).toBeNull();
-            expect(result.current.status.cardFieldsError).toBeNull();
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expect(result.current.status.error).toBeNull();
+            expectCurrentErrorValue(result.current.status.error);
         });
 
         test("should error if there is no sdkInstance and loading is rejected", () => {
@@ -127,10 +127,10 @@ describe("PayPalCardFieldsProvider", () => {
             const { result } = renderCardFieldsProvider();
 
             expect(result.current.session.cardFieldsSession).toBeNull();
-            expect(result.current.status.cardFieldsError).toEqual(
+            expect(result.current.status.error).toEqual(
                 toError("no sdk instance available"),
             );
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expectCurrentErrorValue(result.current.status.error);
         });
 
         test("should clear any sdkInstance related errors if the sdkInstance becomes available", () => {
@@ -145,10 +145,10 @@ describe("PayPalCardFieldsProvider", () => {
             const { result, rerender } = renderCardFieldsProvider();
 
             expect(result.current.session.cardFieldsSession).toBeNull();
-            expect(result.current.status.cardFieldsError).toEqual(
+            expect(result.current.status.error).toEqual(
                 toError("no sdk instance available"),
             );
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expectCurrentErrorValue(result.current.status.error);
 
             // Second render: sdkInstance becomes available, error should clear
             mockUsePayPal.mockReturnValue({
@@ -161,8 +161,8 @@ describe("PayPalCardFieldsProvider", () => {
 
             rerender();
             expect(result.current.session.cardFieldsSession).toBeNull();
-            expect(result.current.status.cardFieldsError).toBeNull();
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expect(result.current.status.error).toBeNull();
+            expectCurrentErrorValue(result.current.status.error);
         });
     });
 
@@ -171,8 +171,8 @@ describe("PayPalCardFieldsProvider", () => {
             const { result } = renderCardFieldsProvider();
 
             expect(result.current.session.cardFieldsSession).toBeNull();
-            expect(result.current.status.cardFieldsError).toBeNull();
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expect(result.current.status.error).toBeNull();
+            expectCurrentErrorValue(result.current.status.error);
             expect(
                 mockSdkInstance.createCardFieldsOneTimePaymentSession,
             ).not.toHaveBeenCalled();
@@ -197,8 +197,8 @@ describe("PayPalCardFieldsProvider", () => {
             expect(result.current.session.cardFieldsSession).toBe(
                 mockCardFieldsOneTimePaymentSession,
             );
-            expect(result.current.status.cardFieldsError).toBeNull();
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expect(result.current.status.error).toBeNull();
+            expectCurrentErrorValue(result.current.status.error);
         });
 
         test("should create save-payment session when sessionType is set to save-payment", () => {
@@ -217,8 +217,8 @@ describe("PayPalCardFieldsProvider", () => {
             expect(result.current.session.cardFieldsSession).toBe(
                 mockCardFieldsSavePaymentSession,
             );
-            expect(result.current.status.cardFieldsError).toBeNull();
-            expectCurrentErrorValue(result.current.status.cardFieldsError);
+            expect(result.current.status.error).toBeNull();
+            expectCurrentErrorValue(result.current.status.error);
         });
 
         test("should handle errors when creating a session", () => {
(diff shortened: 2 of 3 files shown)


packages/react-paypal-js/src/v6/hooks/useCardFieldsOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,221 @@
...
+            cardFieldsSession: mockCardFieldsOneTimePaymentSession,
+            cardFieldsSessionType: CARD_FIELDS_SESSION_TYPES.ONE_TIME_PAYMENT,
+        });
+    });
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    test("should return the expected object structure", () => {
```
kand: I've seen AI generate these types of tests frequently and I don't find them useful; apologies, again, if this is copy/pasted from another suite 😅. This test is not checking the exact structure (e.g. some kind of equality), just that these names exist in the object, and not even that the values are equal to anything.

I think we should remove this test and ensure with other tests in the suite that (1) the property exists and (2) the value is correct.

packages/react-paypal-js/src/v6/hooks/useCardFieldsOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        expect(result.current).toHaveProperty("submitResponse");
+    });
+
+    describe("submit", () => {
+        test("should provide a submit function", () => {
+            const { result } = renderHook(() =>
+                useCardFieldsOneTimePaymentSession(),
+            );
+
+            expect(typeof result.current.submit).toBe("function");
```
kand: Similar to my above comment, this test is covered by other tests on `submit`.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,221 @@
...
+            await act(async () => {
+                await result.current.submit(mockOrderId, mockOptions);
+            });
+
+            expect(
+                mockCardFieldsOneTimePaymentSession.submit,
+            ).toHaveBeenCalledWith(mockOrderId, mockOptions);
+        });
+
+        test("should update submitResponse with the session's submit response", async () => {
```
kand, line 90: I feel like this test and the previous one could be rolled up into one test, for example: "`submit` should to be called with arguments X and return Y".

packages/react-paypal-js/src/v6/hooks/useCardFieldsOneTimePaymentSession.ts
```diff
@@ -0,0 +1,66 @@
...
+export function useCardFieldsOneTimePaymentSession(): useCardFieldsOneTimePaymentSessionReturn {
+    const { cardFieldsSession, cardFieldsSessionType } = useCardFieldsSession();
+    const [submitResponse, setSubmitResponse] =
+        useState<OneTimePaymentFlowResponse | null>(null);
+    const [error, setError] = useError();
+
+    const submit: SubmitCardFieldsOneTimePayment = useCallback(
+        async (orderId, options) => {
+            if (
+                cardFieldsSessionType !==
```
kand: Is there going to be a save payment session type as well? Will it be significantly different than the one time payment type? Is there ever a case where both session types could be under one provider?

IMO it's strange that this hook errors when given a wrong value in the provider. If there's a significant difference would it make sense to also have two different providers to perhaps make it more obvious to merchants that there's an architectural difference in the session type? I think when I see a string value in a provider, I wouldn't expect my application to throw an error if I change the value; rather, the code would respond correctly given that value.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.test.ts
```diff
@@ -0,0 +1,221 @@
...
+
+            const { result } = renderHook(() =>
+                useCardFieldsOneTimePaymentSession(),
+            );
+
+            act(() => {
+                result.current.submit(mockOrderId, mockOptions);
+            });
+
+            const expectedError = toError("CardFields session not available");
```
kand: I'm not sure what the best approach would be here, but I think (1) the provider should throw this error since it's responsible for the session and (2) we should do something else here because if (1) is true, then the merchant will get 2 errors for one issue.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.test.ts
AleGastelum: They reason they were separated is to have one test that verifies the hook's `submit` is correctly passing the expected arguments to the SDK's submit, and then have others that validate the hook's behavior when the SDK's responds (either with a success or an error/reject, having a test for each). Would still be worth combining the tests you proposed considering this?

packages/react-paypal-js/src/v6/hooks/useCardFieldsOneTimePaymentSession.ts
AleGastelum: Great callout! This was discussed with the team, and the architecture was updated so that the hook sets the `CardFieldsProvider`'s sessionType rather than having the provider expect that value as a string prop. This removes the potential for sessionType `mismatch` between provider and hook.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.test.ts
AleGastelum: Any session creation error is handled by the `PayPalCardFieldsProvider`. The provider will update its provided `error` object in case of session error.

The intent of the hook's submit method errors are to inform what caused the submit to error, in this case, the session not being available by the time `submit` was triggered. Would changing the error message to `Submit Error: CardFields session not available` make the error intention more clear?

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import { usePayPalCardFieldsSession } from "./usePayPalCardFields";
+import { useError } from "./useError";
+import { CARD_FIELDS_SESSION_TYPES } from "../components/PayPalCardFieldsProvider";
+import { toError } from "../utils";
+
+import type { ExtraFields, OneTimePaymentFlowResponse } from "../types";
+import type { PayPalCardFieldsProvider } from "../components/PayPalCardFieldsProvider";
+
+export type usePayPalCardFieldsOneTimePaymentSessionReturn = {
```
jtanya17: nit: please consider renaming this to `usePayPalCardFieldsOneTimePaymentSessionResult`

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.ts
```diff
@@ -0,0 +1,69 @@
...
+    submit: SubmitPayPalCardFieldsOneTimePayment;
+    submitResponse: OneTimePaymentFlowResponse | null;
+    error: Error | null;
+};
+
+type SubmitPayPalCardFieldsOneTimePayment = (
+    orderId: Promise<string> | string,
+    options: ExtraFields,
```
jtanya17: Should we make `options` an optional field here?

jtanya17: nit: can we rename `ExtraFields` to `SubmitOptions` to match the decision in the [ADR](https://github.paypal.com/PayPal-R/core-web-sdk/blob/develop/decisions/059-v6-react-card-fields.md)?

jtanya17 (commented):
Great work on this!

packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
```diff
@@ -24,39 +24,51 @@ import type {
...
  * </PayPalProvider>
  */
 export const PayPalCardFieldsProvider = ({
     children,
-    sessionType,
 }: CardFieldsProviderProps): JSX.Element => {
     const { sdkInstance, loadingStatus } = usePayPal();
     const [cardFieldsSession, setCardFieldsSession] =
         useState<CardFieldsSession | null>(null);
+    const [cardFieldsSessionType, setCardFieldsSessionType] =
```
jtanya17, line 70: Nice!

packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
```diff
@@ -85,10 +97,15 @@ export const PayPalCardFieldsProvider = ({
...
+        if (!cardFieldsSessionType) {
+            return;
+        }
+
         // Create Card Fields session based on sessionType
         try {
             const newCardFieldsSession =
-                sessionType === "one-time-payment"
+                cardFieldsSessionType ===
+                CARD_FIELDS_SESSION_TYPES.ONE_TIME_PAYMENT
```
EvanReinstein, line 108: I'm a little confused. If the `cardFieldsSessionType` starts as `null` won't this always default the session to save payment session?

EvanReinstein: ah ok, after reviewing a little more/further I see that the `setCardFieldsSessionType` is accessible via the provider and that the hooks leverage it.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.ts
AleGastelum: Done

AleGastelum: Good call, updated the type.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsOneTimePaymentSession.ts
AleGastelum: Updated!

PR #765, opened 2025-12-16
feat: Updated to use trusted publishing for npm packages.

bywood:
This updates the GitHub Release Action to use trusting publishing for npm packages. See https://docs.npmjs.com/trusted-publishers for details.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,10 @@ on:
         branches:
             - main
 
+permissions:
+    id-token: write # Required for OIDC
+    contents: read
+
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -24,14 +28,13 @@ jobs:
             - name: ⎔ Setup node
               uses: actions/setup-node@v3
               with:
-                  node-version: ${{ env.NVMRC }}
+                  # Hard-coding to 24 to get an npm version greater than 11.5.1 for trusted publishing.
+                  # See https://docs.npmjs.com/trusted-publishers.
+                  node-version: 24 # ${{ env.NVMRC }}
```
EvanReinstein, line 33: Marking this for follow up to update the nvmrc to node [v24](https://nodejs.org/en/about/previous-releases)

PR #766, opened 2025-12-16
Updated permissions for trusted publishing for npm packages.

bywood:
This updates the permissions for the GitHub Release Action, which is using trusted publishing, to allow it to create pull-requests. See https://docs.npmjs.com/trusted-publishers for details.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ on:
 
 permissions:
     id-token: write # Required for OIDC
-    contents: read
+    contents: write
     pull-requests: write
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -8,6 +8,7 @@ on:
 permissions:
     id-token: write # Required for OIDC
     contents: read
```
gregjopa: I think we need this to be able to push branches upstream
```suggestion
    contents: write
```

bywood: Updated.

gregjopa (approved):

PR #771, opened 2025-12-18
Card fields save payment hook

AleGastelum:
This PR creates the card fields `usePayPalCardFieldsSavePaymentSession` hook.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
@@ -73,7 +73,7 @@ describe("usePayPalCardFieldsSavePaymentSession", () => {
     });
 
     describe("submit", () => {
-        test("should call the session's submit method with correct arguments", async () => {
+        test("should update submitResponse with the session's submit response", async () => {
             const { result } = renderHook(() =>
                 usePayPalCardFieldsSavePaymentSession(),
             );
@@ -85,17 +85,6 @@ describe("usePayPalCardFieldsSavePaymentSession", () => {
             expect(
                 mockCardFieldsSavePaymentSession.submit,
             ).toHaveBeenCalledWith(mockVaultSetupToken, mockOptions);
-        });
-
-        test("should update submitResponse with the session's submit response", async () => {
-            const { result } = renderHook(() =>
-                usePayPalCardFieldsSavePaymentSession(),
-            );
-
-            await act(async () => {
-                await result.current.submit(mockVaultSetupToken, mockOptions);
-            });
-
             expect(result.current.submitResponse).toEqual(
                 mockSuccessfulSubmitResponse,
             );
@@ -121,6 +110,9 @@ describe("usePayPalCardFieldsSavePaymentSession", () => {
                 "Submit error: CardFields session not available",
             );
 
+            expect(
+                mockCardFieldsSavePaymentSession.submit,
+            ).not.toHaveBeenCalled();
             expect(result.current.submitResponse).toBeNull();
             expect(result.current.error).toEqual(expectedError);
             expectCurrentErrorValue(result.current.error);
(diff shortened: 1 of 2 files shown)


packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
```diff
@@ -47,7 +48,7 @@ type CardFieldsProviderProps = {
  *
  * @remarks
  * Child components must use either {@link usePayPalCardFieldsOneTimePaymentSession} or
- * usePayPalCardFieldsSavePaymentSession to initialize the appropriate session type.
+ * {@link usePayPalCardFieldsSavePaymentSession} to initialize the appropriate session type.
```
kand, line 51: I love it! It's great when docs are linked up like this.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
```diff
@@ -0,0 +1,201 @@
...
+            await act(async () => {
+                await result.current.submit(mockVaultSetupToken, mockOptions);
+            });
+
+            expect(
+                mockCardFieldsSavePaymentSession.submit,
+            ).toHaveBeenCalledWith(mockVaultSetupToken, mockOptions);
+        });
+
+        test("should update submitResponse with the session's submit response", async () => {
```
kand: IMO this test could be merged with the one above. It's essentially the same code path, just checking different things, right?

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
```diff
@@ -0,0 +1,201 @@
...
+                "Submit error: CardFields session not available",
+            );
+
+            expect(result.current.submitResponse).toBeNull();
+            expect(result.current.error).toEqual(expectedError);
+            expectCurrentErrorValue(result.current.error);
+        });
+
+        test("should set error when session's submit rejects", async () => {
+            const submitRejectError = new Error("Submit failed");
```
kand: A nitpick, but this mock error setup is slightly different than the one on line 120:

```
            const expectedError = toError(
                "Submit error: CardFields session not available",
            );
```

Could we use one way of setting up mock errors / checking for them?

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.ts
```diff
@@ -0,0 +1,69 @@
...
+                )) as SavePaymentFlowResponse;
+
+                setSubmitResponse(submitResult);
+                setError(null);
+            } catch (error) {
+                setError(toError(error));
+                setSubmitResponse(null);
+            }
+        },
+        [cardFieldsSession, setError],
```
nityasp, line 65: Shouldn't we include `setSubmitResponse` in the dependency array since `setError` is included?

AleGastelum: Good question. `setBumitResponse` is stable and will never change its reference after render. Only reactive values referenced inside the `useEffect` need to be added to the dependencies array, so it is safe to omit.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
AleGastelum: They reason they were separated is to have one test that verifies the hook's submit is correctly passing the expected arguments to the SDK's submit method, and then have other tests that validate the hook's behavior when the SDK's submit method responds (either with a success or an error/reject, having a test for each). Would still be worth combining the tests you proposed considering this?

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
AleGastelum: Good callout, I will update the test file to setup errors by using `toError` only, keeping it consistent.

packages/react-paypal-js/src/v6/hooks/usePayPalCardFieldsSavePaymentSession.test.ts
AleGastelum: Giving a second thought, I also think merging them works good, so I ended up making the change 🚀 , thanks for the observation.

srramirez26926-droid (approved):

srramirez26926-droid (approved):

PR #774, opened 2025-12-22
Feat: PayPal Credit OTP and Save Payment Hooks

EvanReinstein:
This PR adds 2 hooks and corresponding tests:
- `usePayPalCreditOneTimePaymentSession`
- `usePayPalCreditSavePaymentSession`

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.ts
@@ -55,7 +55,7 @@ export function usePayPalCreditSavePaymentSession({
 }: PayPalCreditSavePaymentSessionProps): BasePaymentSessionReturn {
     const { sdkInstance, loadingStatus } = usePayPal();
     const isMountedRef = useIsMountedRef();
-    const sessionRef = useRef<SavePaymentSession | null>(null); // handle cleanup
+    const sessionRef = useRef<SavePaymentSession | null>(null);
     const proxyCallbacks = useProxyProps(callbacks);
     const [error, setError] = useError();
 
--- a/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
@@ -36,7 +36,7 @@ export function usePayLaterOneTimePaymentSession({
 }: PayLaterOneTimePaymentSessionProps): BasePaymentSessionReturn {
     const { sdkInstance, loadingStatus } = usePayPal();
     const isMountedRef = useIsMountedRef();
-    const sessionRef = useRef<OneTimePaymentSession | null>(null); // handle cleanup
+    const sessionRef = useRef<OneTimePaymentSession | null>(null);
     const proxyCallbacks = useProxyProps(callbacks);
     const [error, setError] = useError();
 


packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
```diff
@@ -0,0 +1,140 @@
+import { useCallback, useEffect, useRef } from "react";
+
+import { usePayPal } from "./usePayPal";
+import { useIsMountedRef } from "./useIsMounted";
+import { useError } from "./useError";
+import { useProxyProps } from "../utils";
+import { INSTANCE_LOADING_STATE } from "../types/PayPalProviderEnums";
+import {
```
AleGastelum: Should we update this to be `import type`?

packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    const handleClick = useCallback(async () => {
+        if (!isMountedRef.current) {
+            return;
+        }
+
+        if (!sessionRef.current) {
+            setError(new Error("PayPal session not available"));
+            return;
+        }
```
AleGastelum, line 135: Why is it necessary to check that this `isMountedRef` variable has been set to `true`? Wouldn't the `!sessionRef.current` condition be enough for this purpose? or is there any value to having this separated ref boolean value that I'm not considering?

packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
```diff
@@ -0,0 +1,140 @@
...
+        sessionRef.current?.cancel();
+    }, []);
+
+    useEffect(() => {
+        if (sdkInstance) {
+            setError(null);
+        } else if (loadingStatus !== INSTANCE_LOADING_STATE.PENDING) {
+            setError(new Error("no sdk instance available"));
+        }
+    }, [sdkInstance, setError, loadingStatus]);
```
AleGastelum, line 79: This `sdkInstance` validation on a separate hook pattern here is added for the same reason explained in [this](https://github.com/paypal/paypal-js/blob/6453079e11e45f144fc55e999a0fa07cbb013a31/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts#L50-L57) other file correct? If that is the case, it would be worth adding the same comment so it is explicit why the separate hook approach is being used.

packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
EvanReinstein: Good catch 👍

packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
EvanReinstein: Good question! At a high level the first condition is a React lifecycle safety check, and the second is PayPal-specific validation. 

We can't depend solely on `!sessionRef.current` since there could be moments when that condition evaluates to true but the component has unmounted and the `setError` call will cause a console.error at a minimum. We catch those scenarios with `isMountedRef`.

packages/react-paypal-js/src/v6/hooks/usePayPalSavePaymentSession.ts
nityasp: This should be `vaultSetupToken`.
```suggestion
    | (Omit<SavePaymentSessionOptions, "vaultSetupToken"> & {
```

packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
```diff
@@ -0,0 +1,142 @@
...
+          orderId?: never;
+      })
+    | (PayPalCreditOneTimePaymentSessionOptions & {
+          createOrder?: never;
+          orderId: string;
+      })
+) &
+    PayPalPresentationModeOptions;
+
+export function usePayPalCreditOneTimePaymentSession({
```
nityasp, line 49: Can we add some documentation on how to use this hook? Looks like couple of hooks already have usage documentation added. It would be nice to have the usage documentation added for all hooks.

packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.ts
```diff
@@ -0,0 +1,139 @@
...
+          vaultSetupToken?: never;
+      })
+    | (SavePaymentSessionOptions & {
+          createVaultToken?: never;
+          vaultSetupToken: string;
+      })
+) &
+    PayPalPresentationModeOptions;
+
+export function usePayPalCreditSavePaymentSession({
```
nityasp, line 48: Same here. Documentation will be helpful.

packages/react-paypal-js/src/v6/hooks/usePayPalSavePaymentSession.ts
nityasp, line 47: Same here. Documentation will be helpful.

EvanReinstein: Good catch 👍

packages/react-paypal-js/src/v6/hooks/usePayPalCreditOneTimePaymentSession.ts
EvanReinstein: Thanks for the reminder 👍

EvanReinstein: Just pushed a commit adding a doc comment for the hooks in question.

PR #779, opened 2025-12-22
Feat: Update v6 rollup config

EvanReinstein:
This updates the V6 rollup config to copy `.d.ts` files into the `dist/`. Prior to this change `.d.ts` files were not included in the package build.


package-lock.json
EvanReinstein, line 1: This needs to be revisited before merge.

PR #781, opened 2025-12-23
Feature/client token check provider

nityasp:
This PR is for rendering the children of react PayPalProvider while fetching the clientToken.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
@@ -146,10 +146,13 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
                     token = await clientToken;
                     setClientTokenValue(token);
                 } catch (error) {
-                    setError(error);
+                    const clientTokenError = new Error(
+                        `Failed to resolve clientToken. Expected a Promise that resolves to a string, but it was rejected with: ${toError(error).message}`,
+                    );
+                    setError(clientTokenError);
                     dispatch({
                         type: INSTANCE_DISPATCH_ACTION.SET_ERROR,
-                        value: toError(error),
+                        value: clientTokenError,
                     });
                     return;
                 }
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
@@ -371,11 +371,20 @@ describe("PayPalProvider", () => {
             const tokenError = new Error("Token fetch failed");
             const tokenPromise = Promise.reject(tokenError);
 
+            tokenPromise.catch(() => {});
+
             const { state } = renderProvider({
                 clientToken: tokenPromise,
             });
 
-            await waitFor(() => expectRejectedState(state, tokenError));
+            await waitFor(() => {
+                expect(state.loadingStatus).toBe(
+                    INSTANCE_LOADING_STATE.REJECTED,
+                );
+                expect(state.error?.message).toBe(
+                    "Failed to resolve clientToken. Expected a Promise that resolves to a string, but it was rejected with: Token fetch failed",
+                );
+            });
             expect(loadCoreSdkScript).not.toHaveBeenCalled();
         });
 


packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -84,13 +109,33 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
     });
 
     useEffect(() => {
-        if (!isLoading && eligibleMethods) {
+        // Only set eligibility if we have a valid clientToken and methods
+        if (!isLoading && eligibleMethods && clientToken) {
```
EvanReinstein: Is there a scenario where the client token is undefined, but they have passed a value for `FindEligiblePaymentMethodsResponse` to the provider so technically `eligibleMethods` could exist and there shouldn't be anything preventing it being dispatched.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -157,13 +157,11 @@ export function useEligibleMethods({
...
+        // This allows the provider to render while the token is being fetched
         if (!clientToken) {
-            setError(
-                new Error(
-                    "clientToken is required when eligibleMethodsResponse is not provided",
-                ),
-            );
-            setIsLoading(false);
+            setIsLoading(true);
+            setError(null);
```
EvanReinstein: I'm wondering about this change. The early return in the `memoizedEligibleMethodsResponse` condition above handles setting the `eligibleMethods` value when merchants pass the API response from the server. 

Can you elaborate on why `setError` is called here?

EvanReinstein: I should clarify, that we definitely want to handle the optional `clientToken` gracefully which I think this does, I'm just not clear on the `setError`

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
nityasp: Actually I don't need clientToken checked here again, since the eligibleMethods (useEligibility hook) is already checking for the clientToken. I'll remove the clientToken check here.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
nityasp: Thanks for pointing this out. Since we are just waiting for clientToken here, we don't need to clear errors but just return early. I'll remove this `setError`

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -92,6 +117,25 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
         }
     }, [isLoading, eligibleMethods]);
 
+    // Warn if clientToken is not available within a 15s time
+    useEffect(() => {
+        if (clientToken) {
+            return;
+        }
+
+        const timeoutId = setTimeout(() => {
```
gregjopa: What do you think about matching the existing react-paypal-js interface with `deferLoading`: https://github.com/paypal/paypal-js/tree/main/packages/react-paypal-js#deferloading

That way the developer can control when the provider is ready to depend on the clientToken instead of having a setTimeout for this.
```suggestion
```

EvanReinstein (approved):
I tested this out in my local V6 React sample integration and was able to remove the clientToken ternary logic shown below and did not blow up the app 👍 
```
return clientToken ? (return provider here) : null
```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -92,33 +126,55 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+                setClientTokenValue(token);
+            } else {
+                try {
+                    token = await clientToken;
+                    setClientTokenValue(token);
+                } catch (error) {
+                    setError(error);
+                    dispatch({
+                        type: INSTANCE_DISPATCH_ACTION.SET_ERROR,
+                        value: toError(error),
```
kand: What types of errors could happen here? I'm wondering if we should have this error message just tell the merchant that a `Promise` was expected?

nityasp: There could be fetch errors/Network error. Adding a specific error message would be helpful to the merchants. I adde the error message now. Could you review if this wording is good @kand?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -92,33 +126,58 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+                token = clientToken;
+                setClientTokenValue(token);
+            } else {
+                try {
+                    token = await clientToken;
+                    setClientTokenValue(token);
+                } catch (error) {
+                    const clientTokenError = new Error(
+                        `Failed to resolve clientToken. Expected a Promise that resolves to a string, but it was rejected with: ${toError(error).message}`,
+                    );
```
nityasp, line 151: @kand Is this error message good?

kand: Looks good, thanks!

PR #783, opened 2025-12-29
[Feature] Implement `on` and `update` methods on CardFieldsProvider

jtanya17:
https://paypal.atlassian.net/browse/DTPPCPSDK-4153

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
@@ -94,9 +94,6 @@ export const PayPalCardFieldsProvider = ({
     const proxyEventHandlers = useProxyProps(
         eventHandlers as Record<PropertyKey, unknown>,
     );
-    // Deep memoize event handlers to avoid unnecessary re-registration
-    const memoizedProxyEventHandlers =
-        useDeepCompareMemoize(proxyEventHandlers);
 
     // Memoize amount to avoid unnecessary updates when value hasn't changed
     const memoizedAmount = useDeepCompareMemoize(amount);
@@ -169,11 +166,9 @@ export const PayPalCardFieldsProvider = ({
             by iterating over the keys of proxyEventHandlers directly
             */
             (
-                Object.keys(
-                    memoizedProxyEventHandlers,
-                ) as MerchantMessagingEvents[]
+                Object.keys(proxyEventHandlers) as MerchantMessagingEvents[]
             ).forEach((eventName) => {
-                const handler = memoizedProxyEventHandlers[eventName];
+                const handler = proxyEventHandlers[eventName];
                 if (handler && typeof handler === "function") {
                     cardFieldsSession.on(
                         eventName,
@@ -184,7 +179,7 @@ export const PayPalCardFieldsProvider = ({
         } catch (error) {
             handleError(toError(`Failed to register event handlers: ${error}`));
         }
-    }, [cardFieldsSession, memoizedProxyEventHandlers, handleError]);
+    }, [cardFieldsSession, proxyEventHandlers, handleError]);
 
     // Update session configuration when props change
     useEffect(() => {


kand (approved):

packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
```diff
@@ -57,24 +64,36 @@ type CardFieldsProviderProps = {
  *  clientToken={clientToken}
  *  pageType="checkout"
  * >
- *   <PayPalCardFieldsProvider>
- *    <CheckoutForm />
+ *   <PayPalCardFieldsProvider
```
nityasp, line 68: The JSDoc example indicated that we can have an inline object literal for amount prop. This creates a new object reference (because you are tracking amount in the update effect's dependency array line 204) on every parent component re-render even when the amount value haven't changed. I think this will make unnecessary SDK update calls on every parent re-render. Can we memorize the amount object using useMemo to prevent unnecessary updates and use a stable reference in JSDoc example?

AleGastelum: Good call. This [useDeepCompareMemoized](https://github.com/paypal/paypal-js/blob/ab8d07e370d58ed5d28945d35bcd52c0caddc8ac/packages/react-paypal-js/src/v6/utils.ts#L146-L160) utility method might be helpful for this use case (although I worry deep comparisons will be done on every render with this). [useEligibleMethods](https://github.com/paypal/paypal-js/blob/ab8d07e370d58ed5d28945d35bcd52c0caddc8ac/packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts#L125-L126) hook uses it for a similar purpose.

nityasp: oh yes `useDeepCompareMemorize` is a better option here since we are already using it in the eligibility hook. Thanks for pointing to it.

jtanya17: Thanks for the great suggestion. I made the changes!

packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.test.tsx
```diff
@@ -343,4 +353,488 @@ describe("PayPalCardFieldsProvider", () => {
...
+            // The error message will include "Error: " prefix when Error is converted to string
+            expect(result.current.status.error).toEqual(
+                toError(
+                    `Failed to register event handlers: Error: ${errorMessage}`,
+                ),
+            );
+            expectCurrentErrorValue(result.current.status.error);
+        });
+
+        test("should call update when amount object reference changes (even with same values)", () => {
```
nityasp: This test name should be changed "should NOT call update when amount reference changes...." and the testcase needs to be updated since we are using `useDeepCompareMemorize` now.

packages/react-paypal-js/src/v6/components/PayPalCardFieldsProvider.tsx
```diff
@@ -52,29 +58,49 @@ type CardFieldsProviderProps = {
...
-    // The actual error instance is stored in the provider's state.
     const [, setError] = useError();
 
+    // Use proxy props for event handlers to avoid re-renders
+    const proxyEventHandlers = useProxyProps(
+        eventHandlers as Record<PropertyKey, unknown>,
+    );
+    // Deep memoize event handlers to avoid unnecessary re-registration
+    const memoizedProxyEventHandlers =
+        useDeepCompareMemoize(proxyEventHandlers);
```
nityasp: Do we need to apply deep memorize on the proxyEvent.? `useProxyProps` already provides a stable reference via `useRef`. 
https://github.com/paypal/paypal-js/blob/ab8d07e370d58ed5d28945d35bcd52c0caddc8ac/packages/react-paypal-js/src/v6/utils.ts#L114

Adding Deep memorize on the proxy will make it redundant i believe. Is there any specific usecase that require proxy to be memorized?

jtanya17: You are right, `useProxyProps` already provides a stable reference and always reflects the latest values, so wrapping it in `useDeepCompareMemoize` is redundant. Deep memoization is only needed for plain objects/arrays (like `amount`), not for proxies. Removed!

PR #784, opened 2026-01-05
Refactor v6 SDK Web Component declaration

EvanReinstein:
This PR moves the web component type declaration to the root `index.ts` for v6. This should apply the web component type declaration automatically so devs integrating with `react-paypal-js` should be able to use web components like `paypal-button` out of the box without type errors.
 
 In order to test this do not use `npm link` as this does not effectively persist the type declarations. Instead use `npm pack` which creates a `.tgz` tarball of `react-paypal-js` and mimics how NPM would serve it up. Instructions:
 
1. Pull the changes from this branch locally.
2. Run `npm run prepack && npm run build` from `packages/react-paypal-js/`
3. Run `npm pack` to create the `.tgz` file
4. In your test app run `npm install /path/to/the/.tgz/file/you/just/made`

### DIFF
--- a/packages/react-paypal-js/src/v6/index.ts
+++ b/packages/react-paypal-js/src/v6/index.ts
@@ -33,7 +33,7 @@ export { usePayPalMessages } from "./hooks/usePayPalMessages";
 export { usePayPalCreditOneTimePaymentSession } from "./hooks/usePayPalCreditOneTimePaymentSession";
 export { usePayPalCreditSavePaymentSession } from "./hooks/usePayPalCreditSavePaymentSession";
 
-// React 19+ JSX types augmentation
+// React 19+ JSX SDK Web Components type declaration
 declare global {
     // eslint-disable-next-line @typescript-eslint/no-namespace
     namespace JSX {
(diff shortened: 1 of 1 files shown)


kand (approved):
Code is all good, but I get this error when I try to install this in https://github.com/paypal-examples/v6-web-sdk-sample-integration/tree/main/client/prebuiltPages/react/oneTimePayment:

```
➜  oneTimePayment git:(main) ✗ npm install <path-to-tgz>
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: react@0.0.0
npm error Found: react@undefined
npm error node_modules/react
npm error   react@"19.1.0" from the root project
npm error
npm error Could not resolve dependency:
npm error peer react@"^16.8.0 || ^17 || ^18 || ^19" from @paypal/react-paypal-js@8.9.2
npm error node_modules/@paypal/react-paypal-js
npm error   @paypal/react-paypal-js@"file:<path-to-tgz>" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

Which is strange since I think `19.1.0` should match `^19` 🤔

PR #786, opened 2026-01-06
Update v6 jsx namespace with the available list of web components

EvanReinstein:


packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -1,21 +1,73 @@
...
+export interface InternalButtonProps {
     type?: "buynow" | "checkout" | "donate" | "pay" | "subscribe";
     disabled?: boolean;
 }
 
-export interface IntrinsicButtonProps
+export interface ButtonProps
     extends HTMLAttributes<HTMLButtonElement>,
-        ButtonProps {}
+        InternalButtonProps {}
```
EvanReinstein, line 10: This wording was a little verbose so I updated to this. I also considered `PayPalButtonProps` instead of `InternalButtonProps`.

packages/react-paypal-js/src/v6/index.ts
nityasp, line 40: Since the code is repeated twice in the two declares, we can create a reusable type like below

```suggestion
type SDKWebComponents = {
    "paypal-button": ButtonProps;
    "venmo-button": ButtonProps;
    "paypal-pay-later-button": PayLaterButtonProps;
    "paypal-credit-button": PayPalCreditButtonProps;
    "paypal-basic-card-button": PayPalBasicCardButtonProps;
    "paypal-basic-card-container": React.HTMLAttributes<HTMLElement>;
    "paypal-message": PayPalMessagesElement;
};

declare global {
    namespace JSX {
        interface IntrinsicElements extends SDKWebComponents {}
    }
}

declare module "react" {
    namespace JSX {
        interface IntrinsicElements extends SDKWebComponents {}
    }
}
```

EvanReinstein: I ended up moving this back into sdkWebComponents.ts and removing the unused imports from the top of this index file 👍

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -1,21 +1,83 @@
...
+export interface PayLaterButtonProps extends HTMLAttributes<HTMLButtonElement> {
     countryCode?: string;
     productCode?: string;
     disabled?: boolean;
 }
+
+export interface PayPalBasicCardButtonProps
+    extends HTMLAttributes<HTMLButtonElement> {
+    buyerCountry: string;
+    ref?: React.Ref<HTMLButtonElement>;
```
nityasp: We may not need to explicitly declare `ref` here since it is already inherited `HTMLAttributes`

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -1,21 +1,83 @@
...
+export interface PayPalMessagesElement extends HTMLAttributes<HTMLElement> {
+    amount?: string;
+    "auto-bootstrap"?: boolean;
+    "buyer-country"?: string;
+    "currency-code"?: string;
+    "logo-position"?: "INLINE" | "LEFT" | "RIGHT" | "TOP";
+    "logo-type"?: "MONOGRAM" | "WORDMARK";
+    "offer-types"?: string;
+    "presentation-mode"?: "AUTO" | "MODAL" | "POPUP" | "REDIRECT";
+    ref?: React.Ref<HTMLElement>;
```
nityasp: Same here. We may not need to explicitly declare `ref` here since it is already inherited `HTMLAttributes`

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -1,21 +1,83 @@
...
+    setContent?: (content: Record<string, unknown>) => void;
+}
+
+export type SDKWebComponents = {
+    "paypal-button": ButtonProps;
+    "venmo-button": ButtonProps;
+    "paypal-pay-later-button": PayLaterButtonProps;
+    "paypal-credit-button": PayPalCreditButtonProps;
+    "paypal-basic-card-button": PayPalBasicCardButtonProps;
+    "paypal-basic-card-container": React.HTMLAttributes<HTMLElement>;
```
AleGastelum: Will the type for this component be added later? or is it intended to be set this way?

EvanReinstein: This is intentional 👍

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
EvanReinstein: Let me take a look at this and the `ref` property that you called out above.

EvanReinstein: When I remove the `ref` property I end up getting an error with `paypal-message` and `paypal-basic-card-button` 🤔

nityasp: Huddled regarding this issue and confirmed that we need the `ref` declared in types.

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -1,21 +1,83 @@
...
+export interface PayLaterButtonProps extends HTMLAttributes<HTMLButtonElement> {
     countryCode?: string;
     productCode?: string;
     disabled?: boolean;
 }
+
+export interface PayPalBasicCardButtonProps
+    extends HTMLAttributes<HTMLButtonElement> {
+    buyerCountry: string;
+    ref?: Ref<HTMLElement>;
```
EvanReinstein, line 25: This is typed as an `Ref<HTMLElement>` b/c the BCDC ref is typed as such [here](https://github.com/paypal/paypal-js/blob/0212046bd05796e71bced76f54ceab5034f595a1/packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts#L64) and [here](https://github.com/paypal/paypal-js/blob/0212046bd05796e71bced76f54ceab5034f595a1/packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts#L18) which I guess we could update to `HTMLButtonElement` at a later time.

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
EvanReinstein: It turns out with React we do need to specifically declare a `ref` property. I've updated the 2 instances where this occurs.

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
nityasp: Huddled regarding this and confirmed that we need the `ref` declared in the type.

PR #787, opened 2026-01-08
Feature/adding isPending to hooks an buttons to allow deferred clientToken

nityasp:
https://paypal.atlassian.net/browse/DTPPCPSDK-4722

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalOneTimePaymentButton.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalOneTimePaymentButton.tsx
@@ -45,17 +45,15 @@ export const PayPalOneTimePaymentButton = ({
         }
     }, [error]);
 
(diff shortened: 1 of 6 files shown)


packages/react-paypal-js/src/v6/components/PayPalOneTimePaymentButton.tsx
```diff
@@ -44,6 +45,10 @@ export const PayPalOneTimePaymentButton = ({
         }
     }, [error]);
 
+    if (isPending) {
+        return null;
+    }
```
EvanReinstein: Returning `null` here means that there may be a brief moment with no UI. Would it be better to disable the button? At least that way the web component is present, even if it is not functional, since the UI depends on the Core script but the functionality is dependent on the Component scripts (which are client token dependent).

Has this point been discussed already?

PR #794, opened 2026-01-12
Feat: Types for hydrateEligibleMethods

EvanReinstein:


packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
```diff
@@ -45,6 +45,24 @@ export type FindEligibleMethodsGetDetails<T extends FundingSource> =
...
+};
+
+export type FindEligiblePaymentMethodsResponse = {
+    eligible_methods: {
+        [key in EligiblePaymentMethods]?: EligiblePaymentMethodDetails;
+    };
+    supplementary_data?: {
+        buyer_country_code?: string;
+    };
+};
```
EvanReinstein, line 64: The SDK type includes some configuration for ApplePay and GooglePay. Will need to update here and create 2 new components in order to get complete parity. Do we want to do this now, or when we add these features to the React package.

My gut tells me now. I will follow up on this.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -179,6 +180,9 @@ export interface BaseInstance {
      * ```
      */
     updateLocale: (locale: string) => void;
+    hydrateEligibleMethods: (
```
AleGastelum, line 206: I think it would be worth adding a descriptive comment with an example for `hydrateEligibleMethods` method similar to what `updateLocale` and `findEligibleMethods` have.

EvanReinstein: Thank you for the reminder!

gregjopa (approved):

PR #797, opened 2026-01-14
Feature: Update v6 Provider Eligibility Strategy

EvanReinstein:
This PR updates the `PayPalProvider` and `useEligibleMethods` based on changes in the V6 JS SDK to facilitate hydrating eligibility.

The `PayPalProvider` no longer needs to depend on `useEligibleMethods` to either hydrate or query for eligible payment methods. Now the SDK can handle this use case, making much of the complex logic within `useEligibleMethods` redundant and not needed. Now the hook can serve as a convenient abstraction of `usePayPal` and the relevant Context associated with eligibility.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
@@ -642,6 +642,26 @@ describe("PayPalProvider", () => {
             );
             expect(mockHydrateEligibleMethods).toHaveBeenCalledTimes(2);
         });
+
+        test("should handle error when findEligibleMethods fails", async () => {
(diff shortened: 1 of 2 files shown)


.changeset/config.json
```diff
@@ -7,5 +7,5 @@
     "access": "restricted",
     "baseBranch": "main",
     "updateInternalDependencies": "patch",
-    "ignore": ["@paypal/react-paypal-js"]
```
EvanReinstein, line 10: Removing this b/c it isn't helping us and I think that it broke the Github Release Action

nityasp (approved):
eligibility hook looks much cleaner now. Great job!!

PR #803, opened 2026-01-23
Fix: V6 Static Component Hydration Mismatch

EvanReinstein:
**Overview**                                                                                                                                    
                                                                                                                                              
  This PR fixes hydration mismatch errors in v6 static components by replacing the `isServer()` utility check with a centralized `isHydrated` state managed by `PayPalProvider`

  **Problem**                                                                                                                                     
                                                                                                                                              
  Previously, v6 button components used `isServer()` to detect server-side rendering and render a placeholder <div /> on the server. However,   this caused hydration mismatches because:                                                                                                   
                                                                                                                                              
  1. On the server: `isServer()` returns true → renders `<div />`                                                                                
  2. On initial client render: `isServer()` returns false → renders `<paypal-button>`                                                             
  3. React detects the mismatch between server HTML (`<div />`) and client VDOM (`<paypal-button>`)                                               
                                                                                                                                              
  **Solution**                                                                                                                                    
                                                                                                                                              
  Introduced a new `isHydrated` state pattern:                                                                                                  
                                                                                                                                              
  1. New hook: `useIsomorphicLayoutEffect` - Uses `useLayoutEffect` on client (runs synchronously before paint) and `useEffect` on server (avoids SSR warnings)                                                                                                                               
  2. `PayPalProvider` changes:
    - Added `isHydrated` state, initially false
    - Uses `useIsomorphicLayoutEffect` to set `isHydrated = true` after initial client render
    - Exposes `isHydrated` through context
  3. Button component changes:
    - Replace `isServer()` check with `isHydrated` from `usePayPal()` context
    - Render `<div />`when `!isHydrated`, render web component when `isHydrated`
    - This ensures server and initial client render both produce `<div />`, then switches to web component

 **Test Updates**                                                                                                                                 
                                                                                                                                              
 All test files updated to:                                                                                                                   
 1. Mock `isHydrated: true` in default mock context                                                                                             
 2. Remove `isServer` mock                                                                                                                      
 3. Update test names from "server side" to "hydrated"

### DIFF
--- a/packages/react-paypal-js/src/v6/utils.ts
+++ b/packages/react-paypal-js/src/v6/utils.ts
@@ -3,13 +3,6 @@ import { useRef } from "react";
 import type { Components } from "./types";
 
 /**
- * @returns {boolean} `true` if code is running in a server environment where `window` is not available; `false` otherwise.
- */
-export function isServer(): boolean {
-    return typeof window === "undefined";
-}
-
-/**
  * Performs a shallow equality check on two arrays.
  *
  * This function compares two arrays element-by-element using strict equality (===).


AleGastelum (approved):
Great work on this, it looks awesome! Just one question, non-blocker at all, It seems like `isServer` from react package [utils.ts](https://github.com/paypal/paypal-js/blob/ff32a5e4cd0f72670880a9694b4ab2910880cf3c/packages/react-paypal-js/src/v6/utils.ts#L8-L10) file is no longer in use after the changes in this PR. Do wee foresee any future usage of it? Otherwise it might be worth removing it so we don't have an unused method around.

PR #807, opened 2026-01-30
Feature: V6 Client Component Strategy

EvanReinstein:
## Summary                                                                                                                                  

  Adds `"use client"` directives to the v6 bundle for Next.js App Router / React Server Components compatibility.                             

  ## Problem

  When importing `PayPalProvider` from `@paypal/react-paypal-js/sdk-v6` in a Next.js application, I encountered:                 

  TypeError: createContext only works in Client Components.                                                                                   
  Add the "use client" directive at the top of the file to use it.                                                                            

  The v6 bundle uses React Context and hooks (`createContext`, `useContext`, `useState`, etc.), which are client-only APIs in React Server Components. Without the `"use client"` directive in the library, consumers must add their own wrapper file with the directive—a poor developer experience.                                                                                                                       

  ## Solution                                                                                                                                 

  Add per-file `"use client"` directives to client-only source files and configure rollup to preserve them in the output using          `preserveModules`.                                                                                                                          

  ### Key Changes                                                                                                                             

  1. **rollup.config.js**: Updated V6 ESM build to use `preserveModules: true`, which outputs individual files per source module instead of a single bundle. Added `rollup-preserve-directives` plugin to maintain directives through the build.                                          
  2. **package.json**: Updated `./sdk-v6` export to point to `./dist/v6/esm/index.js` (the new entry point).                                  
  3. **Source files**: Added `"use client";` to ~20 client-only files (components, hooks, context providers).                                 
  4. **Server-safe utility preserved**: `useFetchEligibleMethods` intentionally does NOT have the directive, allowing it to be used in Server Components or Route Handlers.                                                                                                               

  ## Options Considered                                                                                                                       

  | Approach | Description | Trade-offs |                                                                                                     
  |----------|-------------|------------|                                                                                                     
  | **Bundle-level banner** | Add `"use client"` as a rollup banner to the entire bundle | Simplest, but marks everything as client-only including server-safe utilities |                                                                                                           
  | **Per-file with single bundle** | Add directives to source files, use plugin to preserve | Plugin hoists ALL directives to bundle level when outputting single file—same effect as banner |                                                                                         
  | **Per-file with preserveModules** ✓ | Add directives to source files, output separate files | True granular control; server-safe utilities remain usable on server |                                                                                                                  
  | **Separate `/sdk-v6/client` entry** | Create new export path for client components | Confusing API; doesn't solve the original import error |                                                                                                                                     

  We chose **per-file with preserveModules** because:
  - Allows for more flexibility in the future WRT client and server components living side by side
  - Preserves `useFetchEligibleMethods` as a server-safe utility
  - Follows patterns used by libraries like Headless UI
  - Improves tree-shaking for selective imports
  - Standard imports work identically (not a breaking change)

  ## Output Structure

  dist/v6/esm/
  ├── index.js                              (no directive - just re-exports)
  ├── components/
  │   ├── PayPalProvider.js                 ('use client')
  │   └── ...
  ├── hooks/
  │   ├── usePayPal.js                      ('use client')
  │   ├── useFetchEligibleMethods.js        (NO directive - server-safe)
  │   └── ...
  └── context/
      └── PayPalProviderContext.js          ('use client')

  ## Consumer Impact                                                                                                                          

  **Not a breaking change.** Standard imports work identically:                                                                               

  ```jsx                                                                                                                                      
  // This import works the same before and after                                                                                              
  import { PayPalProvider, usePayPal } from "@paypal/react-paypal-js/sdk-v6";                                                                 
  ```

  Users no longer need to create their own "use client" wrapper files.                                                                        

  Test Plan                                                                                                                                   

  - All 489 existing tests pass                                                                                                               
  - Verified client components have 'use client'; at top of output files                                                                      
  - Verified useFetchEligibleMethods does NOT have directive (server-safe)                                                                    
  - Verified @paypal/paypal-js is externalized (not bundled)                                                                                  
  - Test with Next.js App Router POC to confirm no createContext error

### DIFF
--- a/.changeset/curvy-snakes-prove.md
+++ b/.changeset/curvy-snakes-prove.md
@@ -2,4 +2,4 @@
 "@paypal/react-paypal-js": minor
 ---
 
-Adds use client directives to v6 components and hooks.
+Updates the V6 rollup config to apply bundle-level use client directive to client bundle.


packages/react-paypal-js/package.json
```diff
@@ -26,7 +26,11 @@
         },
         "./sdk-v6": {
             "types": "./dist/types/v6/index.d.ts",
-            "import": "./dist/v6/esm/react-paypal-js.js"
+            "import": "./dist/v6/esm/index.js"
+        },
+        "./sdk-v6/server": {
+            "types": "./dist/types/v6/server.d.ts",
+            "import": "./dist/v6/esm/server.js"
```
EvanReinstein, line 33: This new entry point is to create a clear separation for any server method `useFetchEligibleMethods` that we might want to provide.

packages/react-paypal-js/src/v6/context/PayPalProviderContext.tsx
```diff
@@ -1,3 +1,5 @@
+"use client";
```
HackTheW2d: I know `PayPalDispatchContext.js` has been merged into main. And I think since `PayPalProviderContext` has consumed the dispatch context so it will be fine now. If we could add 'use client' to:
```
PayPalDispatchContext.js            
usePayPalDispatch.js      
utils.js    
```
it would be more robust in the future. Thanks! :)

PR #809, opened 2026-02-02
Feature/react v6 storybook update

HackTheW2d:
Create storybook for v6 React package
**Canvas View:**
Screenshot 2026-02-04 at 11 36 35 AM

**Doc View:**
Screenshot 2026-02-04 at 11 36 59 AM
Screenshot 2026-02-04 at 11 37 13 AM

Didn't implement stack pattern like v5 since v6 offer different presentation modes. Developers can select combinations in controllers.

### DIFF
--- a/packages/react-paypal-js-storybook/v6/package.json
+++ b/packages/react-paypal-js-storybook/v6/package.json
@@ -3,9 +3,9 @@
     "version": "0.1.0",
     "private": true,
     "description": "Storybook for PayPal React SDK V6 components",
-    "_comments": {
-        "NODE_PATH": "TODO: Remove NODE_PATH=./node_modules once we upgrade v5 SDK's storybook to v10. Currently needed because storybook CLI is hoisted to root node_modules/, but @storybook/react-vite is installed locally due to version conflict with storybook v6."
-    },
+    "//": [
+        "TODO: Remove NODE_PATH=./node_modules once we upgrade v5 SDK's storybook to v10. Currently needed because storybook CLI is hoisted to root node_modules/, but @storybook/react-vite is installed locally due to version conflict with storybook v6."
+    ],
     "scripts": {
         "storybook": "NODE_PATH=./node_modules storybook dev -p 6006",
         "build-storybook": "NODE_PATH=./node_modules storybook build -o ../../../docs/web-sdk-v6-react-storybook"


EvanReinstein (commented):
This is a tidy implementation 👍 I know that we discussed updating the deployed storybook's domain name. Did you want to include that change in a different PR?

packages/react-paypal-js-storybook/v6/src/shared/utils.ts
```diff
@@ -0,0 +1,113 @@
...
+export const oneTimePaymentCallbacks = {
+    onApprove: async (data: OnApproveDataOneTimePayments) => {
+        console.log("Payment approved:", data);
+        await captureOrder(data.orderId);
+        console.log("Order captured successfully");
+    },
+    onCancel: (data: OnCancelDataOneTimePayments) => {
+        console.log("Payment cancelled:", data);
+    },
+    onError: (error: unknown) => {
```
HackTheW2d, line 65: @EvanReinstein Here I used unknown because in [this line](https://github.com/paypal/paypal-js/blob/b4833c26f34853b44d59ce086e3849c0c859c830/packages/react-paypal-js/src/v6/types/sdkWebComponents.ts#L9) there's `ButtonProps extends HTMLAttributes<HTMLButtonElement>` and this `HTMLAttributes<HTMLButtonElement>` has its own onError which is conflicted with onErrorData. So if I put `onErrorData` type here there will be error. Do you think we want to fix button props like this:
```
export interface ButtonProps extends Omit<HTMLAttributes<HTMLButtonElement>, 'onError'>,     
  InternalButtonProps {                                                                        
  } 
```
to allow using onErrorData? Thanks! :)

packages/react-paypal-js-storybook/v6/src/stories/buttons/PayPalGuestPaymentButton.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    oneTimePaymentCallbacks,
+    disabledArgType,
+} from "../../shared/utils";
+import { V6DocPageStructure } from "../../components";
+import { getPayPalGuestPaymentButtonCode } from "../../shared/code";
+
+const GuestPaymentButtonWithContainer = (
+    props: React.ComponentProps<typeof PayPalGuestPaymentButton>,
+) => (
+    <paypal-basic-card-container>
```
HackTheW2d, line 16: @EvanReinstein Here merchants must use this container wrapper.
May I know do we have any future plan to include this wrapper inside `PayPalGuestPaymentButton` component? Thanks! :)

package-lock.json
EvanReinstein, line 1: Looks like some internal registry URLs are being used. We can huddle on this to update

packages/react-paypal-js-storybook/v6/package.json
```diff
@@ -0,0 +1,36 @@
+{
+    "name": "@paypal/react-paypal-js-storybook-v6",
+    "version": "0.1.0",
+    "private": true,
+    "description": "Storybook for PayPal React SDK V6 components",
+    "scripts": {
+        "storybook": "NODE_PATH=./node_modules storybook dev -p 6006",
```
HackTheW2d, line 10: here we must include this `NODE_PATH=./node_modules` because this is a sub package, and storybook v10 include a storybook CLI which is hoisted to the root package's `node_modules/`. And when running `npm run storybook:v6`, it will need `@storybook/react-vite` which is installed in the v6 storybook package's `node_modules/` due to version conflict with storybook v6. Therefore, in order to run this command, we need to set this environmental variable. In the future if we upgrade v5 sdk's storybook version to v10, then we can remove these.

packages/react-paypal-js-storybook/v6/src/shared/utils.ts
EvanReinstein: Ah interesting find, yea I think we want to update this.

package-lock.json
HackTheW2d: Happy to huddle on this! :)

packages/react-paypal-js-storybook/v6/src/stories/buttons/PayPalGuestPaymentButton.stories.tsx
EvanReinstein: Great question, @nityasp what do you think? I'm not sure if there was a specific reason to not wrap the `paypal-basic-card-container`.

package-lock.json
HackTheW2d: Fixed :)

packages/paypal-js/tsconfig.json
```diff
@@ -7,6 +7,7 @@
         "moduleResolution": "bundler",
         "target": "es5",
         "strict": true,
+        "skipLibCheck": true,
```
EvanReinstein, line 14: What was causing this? Is it b/c Storybook v10 is on a different typescript version?

.changeset/long-snakes-wear.md
EvanReinstein, line 1: I wonder if this changeset should include a patch change for `paypal-js` b/c of the tsconfig.json update.

packages/paypal-js/tsconfig.json
HackTheW2d: The error is here: https://github.com/paypal/paypal-js/actions/runs/21690207720/job/62548052692
                                                                              
  ├── @paypal/paypal-js                                                                        
  │   └── vitest@1.6.1          ← our test framework                                   
  │       └── @vitest/expect@1.6.1  ←   original version                
  │                                                                                            
  └── storybook@10.2.6 (extraneous)  ← new library                      
      └── @vitest/expect@3.2.4       ← new version as dependency                             
          └── @types/chai@5.2.2      ← introduce conflict here
So there are chai type conflict :(

gregjopa (approved):
Awesome work @HackTheW2d! 💯

.changeset/long-snakes-wear.md
HackTheW2d: Resolved!

PR #811, opened 2026-02-04
Feature: Update v6 PayPalProvider Eligibility

EvanReinstein:
This PR removes the default `sdkInstance.findEligibleMethods` request in the `PayPalProvider` and opts to provide a client-side hook `useEligibleMethods` that will:
- Use `sdkInstance.findEligibleMethods` to query for payment method eligibility and then add response to the PayPal Context.
- Return `eligiblePaymentMethods` directly from Context rather than making the SDK request.

This PR also adds an internal Dispatch context that can be used to update the PayPal context.

### DIFF
--- a/packages/react-paypal-js/src/v6/index.ts
+++ b/packages/react-paypal-js/src/v6/index.ts
@@ -37,6 +37,7 @@ export { VenmoOneTimePaymentButton } from "./components/VenmoOneTimePaymentButto
 // Core hooks
 export { usePayPal } from "./hooks/usePayPal";
 export * from "./hooks/useEligibleMethods";
+export * from "./hooks/useFetchEligibleMethods";
 export { usePayPalMessages } from "./hooks/usePayPalMessages";
 
 // Card fields hooks


packages/react-paypal-js/src/v6/hooks/useFetchEligibleMethods.ts
```diff
@@ -0,0 +1,127 @@
...
+ * // Next.js server component
+ * const response = await fetchEligibleMethods({
+ *     clientToken: token,
+ *     environment: "sandbox",
+ *     payload: { purchase_units: [{ amount: { currency_code: "USD" } }] }
+ * });
+ *
+ * <PayPalProvider eligibleMethodsResponse={response} ... />
+ */
+export async function useFetchEligibleMethods(
```
HackTheW2d, line 91: This seems like a regular fetch function to me. May I ask why we want to call this "use"? I may have sth I didn't consider before.

EvanReinstein: I think the common convention for a hook is that they start with `use`, see here https://react.dev/learn/reusing-logic-with-custom-hooks#hook-names-always-start-with-use

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -1,120 +1,144 @@
...
-                    "Accept-Language": "en-US,en;q=0.9",
-                },
-                body: JSON.stringify(defaultPayload),
-                signal,
-            },
-        );
+    // todo
+    // add doc comments
+    // open PR
+    // make PR description
```
nityasp: I think we can cleanup these comments.

nityasp (approved):
Left a minor comment. Other than that looks good to me.

packages/react-paypal-js/src/v6/context/PayPalDispatchContext.tsx
```diff
@@ -0,0 +1,13 @@
...
+import type { InstanceAction } from "./PayPalProviderContext";
+
+/**
+ * Internal context for dispatching PayPal instance state updates.
+ * This is NOT exported to external consumers.
+ * Only internal hooks like useFetchEligibleMethods can access this.
+ *
+ * @internal
+ */
+export const PayPalDispatchContext =
```
HackTheW2d, line 12: just wonder any reason why we need a context for dispatch only. Can we make it inside PayPalContext instead? Just a curious thought, i might overlook something :)

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -1,120 +1,137 @@
...
-};
+    useEffect(() => {
+        // Only fetch if:
+        // 1. sdkInstance is available
+        // 2. eligiblePaymentMethods not in context
+        // 3. Haven't already fetched for THIS sdkInstance
+        if (!sdkInstance) {
+            return;
+        }
+        if (eligiblePaymentMethods) {
```
HackTheW2d: This is great auto-save and auto-fetch mechanism! Just curious if a merchant changed from
`useEligibleMethods({ payload: { currency: "USD" } }); ` to
`useEligibleMethods({ payload: { currency: "EUR" } });`
Would this trigger sdk to re-render? Or it's ignoring the requests here? :)

HackTheW2d (approved):
Great work! 🚀 Just curious about a few things

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
EvanReinstein: Good catch. We definitely want to re-fetch if the payload has changed, even if there is already state in context.

packages/react-paypal-js/src/v6/context/PayPalDispatchContext.tsx
EvanReinstein: Ah I had this great response typed up but then the Github wonkiness removed it even though I clicked comment.

The TLDR is that we needed `dispatch` to be available to update Context from outside of the `PayPalProvider`, using `PayPalDispatchContext` allows for that AND for us to keep it internal.

PR #814, opened 2026-02-09
Feature/v6 react storybook enhancement

HackTheW2d:
Create Actions for storybooks to view payment flow:

Screenshot 2026-02-09 at 3 55 01 PM

### DIFF
--- a/packages/react-paypal-js-storybook/v6/src/decorators/PayPalProviderDecorator.tsx
+++ b/packages/react-paypal-js-storybook/v6/src/decorators/PayPalProviderDecorator.tsx
@@ -76,12 +76,8 @@ function SdkStatusMonitor({ children }: { children: React.ReactNode }) {
     }, [loadingStatus]);
 
     const handleClick = (e: React.MouseEvent) => {
-        const target = e.target as HTMLElement;
-        if (
-            target.closest("[data-paypal-button]") ||
-            target.closest("paypal-button") ||
-            target.tagName.toLowerCase().includes("paypal")
-        ) {
+        const tag = (e.target as HTMLElement).tagName.toLowerCase();
+        if (tag.endsWith("-button")) {
             action("button")(
                 "Click event dispatched from the PayPal payment button",
             );


packages/react-paypal-js-storybook/v6/src/decorators/PayPalProviderDecorator.tsx
```diff
@@ -60,6 +65,37 @@ function ErrorDisplay({ message }: { message: string }) {
...
 
+// This is for outputing in actions for logging purpose.
+function SdkStatusMonitor({ children }: { children: React.ReactNode }) {
+    const { loadingStatus } = usePayPal();
+    const hasLoggedInit = useRef(false);
+
+    useEffect(() => {
+        if (
+            loadingStatus === INSTANCE_LOADING_STATE.RESOLVED &&
+            !hasLoggedInit.current
```
AleGastelum: What is the purpose of `hasLoggedInit` ref boolean? Is it to ensure that the message `Library initialized and rendered` is only logged once in case the instance somehow re initializes?

HackTheW2d: Great catch! I've updated to remove hasLoggedInit. :) I believe every time we clicked switch it will re-load sdk. So that will output anyway. Therefore this is unnecessary

packages/react-paypal-js-storybook/v6/.storybook/preview.ts
```diff
@@ -3,6 +3,10 @@ import { withPayPalProvider } from "../src/decorators";
 
 const preview: Preview = {
     parameters: {
+        actions: {
+            // Automatically create actions for props matching on* pattern
+            argTypesRegex: "^on[A-Z].*",
+        },
```
AleGastelum, line 9: Nice automation

packages/react-paypal-js-storybook/v6/src/decorators/PayPalProviderDecorator.tsx
```diff
@@ -60,6 +65,32 @@ function ErrorDisplay({ message }: { message: string }) {
...
+    useEffect(() => {
+        if (loadingStatus === INSTANCE_LOADING_STATE.RESOLVED) {
+            action("SDK")("Library initialized and rendered");
+        }
+    }, [loadingStatus]);
+
+    const handleClick = (e: React.MouseEvent) => {
+        const target = e.target as HTMLElement;
+        if (
+            target.closest("[data-paypal-button]") ||
```
AleGastelum: Not very familiar with this. Do all PayPal buttons contain this `[data-paypal-button]` selector?

HackTheW2d: good catch! I've updated handleClick selectors. This should be more neat implementation now.

AleGastelum (approved):
Great work on this! LGTM

PR #815, opened 2026-02-10
Updating PayPalProvider with clientId enhancements

nityasp:
Added support for `clientId` as an authentication credential option in `PayPalProvider`, providing flexibility to authenticate using either `clientToken` or `clientId`.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
@@ -52,6 +52,10 @@ type PayPalProviderProps =
  * SDK loading is automatically deferred until clientToken or clientId is available.
  * Both can be either a string, Promise, or undefined.
  *
+ * ** Important: When passing a Promise, you must ensure referential stability across renders.
+ * An unstable Promise reference (e.g., calling `fetchClientToken()` or `fetchClientId()` inline)
+ * will cause the SDK to re-initialize on every render. Wrap the promise in `useMemo` or store it in state.
+ *
  * @example
  * // With string clientToken
  * <PayPalProvider
@@ -267,6 +271,7 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
                     partnerAttributionId,
                     shopperSessionId,
                     testBuyerCountry,
+                    merchantId: memoizedMerchantId,
                 };
 
                 const instanceOptions = isClientToken


.changeset/solid-memes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
```
EvanReinstein: Would you mind updating this to minor?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -134,34 +183,37 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-            if (typeof clientToken === "string") {
-                token = clientToken;
-                setClientTokenValue(token);
+            if (typeof authCredential === "string") {
+                if (isClientToken) {
+                    setClientTokenValue(authCredential);
+                } else {
+                    setClientIdValue(authCredential);
+                }
             } else {
```
EvanReinstein: Would the `else` branch be taking if `authCredential` was undefined?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -134,34 +183,37 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
                 try {
-                    token = await clientToken;
-                    setClientTokenValue(token);
+                    const resolved = await authCredential;
+                    if (isClientToken) {
+                        setClientTokenValue(resolved);
+                    } else {
+                        setClientIdValue(resolved);
+                    }
```
EvanReinstein: I'm not sure if there is a use-case for a merchant to be using `clientToken` and then switch to use `clientId`, or vice-vera. Would you mind following up on that?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
nityasp: Yes it does include `undefined`in this `else`. However I've added an early return now incase `authCredential` is `undefined`.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
nityasp: I don't think they can switch between the auth methods. Merchants have to unmount and remount the entire `PayPalProvider` with different props. That creates a new component instance where the states of `clientTokenValue` and `clientIdValue` would be reset. The discrimination union type prevents the switching of auth   methods.

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
```diff
@@ -322,26 +336,11 @@ describe("PayPalProvider", () => {
...
+                resolveToken!(TEST_CLIENT_TOKEN);
             });
 
             await waitFor(() => expect(loadCoreSdkScript).toHaveBeenCalled());
             await waitFor(() => expectResolvedState(state));
-            expect(mockCreateInstance).toHaveBeenCalledWith(
-                expect.objectContaining({
-                    clientToken: TEST_CLIENT_TOKEN,
-                }),
-            );
```
EvanReinstein, line 346: Don't we still want this test to confirm that client token was used in createInstance?

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
```diff
@@ -711,6 +710,97 @@ describe("PayPalProvider", () => {
...
+    describe("ClientId Support", () => {
+        const TEST_CLIENT_ID = "test-client-id";
+
+        test("should load SDK with clientId instead of clientToken", async () => {
+            const { state } = await renderProvider({
+                clientId: TEST_CLIENT_ID,
+            });
+
+            await waitFor(() => expectResolvedState(state));
+            expect(loadCoreSdkScript).toHaveBeenCalled();
```
EvanReinstein: I think this test should really be focused on whether `createInstance` is called with client ID rather than `loadCoreSdkScript`

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
nityasp: Great call. Yes we should be confirm it. The test cases are updated now.

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
```diff
@@ -711,6 +718,116 @@ describe("PayPalProvider", () => {
...
+                    INSTANCE_LOADING_STATE.REJECTED,
+                );
+                expect(state.error?.message).toBe(
+                    "Failed to resolve clientId. Expected a Promise that resolves to a string, but it was rejected with: ClientId fetch failed",
+                );
+            });
+            expect(loadCoreSdkScript).not.toHaveBeenCalled();
+        });
+    });
+
```
HackTheW2d, line 830: Good test cases! I just wonder what would happen when a merchant switched from clientToken to clientId via rerender.  This should be prevented in typescript right 🤔

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -134,34 +183,41 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+            if (!authCredential) {
                 return;
             }
 
-            if (typeof clientToken === "string") {
-                token = clientToken;
-                setClientTokenValue(token);
+            if (typeof authCredential === "string") {
+                if (isClientToken) {
+                    setClientTokenValue(authCredential);
```
HackTheW2d: Do we want to also clear clientId value here by setting it undefined? Considering switching edge cases :)

packages/react-paypal-js/src/v6/components/PayPalProvider.test.tsx
nityasp: Great point. Ideally in production i don't think the merchant can switch between the auth methods at the runtime. @EvanReinstein Please correct me if I'm wrong. 
However the developers can play around and try to switch in dev environments. Also typescript doesn't prevent this switching. Added a fix to safeguard this edge case.

EvanReinstein: I'm learning that there are merchants that will switch to new clientIds but I don't know if they'll switch between clientToken and clientId.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -218,18 +231,50 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-                    clientToken: clientTokenValue,
                     components: memoizedComponents,
                     locale,
                     pageType,
                     partnerAttributionId,
                     shopperSessionId,
                     testBuyerCountry,
-                    merchantId: memoizedMerchantId,
-                });
+                };
```
EvanReinstein, line 275: Please don't forget to persist `merchantId: memoizedMerchantId`

```suggestion
                const baseOptions = {
clientMetadataId,
components: memoizedComponents,
locale,
pageType,
partnerAttributionId,
shopperSessionId,
testBuyerCountry,
merchantId: memoizedMerchantId
                };
```

.changeset/solid-memes-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/react-paypal-js": minor
+---
+
+Updating PayPalProvider with clientId enhancements
```
HackTheW2d, line 5: Can we make this change set more specific as it's user-facing change :)
something like:
```
  Added support for `clientId` as an authentication credential option in PayPalProvider to provide flexibility to authenticate using either `clientToken` or `clientId`.  
```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
HackTheW2d: Nice catch! 👍

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -25,26 +25,35 @@ import type {
...
+      });
+
 /**
  * {@link PayPalProvider} creates the SDK script, component scripts, runs eligibility, then
  * provides these in context to child components via the {@link usePayPal} hook.
  *
- * SDK loading is automatically deferred until clientToken is available.
- * clientToken can be a string, Promise, or undefined.
+ * SDK loading is automatically deferred until clientToken or clientId is available.
+ * Both can be either a string, Promise, or undefined.
```
EvanReinstein, line 53: I know this is mentioned elsewhere in the JS doc comment, but i wanted to make extra sure this is documented since we are not memoizing these props ourselves.

```suggestion
 * SDK loading is automatically deferred until clientToken or clientId is available.
 * Both can be either a string, Promise, or undefined.
 * 
 
    * **Important:** When passing a Promise, you must ensure referential stability
   * across renders. An unstable Promise reference (e.g., calling `fetchClientToken()` or `fetchClientId()`
   * inline) will cause the SDK to re-initialize on every render. Wrap the
   * Promise in `useMemo` or store it in state.

```

PR #818, opened 2026-02-11
Fix: Update Internal Button Props onError

EvanReinstein:
This PR uses `Omit` to allow an internal definition of `onError` using `OnErrorData` for the `paypal-button` and `venmo-button`


nityasp (approved):
Looks good.

PR #819, opened 2026-02-11
Fix/use eligible methods hook

HackTheW2d:


EvanReinstein (approved):

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -133,13 +133,12 @@ export function useEligibleMethods(
                 }
             })
             .finally(() => {
-                if (isSubscribed) {
-                    setIsFetching(false);
-                }
+                setIsFetching(false);
```
nityasp: Don't we need to still check if the component is still mounted here. Without `isSubscribed` check, this still tries to update states on unmounted component.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -133,13 +133,12 @@ export function useEligibleMethods(
...
             });
 
         return () => {
             isSubscribed = false;
+            lastFetchRef.current = null; // Reset fetch tracking on unmount or dependency change
```
nityasp, line 150: Instead of setting `lastFetchRef.current` to null, can we let the dependency array trigger new fetches when needed. That way only when payload/instance changes it triggers a new fetch.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
EvanReinstein: This is a good point. In React 18+ the warning for this behavior has been removed but we should try and not be setting state on an unmounted component.

Lets make this change and make sure the hook is working as expected 👍 
```suggestion
              if (isSubscribed) {
                setIsFetching(false);
              }
```

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
EvanReinstein: Good question. I think this fix was related to StrictMode double mounting practice.

HackTheW2d: Hi @nityasp This is to pass our StrictMode test in react, Consider this flow:
  Step: Mount 1                                                             
  Without = null: fetch, lastFetchRef = {A, X}                             
  With = null: fetch, lastFetchRef = {A, X}                                                           
  ────────────────────────────────────────                                  
  Step: Unmount                                                             
  Without = null: isSubscribed=false, lastFetchRef = {A, X}                  
  With = null: isSubscribed=false, lastFetchRef=null                          
  ────────────────────────────────────────                                  
  Step: Mount 2                                                             
  Without = null: hasFetchedThisConfig=true, will skip fetch                         
  With = null: hasFetchedThisConfig=false, will do a new fetch                            
  ────────────────────────────────────────                                  
  Step: Result                                                                
  without = null: ❌  The first fetch's dispatch won't execute (isSubscribed=false), no data of eligibleMethods
  with = null: ✅ Second fetch can dispatch and there's result stored in context

I hope this illustration can clarify your confusion. Lmk if you have any other questions. Thank you!

nityasp: Thanks for explaining it in detail. I understood it.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
nityasp: Just a minor thing. I think we don't need `eligibilePaymentMethods` in the dependency array. This is the output of the effect and this causes the effect to re-run everytime there is a successful fetch.

HackTheW2d: Good catch! Adopted the change. @EvanReinstein What do you think? I believe `eligibilePaymentMethods` would cause useEffect to run again(tho prevented by if check).

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -140,26 +147,28 @@ export function useEligibleMethods(
...
-        setError,
-    ]);
+    }, [sdkInstance, memoizedPayload, dispatch, setError]);
+
+    // isLoading should be true if:
+    // 1. We're actively fetching, OR
+    // 2. We don't have eligibility data yet and no error occurred
+    // This prevents a flash where isLoading=false before the effect runs
+    const isLoading =
+        isFetching || (!eligiblePaymentMethods && !eligibilityError);
```
EvanReinstein, line 159: Nice 👍

PR #821, opened 2026-02-13
Feat: Add merchantId as a PayPalProvider prop

EvanReinstein:
This PR enables the `PayPalProvider` to accept a `merchantId` string or an array of strings and pass to `createInstance`.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
@@ -12,7 +12,7 @@ import {
     INSTANCE_LOADING_STATE,
     INSTANCE_DISPATCH_ACTION,
 } from "../types/PayPalProviderEnums";
-import { toError, useCompareMemoize } from "../utils";
+import { toError, useCompareMemoize, useDeepCompareMemoize } from "../utils";
 import { useError } from "../hooks/useError";
 
 import type {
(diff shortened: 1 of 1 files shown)


packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -107,9 +107,11 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
     testBuyerCountry,
     eligibleMethodsResponse,
     children,
+    merchantId,
     ...scriptOptions
 }) => {
     const memoizedComponents = useCompareMemoize(components);
+    const memoizedMerchantId = useDeepCompareMemoize(merchantId);
```
AleGastelum, line 114: Why do we need `useDeepCompareMemoize` instead of just `useCompareMemoize`? If I'm not mistaken, both `components` and `merchantId` could be essentially an array of strings, so pretty curious where the difference lies

HackTheW2d: same question here :)

EvanReinstein: @AleGastelum and @HackTheW2d great question. `useCompareMemoize` is typed very tightly with `Components` and isn't generic enough to handle this use case. However, I do think this highlights the need to revisit this part of the v6 wrapper. I will make a follow up task 👍

PR #822, opened 2026-02-17
Adding dataSdkIntegrationSource to the script tag for analytics purpose

nityasp:
This PR is for adding dataSdkIntegrationSource to the script tag for analytics purpose.

**How to test this:**

  1. Run the sample integration app. You will see the `data-integration-source` populated in the script tag.
      
image

  2. You can also see the data source populated in FPTI
  
image

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
@@ -29,7 +29,7 @@ type PayPalProviderProps = Omit<
     CreateInstanceOptions<readonly [Components, ...Components[]]>,
     "components" | "clientToken"
 > &
-    LoadCoreSdkScriptOptions & {
+    Omit<LoadCoreSdkScriptOptions, "dataSdkIntegrationSource"> & {
         components?: Components[];
         eligibleMethodsResponse?: FindEligiblePaymentMethodsResponse;
         children: React.ReactNode;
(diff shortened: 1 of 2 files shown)


.changeset/eighty-yaks-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
```
EvanReinstein: Would you mind updating these to `minor`?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein, line 1: Lets remove the path for merchants to set this value and use a default string "react-paypal-js" 👍

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -29,7 +29,7 @@ type PayPalProviderProps = Omit<
     CreateInstanceOptions<readonly [Components, ...Components[]]>,
     "components" | "clientToken"
 > &
-    LoadCoreSdkScriptOptions & {
+    Omit<LoadCoreSdkScriptOptions, "dataSdkIntegrationSource"> & {
```
EvanReinstein, line 32: I'm not sure we need this `Omit` helper

nityasp: If we remove this Omit, the merchants can still pass the `dataSdkIntegrationSource` to the Provider but will be overrridden by the `react-paypal-js` in the Provider. Using this Omit to avoid merchants from passing any datasource attribute to Provider.

EvanReinstein: Thank you for explaining that 👍

PR #823, opened 2026-02-17
Created card fields components

AleGastelum:
This PR introduces the following components:
- PayPalCardField
- PayPalCardNumberField
- PayPalCardExpiryField
- PayPalCardCvvField

`PayPalCardField` is the base component that actually handles field creation and rendering logic of the fields. `PayPalCardNumberField`, `PayPalCardExpiryField`, and `PayPalCardCvvField` components are just small wrapper components that wrap `PayPalCardField.tsx` component to pass the appropriate type to it, abstracting this from the merchant.

Additional to this, types in `paypal-js` where updated to account the new `destroy` methods added to Card Fields.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/PayPalCardField.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalCardField.test.tsx
@@ -98,6 +98,31 @@ describe("PayPalCardField", () => {
         expect(mockCardFieldElement.destroy).toHaveBeenCalled();
     });
 
+    it("should call field's destroy method if session becomes null", () => {
+        const mockType = "number";
+        const { container, rerender } = render(
+            <PayPalCardField type={mockType} />,
+        );
+
+        expect(mockCreateCardFieldsComponent).toHaveBeenCalledTimes(1);
+        const cardFieldContainer = container.querySelector("div");
+        expect(cardFieldContainer).toBeInTheDocument();
+        expect(cardFieldContainer).toContainElement(mockCardFieldElement);
+        expect(mockCardFieldElement.destroy).toHaveBeenCalledTimes(0);
+
+        // Update session to null to trigger destroy
+        mockUsePayPalCardFieldsSession.mockReturnValueOnce({
+            cardFieldsSession: null,
+            setCardFieldsSessionType: jest.fn(),
+            setError: jest.fn(),
+        });
+        rerender(<PayPalCardField type={mockType} />);
+
+        expect(mockCreateCardFieldsComponent).toHaveBeenCalledTimes(1);
+        expect(mockCardFieldElement.destroy).toHaveBeenCalledTimes(1);
+        expect(cardFieldContainer?.children.length).toBe(0);
+    });
+
     it("should pass latest options to the session on session change", () => {
         const mockType = "number";
         const initialConfig = {
--- a/packages/react-paypal-js/src/v6/components/PayPalCardField.tsx
+++ b/packages/react-paypal-js/src/v6/components/PayPalCardField.tsx
@@ -49,6 +49,7 @@ export const PayPalCardField = ({
         }
 
         return () => {
+            cardField?.remove();
             cardField?.destroy();
         };
         // eslint-disable-next-line react-hooks/exhaustive-deps


packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -153,7 +174,8 @@ type BaseCardFieldsSession = {
           .appendChild(numberField);
      * ```
      */
-    createCardFieldsComponent: (config: CardFieldOptions) => HTMLElement;
+
+    createCardFieldsComponent: (config: CardFieldOptions) => CardFieldComponent;
```
EvanReinstein: Will this space above the key remove the intellisense hover to show the comment?

AleGastelum: It seems to keep working with the space, but I will remove it.

packages/react-paypal-js/src/v6/components/PayPalCardCvvField.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * // Basic usage creating a CVV field
+ * <PayPalCardCvvField
+ *   placeholder="Enter CVV"
+ *   containerStyles={{ height: "3rem", marginBottom: "1rem" }}
+ * />
+ */
+export const PayPalCardCvvField = (
+    options: PayPalCardCvvFieldProps,
+): JSX.Element | null => {
+    return <PayPalCardField type="cvv" {...options} />;
```
EvanReinstein: I've learned recently that the spread operator negates some typescript functionality, for example using a spread operator can negate a discriminating union. Can you double check that the options are typed correctly?

Also, I am concerned that `options` is a non-primitive type prop and will trigger a re-render each time even though none of the options have changed. In other places we use `deepCompareMemoize` to prevent this. Let me know what you think

packages/react-paypal-js/src/v6/components/PayPalCardExpiryField.tsx
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @example
+ * // Basic usage creating an expiry field
+ * <PayPalCardExpiryField
+ *   placeholder="Enter an expiry date"
+ *   containerStyles={{ height: "3rem", marginBottom: "1rem" }}
+ * />
+ */
+export const PayPalCardExpiryField = (
+    options: PayPalCardExpiryFieldProps,
```
EvanReinstein: Similar `options` non-primitive comment as above, as well as spread operator curiousity. 👍

packages/react-paypal-js/src/v6/components/PayPalCardField.tsx
```diff
@@ -0,0 +1,64 @@
...
+            cardFieldContainerRef.current?.appendChild(cardField);
+        } catch (error) {
+            setError(toError(error));
+        }
+
+        return () => {
+            cardField?.destroy();
+        };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [cardFieldsSession]);
```
EvanReinstein, line 56: Is this useEffect ignoring `options` updates? Should merchants be able to update the field?

.changeset/cold-regions-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
```
EvanReinstein: Would you mind updating this to `minor`?

.changeset/quick-chefs-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
```
EvanReinstein: Same comment as above, please update to `minor`

packages/paypal-js/types/v6/components/card-fields.d.ts
```diff
@@ -135,6 +135,27 @@ export type CardFieldOptions = {
...
 
+export type CardFieldComponent = HTMLElement & {
+    /**
+     * Use this method to destroy the Card Field component and clean up any associated resources. After calling this method, the Card Field component will no longer be usable.
+     * 
+     * @example
+     * ```typescript
+     * const numberField = cardFieldsInstance.createCardFieldsComponent({
+             type: "number",
+            placeholder: "Enter a number",
```
EvanReinstein:
```suggestion
            type: "number",
            placeholder: "Enter a number",
```

packages/react-paypal-js/src/v6/components/PayPalCardNumberField.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * // Basic usage creating a number field
+ * <PayPalCardNumberField
+ *   placeholder="Enter a card number"
+ *   containerStyles={{ height: "3rem", marginBottom: "1rem" }}
+ * />
+ */
+export const PayPalCardNumberField = (
+    options: PayPalCardNumberFieldProps,
+): JSX.Element | null => {
+    return <PayPalCardField type="number" {...options} />;
```
EvanReinstein: Same comment re: non-primitive `options` and the spread operator 👍

packages/react-paypal-js/src/v6/components/PayPalCardCvvField.tsx
EvanReinstein: Ok, finally got to `PayPalCardField` and saw the Ref there!

packages/react-paypal-js/src/v6/components/PayPalCardField.test.tsx
EvanReinstein, line 1: I am pretty sure there is a missing test case here, for when the session transitions from non-null to null while the component stays mounted (when sdkInstance changes and the provider resets). 

We should verify that destroy() is called on the existing field and the container is empty afterward.

.changeset/cold-regions-rescue.md
AleGastelum: Updated!

.changeset/quick-chefs-occur.md
AleGastelum: Updated as well!

packages/react-paypal-js/src/v6/components/PayPalCardNumberField.tsx
AleGastelum: Removed the use of the spread operator and made it more explicit

packages/react-paypal-js/src/v6/components/PayPalCardExpiryField.tsx
AleGastelum: Removed the use of the spread operator and made it more explicit

packages/react-paypal-js/src/v6/components/PayPalCardCvvField.tsx
AleGastelum: I removed the use of the spread operator and made it more explicit.

For the options concern, this current Card Fields implementation won't trigger re-renders if the options change, as the `PayPalCardField` component stores that object in a ref. Current implementation won't react if options object change after field creation.

packages/react-paypal-js/src/v6/components/PayPalCardField.tsx
AleGastelum: That is correct, current implementation is ignoring `option` updates, as updating options after field creation is not a feature the Card Fields Web SDK provides currently.

packages/react-paypal-js/src/v6/components/PayPalCardField.test.tsx
AleGastelum: Good call, I added the missing test named `"should call field's destroy method if session becomes null"`

PR #826, opened 2026-02-19
Feat: Add savePayment Property

EvanReinstein:
This also passes the `testBuyerCountry` to the `createOneTimePaymentSession` method.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
@@ -137,6 +137,7 @@ export function usePayPalOneTimePaymentSession({
         presentationMode,
         setError,
         savePayment,
+        testBuyerCountry,
     ]);
 
     const handleClick = useCallback(async () => {


packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
```diff
@@ -132,6 +136,7 @@ export function usePayPalOneTimePaymentSession({
         handleDestroy,
         presentationMode,
         setError,
+        savePayment,
```
gregjopa, line 139: Should we add `testBuyerCountry` too?

EvanReinstein: This is part of the base session options I believe

EvanReinstein: For posterity's sake here is the [type](https://github.com/paypal/paypal-js/blob/f047f6531ef7aae08526b2e8f8023c81fef21ef2/packages/paypal-js/types/v6/components/base-component.d.ts#L39-L45)

EvanReinstein: Ah, yes your original question was for the dependency array! I've got to drink some more coffee 😆

PR #829, opened 2026-02-20
Feature/fix bug for eligible methods

HackTheW2d:
Fix the payload change issue: when users move around page to use different payload, the fetch-eligible-methods call won't be triggered


packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -8,7 +8,7 @@ import {
     type FindEligibleMethodsOptions,
 } from "../types";
 import { useError } from "./useError";
-import { useDeepCompareMemoize } from "../utils";
+import { deepEqual, useDeepCompareMemoize } from "../utils";
```
EvanReinstein, line 11: this may be the first use of deepEqual, another example for why these utilities need to be revisited. This looks good 👍

PR #833, opened 2026-02-23
Feat: Adding error handling to react hooks

nityasp:
This PR adds comprehensive error handling to all payment session hooks to help merchants quickly identify and resolve SDK configuration issues when session creation methods throw errors due to missing components.

Implemented a consistent error handling pattern across all the payment session hooks.
1. Catches session creation errors with try-catch blocks
2. Provides a detailed error message.
3. Prevents infinite retry looks using 2 ref-based guards:
    - `creationErrorRef`: Prevents retrying session creation after failure
    - `lastSdkInstanceREf`: Allows recovery when SDK instance changes
4. Uses two separate `useEffect` hooks to avoid infinite re-render loops cause by state updates.

Hooks updated:
   - `usePayPalOneTimePaymentSession`
   - `usePayPalSavePaymentSession`
   - `usePayPalSubscriptionPaymentSession`
   - `useVenmoOneTimePaymentSession`
   - `usePayPalGuestPaymentSession`
   - `usePayPalCreditSavePaymentSession`
   - `usePayPalCreditOneTimePaymentSession`
   - `usePayLaterOneTimePaymentSession`

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.test.ts
@@ -23,7 +23,7 @@ const mockUseProxyProps = useProxyProps as jest.MockedFunction<
     typeof useProxyProps
 >;
 
-const createMockSavePaymentSession = (): SavePaymentSession => ({
+const createMockPayPalCreditSavePaymentSession = (): SavePaymentSession => ({
     start: jest.fn().mockResolvedValue(undefined),
     cancel: jest.fn(),
     destroy: jest.fn(),
@@ -32,7 +32,7 @@ const createMockSavePaymentSession = (): SavePaymentSession => ({
 });
 
 const createMockSdkInstance = (
-    savePaymentSession = createMockSavePaymentSession(),
+    savePaymentSession = createMockPayPalCreditSavePaymentSession(),
 ) => ({
     createPayPalSavePaymentSession: jest
         .fn()
@@ -46,7 +46,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
     beforeEach(() => {
         mockUseProxyProps.mockImplementation((callbacks) => callbacks);
 
-        mockSavePaymentSession = createMockSavePaymentSession();
+        mockSavePaymentSession = createMockPayPalCreditSavePaymentSession();
         mockSdkInstance = createMockSdkInstance(mockSavePaymentSession);
 
         mockPayPalContext({ sdkInstance: mockSdkInstance });
@@ -101,7 +101,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
         });
 
         test("should clear any sdkInstance related errors if the sdkInstance becomes available", () => {
-            const mockSession = createMockSavePaymentSession();
+            const mockSession = createMockPayPalCreditSavePaymentSession();
             const mockSdkInstanceNew = createMockSdkInstance(mockSession);
 
             mockPayPalRejected();
@@ -345,7 +345,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
 
             jest.clearAllMocks();
 
-            const newMockSession = createMockSavePaymentSession();
+            const newMockSession = createMockPayPalCreditSavePaymentSession();
             const newMockSdkInstance = createMockSdkInstance(newMockSession);
 
             mockPayPalContext({ sdkInstance: newMockSdkInstance });
@@ -389,7 +389,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
 
     describe("resume flow", () => {
         test("should check hasReturned and call resume when presentationMode is redirect", async () => {
-            const mockSession = createMockSavePaymentSession();
+            const mockSession = createMockPayPalCreditSavePaymentSession();
             (mockSession.hasReturned as jest.Mock).mockReturnValue(true);
             const mockSdk = createMockSdkInstance(mockSession);
 
@@ -410,7 +410,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
         });
 
         test("should check hasReturned and call resume when presentationMode is direct-app-switch", async () => {
-            const mockSession = createMockSavePaymentSession();
+            const mockSession = createMockPayPalCreditSavePaymentSession();
             (mockSession.hasReturned as jest.Mock).mockReturnValue(true);
             const mockSdk = createMockSdkInstance(mockSession);
 
@@ -431,7 +431,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
         });
 
         test("should not call resume if hasReturned returns false", async () => {
-            const mockSession = createMockSavePaymentSession();
+            const mockSession = createMockPayPalCreditSavePaymentSession();
             (mockSession.hasReturned as jest.Mock).mockReturnValue(false);
             const mockSdk = createMockSdkInstance(mockSession);
 
@@ -465,7 +465,7 @@ describe("usePayPalCreditSavePaymentSession", () => {
         });
 
         test("should handle errors during resume flow", async () => {
-            const mockSession = createMockSavePaymentSession();
+            const mockSession = createMockPayPalCreditSavePaymentSession();
             const resumeError = new Error("resume failed");
             (mockSession.hasReturned as jest.Mock).mockReturnValue(true);
             (mockSession.resume as jest.Mock).mockRejectedValue(resumeError);
@@ -607,7 +607,8 @@ describe("usePayPalCreditSavePaymentSession", () => {
                 });
 
                 jest.clearAllMocks();
-                mockSavePaymentSession = createMockSavePaymentSession();
+                mockSavePaymentSession =
+                    createMockPayPalCreditSavePaymentSession();
                 mockSdkInstance = createMockSdkInstance(mockSavePaymentSession);
                 mockPayPalContext({ sdkInstance: mockSdkInstance });
             }
--- a/packages/react-paypal-js/src/v6/utils.ts
+++ b/packages/react-paypal-js/src/v6/utils.ts
@@ -2,17 +2,6 @@ import { useRef } from "react";
 
 import type { Components } from "./types";
 
-// Extend the Window interface to include __paypal_sdk__
-declare global {
-    interface Window {
-        __paypal_sdk__?: {
-            v6: {
-                components?: string[];
-            };
-        };
-    }
-}
-
 /**
  * Performs a shallow equality check on two arrays.
  *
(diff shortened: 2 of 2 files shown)


packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
```diff
@@ -81,53 +86,73 @@ export function usePayPalOneTimePaymentSession({
...
-            };
-
-            handleReturnFromPayPal();
+        if (creationErrorRef.current) {
+            return;
         }
 
+        try {
+            const newSession = sdkInstance.createPayPalOneTimePaymentSession({
+                orderId,
```
HackTheW2d: so in the original code, we also have `savePayment`, `testBuyerCountry`. May I know why we want to drop them here? :)

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
```diff
@@ -81,53 +86,73 @@ export function usePayPalOneTimePaymentSession({
...
+                            await newSession.resume?.();
+                        }
+                    } catch (err) {
+                        setError(err as Error);
+                    }
+                };
+
+                handleReturnFromPayPal();
+            }
+        } catch (err) {
```
HackTheW2d: I see this pattern repeated for 8 times and the difference is the messages it conveyed. Do you think we can make this into a utility function? eg.
```
  export function createSessionSafely<T>(
      createFn: () => T,
      sessionLabel: string,
      failedSdkRef: React.MutableRefObject<unknown>,
      sdkInstance: unknown,
      setError: TypeSetError,
  ): T | null {
      if (failedSdkRef.current === sdkInstance) {
          return null;
      }

      try {
          return createFn();
      } catch (err) {
          failedSdkRef.current = sdkInstance;
          setError(
              new Error(
                  `Failed to create ${sessionLabel}. ` +
                  "This may occur if the required
  components are not included " +
                  "in the SDK components array.",
                  { cause: err },
              ),
          );
          return null;
      }
  }
```

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
nityasp: Nice catch. Yes looks like that got delete when I merged with the main. Its back now.

packages/react-paypal-js/src/v6/hooks/usePayPalSavePaymentSession.ts
```diff
@@ -57,53 +57,83 @@ export function usePayPalSavePaymentSession({
     const sessionRef = useRef<SavePaymentSession | null>(null);
     const proxyCallbacks = useProxyProps(callbacks);
     const [error, setError] = useError();
+
+    // Track if we encountered a creation error to prevent infinite retry loops
+    const creationErrorRef = useRef(false);
```
HackTheW2d: I don't know if I understand this correctly. Maybe for this variable we can try only using `lastSdkInstanceRef` to achieve same effect?
for instance:
```
// rename lastSdkInstanceRef to failedSdkRef and we can init it using null
const failedSdkRef = useRef<unknown>(null);

// useEffect block
if (failedSdkRef.current === sdkInstance) {
      return;
  }

// then in catch block
failedSdkRef.current = sdkInstance;
```
so in this way we can reduce using 2 variables to 1. What do you think? :)

packages/react-paypal-js/src/v6/hooks/usePayPalOneTimePaymentSession.ts
nityasp: Sounds good. I'll consolidate the code accordingly.

packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts
```diff
@@ -39,54 +39,84 @@ export function usePayLaterOneTimePaymentSession({
...
+                handleReturnFromPayPal();
+            }
+        } catch (err) {
+            creationErrorRef.current = true;
+
+            const detailedError = new Error(
+                "Failed to create PayLater one-time payment session. " +
+                    "This may occur if the required components are not included in the SDK components array. " +
+                    "Please ensure you have added the necessary components when loading the PayPal SDK. " +
+                    `Original error: ${err instanceof Error ? err.message : String(err)}`,
```
HackTheW2d: Do you think we can try using `{ cause: err }` here so we can drop the condition checking.
[Reference](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)
And this way developers can also see full stack trace.

nityasp: I got build error when I use `{ cause: err }` because we are currently using ES5 as the compilation target. Whereas `{ cause: err}` was from ES2022+. Due to that I couldn't use it directly. However to get the cause details, we can manually assign like this `(detailedError as Error & { cause: unknown }).cause = err;`

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
```diff
@@ -75,30 +79,47 @@ export function usePayPalGuestPaymentSession({
...
+
+        if (!newSession) {
+            return;
+        }
+
         sessionRef.current = newSession;
 
         return () => {
-            newSession.destroy();
+            sessionRef.current?.destroy();
```
EvanReinstein: Should this still be `newSession.destroy`? There could be an issue if the `sessionRef.current` is overwritten between effect setup and clean up, resulting in the wrong session getting destroyed.

packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.test.ts
```diff
@@ -127,6 +128,55 @@ describe("usePayPalCreditSavePaymentSession", () => {
...
+            {
+                description: "Error object",
+                thrownError: new Error("Required components not loaded in SDK"),
+            },
+            {
+                description: "non-Error string",
+                thrownError: "String error message",
+            },
+        ])(
+            "should handle $description thrown by createPayPalSavePaymentSession",
```
EvanReinstein:
```suggestion
            "should handle $description thrown by createPayPalCreditSavePaymentSession",
```

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -222,3 +222,50 @@ export function deepEqual(
...
+        return sessionCreator();
+    } catch (err) {
+        failedSdkRef.current = sdkInstance;
+
+        const detailedError = new Error(
+            "Failed to create payment session. " +
+                "This may occur if the required components are not included in the SDK components array. " +
+                "Please ensure you have added the necessary components when loading the PayPal SDK.",
+        );
+        (detailedError as Error & { cause: unknown }).cause = err;
```
EvanReinstein: I do wonder here if this is too generic and should include some reference to the correct component to pass?

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -222,3 +222,50 @@ export function deepEqual(
...
+ * );
+ *
+ * if (!session) return;
+ */
+export function createPaymentSession<T>(
+    sessionCreator: () => T,
+    failedSdkRef: { current: unknown },
+    sdkInstance: unknown,
+    setError: (error: Error | null) => void,
+): T | null {
```
EvanReinstein, line 253: What if we added a `component` key here and used that to enhance the error?

packages/react-paypal-js/src/v6/utils.ts
EvanReinstein: Also i've run into the Error/cause issue re: ES2022 that you mention [here](https://github.com/paypal/paypal-js/pull/833#discussion_r2847808459) I wonder if there would be any issue with updating the TS config to account for this?

gregjopa: We can technically use the internal global window namespace to check and see what components have been loaded on the page if we want confidence in the "missing required component" error message. Ex:

```js
// array containing all the components loaded on the page
window.__paypal_sdk__.v6.components
```

packages/react-paypal-js/src/v6/hooks/usePayPalGuestPaymentSession.ts
nityasp: You are right. It should be `newSession.destroy()` since we need to capture specific instance created in that effect run via closure. 

Looking at the other hooks, I see we are returning `handleDestroy()` [here](https://github.com/paypal/paypal-js/blob/823a8139ff62ffddb61da474658d7091bc2c15b7/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts#L92) which in turn is doing `sessionRef.current?.destroy()` [here](https://github.com/paypal/paypal-js/blob/823a8139ff62ffddb61da474658d7091bc2c15b7/packages/react-paypal-js/src/v6/hooks/usePayLaterOneTimePaymentSession.ts#L44) which created circular dependency where the effect depends on `handleDestroy`, which would need to depend on effect. Fixing those hooks to return `newSession.destroy()`

I'm still keeping the `handleDestroy()` callback for users to manually destroy current session.

packages/react-paypal-js/src/v6/utils.ts
nityasp: yes we can do that. I've updated the tsconfig now.

packages/react-paypal-js/src/v6/hooks/usePayPalCreditSavePaymentSession.test.ts
EvanReinstein: Can we update this to be:
```suggestion
const createMockPayPalCreditSavePaymentSession = (): SavePaymentSession => ({
start: jest.fn().mockResolvedValue(undefined),
cancel: jest.fn(),
destroy: jest.fn(),
hasReturned: jest.fn().mockReturnValue(false),
resume: jest.fn().mockResolvedValue(undefined),
});
```

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -2,6 +2,17 @@ import { useRef } from "react";
...
+declare global {
+    interface Window {
+        __paypal_sdk__?: {
+            v6: {
+                components?: string[];
+            };
+        };
+    }
+}
+
```
EvanReinstein: I don't think we want to declare this globally, since the SDK should be the only entity that looks at `__paypal_sdk__`. I think that we can replace `loadedComponents` with my suggestion below and remove this.

packages/react-paypal-js/src/v6/utils.ts
```diff
@@ -222,3 +233,75 @@ export function deepEqual(
...
+    if (failedSdkRef.current === sdkInstance) {
+        return null;
+    }
+
+    try {
+        return sessionCreator();
+    } catch (err) {
+        failedSdkRef.current = sdkInstance;
+
+        const loadedComponents = window.__paypal_sdk__?.v6?.components;
```
EvanReinstein:
```suggestion
                const loadedComponents = (
            window as Window & {
                __paypal_sdk__?: { v6: { components?: string[] } };
            }
        ).__paypal_sdk__?.v6?.components;
```

PR #845, opened 2026-02-27
Fix: Enhance loadCoreSdkScript

EvanReinstein:
`loadCoreSdkScript` has a condition that prevents duplicate script loads, however there is an edge-case in the current logic. If loadCoreSdkScript is called twice (like by React StrictMode for example) the core script is added to the DOM and is in the process of loading, i.e. `window.paypal` is still undefined, when the function runs for the second time. The current condition evaluates to false, resulting in another script added to the DOM.

This change updates `loadCoreSdkScript` to prevent the double load.

### DIFF
--- a/packages/paypal-js/src/v6/index.ts
+++ b/packages/paypal-js/src/v6/index.ts
@@ -61,9 +61,7 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
                 () => {
                     reject(
                         new Error(
-                            `The script "${currentScript.src}" failed to load.
-                        Check the HTTP status code and response body in DevTools to
-                        learn more.`,
+                            `The script "${currentScript.src}" failed to load. Check the HTTP status code and response body in DevTools to learn more.`,
                         ),
                     );
                 },
--- a/packages/paypal-js/src/v6/index.test.ts
+++ b/packages/paypal-js/src/v6/index.test.ts
@@ -268,7 +268,7 @@ describe("loadCoreSdkScript()", () => {
             script.dispatchEvent(new Event("error"));
 
             await expect(promise).rejects.toThrow(
-                "The PayPal SDK script failed to load.",
+                `The script "https://www.sandbox.paypal.com/web-sdk/v6/core" failed to load. Check the HTTP status code and response body in DevTools to learn more.`,
             );
             expect(mockedInsertScriptElement).not.toHaveBeenCalled();
         });


nityasp (approved):
LGTM!!

packages/paypal-js/src/v6/index.ts
```diff
@@ -18,10 +19,41 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
     );
 
-    if (window.paypal?.version.startsWith("6") && currentScript) {
+    // Script already loaded and namespace is available — return immediately
+    if (window.paypal?.version?.startsWith("6") && currentScript) {
         return Promise.resolve(window.paypal as unknown as PayPalV6Namespace);
     }
 
+    // Script tag exists but hasn't finished loading yet (e.g., React StrictMode double-invoke)
+    if (currentScript) {
```
HackTheW2d, line 40: Do we need to add some tests for this situation? since this impacts the load logic 🤔

EvanReinstein: Yea great point, I will push some 👍

packages/paypal-js/src/v6/index.ts
```diff
@@ -18,10 +19,41 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
         'script[src*="/web-sdk/v6/core"]',
     );
 
-    if (window.paypal?.version.startsWith("6") && currentScript) {
+    // Script already loaded and namespace is available — return immediately
+    if (window.paypal?.version?.startsWith("6") && currentScript) {
```
gregjopa: Does this need to take into account the custom namespace option?

```suggestion
  const windowNamespace = options.dataNamespace ?? "paypal";
        if (window[windowNamespace]?.version?.startsWith("6") && currentScript) {
```

packages/paypal-js/src/v6/index.ts
```diff
@@ -18,10 +19,41 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
         return Promise.resolve(window.paypal as unknown as PayPalV6Namespace);
```
gregjopa: Same here with what gets returned.
```suggestion
        return Promise.resolve(window[windowNamespace] as unknown as PayPalV6Namespace);
```

packages/paypal-js/src/v6/index.ts
HackTheW2d: Great catch! Same doubt here. But don't we always load at window.paypal internally? 🤔 Correct me if I am wrong :)

packages/paypal-js/src/v6/index.ts
```diff
@@ -18,10 +19,41 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+                            `The window.${namespace} global variable is not available`,
+                        );
+                    }
+                },
+                { once: true },
+            );
+            currentScript.addEventListener(
+                "error",
+                () => {
+                    reject(new Error(`The PayPal SDK script failed to load.`));
```
HackTheW2d: This error message seems a bit vague to me. Can we make it more specific? :)
```suggestion
                     reject(new Error(
                            `The script "${currentScript.src}" failed to load.
                    Check the HTTP status code and response body in DevTools to
                    learn more.`
                     ));
```

packages/paypal-js/src/v6/index.ts
gregjopa: In V6 we use `window__paypal_sdk__` as the internal namespace. Here's an example when using `data-namespace="paypalV6"`:

```html
<script async="" src="https://www.sandbox.paypal.com/web-sdk/v6/core" data-namespace="paypalV6"></script>
```

Ex:

Image

packages/paypal-js/src/v6/index.ts
```diff
@@ -18,10 +19,41 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+    // Script already loaded and namespace is available — return immediately
+    if (window.paypal?.version?.startsWith("6") && currentScript) {
         return Promise.resolve(window.paypal as unknown as PayPalV6Namespace);
     }
 
+    // Script tag exists but hasn't finished loading yet (e.g., React StrictMode double-invoke)
+    if (currentScript) {
+        return new Promise<PayPalV6Namespace>((resolve, reject) => {
+            const namespace = options.dataNamespace ?? "paypal";
+            currentScript.addEventListener(
```
gregjopa, line 43: If we go this route, we should look into sharing these callbacks with what is passed into `insertScriptElement` with onSuccess and onError. That function uses these same event listeners under the hood

packages/paypal-js/src/v6/index.ts
EvanReinstein: Screenshot 2026-02-27 at 3 14 10 PM

Confirmed that this update works with the custom namespace as well as `window.paypal`

packages/paypal-js/src/v6/index.ts
HackTheW2d: Thanks for the clarification! TIL

HackTheW2d (approved):

packages/paypal-js/src/v6/index.ts
EvanReinstein: Ok i'll add this to my near-term ToDos

EvanReinstein:
Will go with the solution offered in https://github.com/paypal/paypal-js/pull/848 👍

PR #846, opened 2026-02-27
Teach clientId for V6 SDK Instance Creation in READMEs

EvanReinstein:
Updates READMEs to prefer `clientId` for SDK instance creation.


packages/react-paypal-js/README.md
HackTheW2d:
```suggestion
Alternative: With Promise-based Token
```

PR #849, opened 2026-03-02
feat(v6): add support for deterministic loading

gregjopa:
This PR adds a `data-loading-state` attribute to the V6 Core SDK script tag. This is leveraged to ensure that the v6 script only loads once, regardless of how many times the `loadCoreSdkScript()` function is called. This will help resolved any issue around `useEffect` running twice in react in strict mode.

### DIFF
--- a/packages/paypal-js/src/v6/index.test.ts
+++ b/packages/paypal-js/src/v6/index.test.ts
@@ -68,7 +68,7 @@ describe("loadCoreSdkScript()", () => {
         expect(window.paypal).toBeDefined();
     });
 
-    test("should avoid inserting two script elements when called twice", async () => {
+    test("should avoid inserting two script elements when called twice sequentially", async () => {
         const result1 = await loadCoreSdkScript();
         const result2 = await loadCoreSdkScript();
         // should only insert the script once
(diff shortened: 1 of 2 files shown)


packages/paypal-js/src/v6/index.ts
```diff
@@ -36,42 +39,53 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+                return reject(
+                    `The window.${namespace} global variable is not available`,
+                );
+            }
+            scriptElement.setAttribute(
+                DATA_ATTRIBUTE_LOADING_STATE,
+                SCRIPT_LOADING_STATE.RESOLVED,
+            );
+            return resolve(paypalWindowReference);
+        });
```
HackTheW2d:
```suggestion
        }, { once: true });
```
to remove listener when triggered

packages/paypal-js/src/v6/index.ts
```diff
@@ -36,42 +39,53 @@ function loadCoreSdkScript(options: LoadCoreSdkScriptOptions = {}) {
...
+            const defaultError = new Error(
+                `The script "${url.toString()}" failed to load. Check the HTTP status code and response body in DevTools to learn more.`,
+            );
+
+            scriptElement.setAttribute(
+                DATA_ATTRIBUTE_LOADING_STATE,
+                SCRIPT_LOADING_STATE.REJECTED,
+            );
+            return reject(defaultError);
         });
```
HackTheW2d:
```suggestion
        }, { once: true });
```
same here

packages/paypal-js/src/v6/index.test.ts
```diff
@@ -1,80 +1,165 @@
...
         );
-        expect(mockedInsertScriptElement).toHaveBeenCalledTimes(1);
+        expect(scriptElement.getAttribute("data-loading-state")).toBe(
+            "resolved",
+        );
+        expect(result).toBeDefined();
+        expect(window.paypal).toBeDefined();
+    });
+
+    test("should avoid inserting two script elements when called twice", async () => {
```
HackTheW2d: currently we await the first load to be fully resolved then trigger the second one.

Do we want to add one more test to test concurrency like this? :) Thanks!

```typescript
  test("should handle concurrent calls without duplicating script elements",    
  async () => {   
      // no await here!                         
      const promise1 = loadCoreSdkScript();
      const promise2 = loadCoreSdkScript();

      const [result1, result2] = await Promise.all([promise1, promise2]);

      // and we only expect one insert
      expect(scriptAppendChildSpy).toHaveBeenCalledTimes(1);

      // both get resolved correctly without hanging
      expect(result1).toBeDefined();
      expect(result2).toBeDefined();

      // they get same ref
      expect(result1).toBe(result2);

      // scriptElement to be resolved
      const scriptElement = scriptAppendChildSpy.mock.calls[0][0];
      expect(scriptElement.getAttribute("data-loading-state")).toBe("resolved");
  });

```

packages/paypal-js/src/v6/index.ts
gregjopa: Great suggestion @HackTheW2d! I added this `once` option to both listeners

packages/paypal-js/src/v6/index.test.ts
gregjopa: I dig it! I added this parallel test 💯

PR #852, opened 2026-03-03
chore(types): rename v6 subscription onApprove type

gregjopa:


packages/paypal-js/types/v6/components/paypal-subscriptions.d.ts
```diff
@@ -29,7 +29,10 @@ export type PayPalSubscriptionSessionOptions = Omit<
     ) => Promise<void>;
 };
 
-export type PayPalSubscriptionPaymentSession = BasePaymentSession & {
+export type PayPalSubscriptionPaymentSession = Omit<
+    BasePaymentSession,
+    "start"
```
gregjopa, line 34: We have to omit the existing `start()` type to ensure `orderId` is not allowed, and that only `subscriptionId` is allowed with the createSubscriptionPromise

HackTheW2d (approved):

PR #855, opened 2026-03-12
Feature/google payments types

HackTheW2d:
Add google payments types to v6 vanilla JS types

### DIFF
--- a/packages/paypal-js/types/v6/tests/googlepay-payments.test.ts
+++ b/packages/paypal-js/types/v6/tests/googlepay-payments.test.ts
@@ -125,4 +125,56 @@ async function main() {
     // Verify GooglePayPaymentsInstance narrowing from SdkInstance
     const instance: GooglePayPaymentsInstance = sdkInstance;
     instance.createGooglePayOneTimePaymentSession();
+
+    // Verify getDetails config can be passed directly to formatConfigForPaymentRequest
+    const mockConfig: GooglePayConfigFromFindEligibleMethods = {
+        eligible: true,
+        merchantCountry: "US",
+        apiVersion: 2,
+        apiVersionMinor: 0,
+        allowedPaymentMethods: [
+            {
+                type: "CARD",
+                parameters: {
+                    allowedAuthMethods: ["PAN_ONLY", "CRYPTOGRAM_3DS"],
+                    supportedNetworks: [
+                        "VISA",
+                        "MASTERCARD",
+                        "AMEX",
+                        "DISCOVER",
+                    ],
+                    billingAddressRequired: true,
+                    assuranceDetailsRequired: true,
+                    billingAddressParameters: {
+                        format: "FULL",
+                    },
+                },
+                tokenizationSpecification: {
+                    type: "PAYMENT_GATEWAY",
+                    parameters: {
+                        gateway: "paypalsb",
+                        gatewayMerchantId: "mockGatewayMerchantId",
+                    },
+                },
+            },
+        ],
+        merchantInfo: {
+            merchantOrigin: "https://example.com",
+            merchantId: "mockMerchantId",
+        },
+    };
+
+    // This is the merchant's primary flow: getDetails → formatConfigForPaymentRequest
+    const formattedConfig: GooglePayConfig =
+        googlePaySession.formatConfigForPaymentRequest(mockConfig);
+
+    // Verify output matches Google Pay API requirements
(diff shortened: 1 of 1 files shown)


packages/paypal-js/types/v6/components/googlepay-payments.d.ts
```diff
@@ -0,0 +1,369 @@
...
+     */
+    cancel(): void;
+
+    /**
+     * Destroys the session
+     *
+     * @remarks
+     * Performs complete cleanup and makes the session unusable.
+     */
+    destroy(): void;
```
EvanReinstein: Would you mind double checking whether these methods are included on the Session class?

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
```diff
@@ -0,0 +1,369 @@
+/* eslint-disable tsdoc/syntax */
+/**
+ * @module
+ * @internal
+ */
```
EvanReinstein, line 4: I think we should remove this `@internal` tag since there are types being exported publically from the file.

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
```diff
@@ -0,0 +1,369 @@
...
+};
+
+/**
+ * Options for confirming a Google Pay order
+ *
+ * @remarks
+ * These are the parameters passed to session.confirmOrder().
+ * The clientId is managed internally by the SDK.
+ */
+type GooglePayConfirmOrderOptions = {
```
EvanReinstein: What do you think about exporting this? It would be helpful for a dev typing the options object, in case it is created dynamically or something.

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
HackTheW2d: Yeah. Good callout! I was thinking if it's helpful when merchants cancel the payment. But looks like it should be handled by google side. I'll remove this

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
HackTheW2d: Fixed :)

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
HackTheW2d: Fixed :)

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
```diff
@@ -0,0 +1,350 @@
+/* eslint-disable tsdoc/syntax */
+/**
+ * @module
+ */
+/* eslint-enable tsdoc/syntax */
```
EvanReinstein, line 5: I am taking a closer look at the `base-component.d.ts` and I see this syntax below, and I am realizing that my prior comment may have been incorrect or misguided. I am not familiar enough with our TS Docs approach to know how the current iteration with just `@module` will effect this generated site https://paypal.github.io/paypal-js/web-sdk-v6-reference/modules.html 

```
/* eslint-disable tsdoc/syntax */
/**
 * @module
 * @internal
 */
/* eslint-enable tsdoc/syntax */
```

HackTheW2d: Resolved, all can be shown on the docs generated. :)

HackTheW2d: image

This is current public interface FYI

packages/paypal-js/types/v6/components/googlepay-payments.d.ts
```diff
@@ -0,0 +1,355 @@
...
+                gatewayMerchantId: string;
+            };
+        };
+    }>;
+    merchantInfo: {
+        merchantOrigin: string;
+        merchantId: string;
+        authJwt?: string;
+    };
+};
```
hamzanasir, line 90: This is the raw response from the find eligible methods right? In that case shouldn't it be snake cased?

gregjopa: I think the scope of these types is for typing out the response of `sdkInstance.findEligibileMethods()` or `sdkInstance.hydrateEligibleMethods()`. Both of these functions will always return the camelCase versions.

PR #859, opened 2026-03-17
Adding Apple Pay paypal-js types

nityasp:
In this PR, we are adding Apple Pay payment types to `paypal-js`.

### DIFF
--- a/packages/paypal-js/types/v6/components/applepay-payments.d.ts
+++ b/packages/paypal-js/types/v6/components/applepay-payments.d.ts
@@ -95,6 +95,23 @@ export type ConfirmOrderOptions = {
 };
 
 /**
+ * Response from confirming an Apple Pay order.
+ */
+export type ConfirmOrderResponse = {
+    approveApplePayPayment: {
+        id: string;
+        status: string;
+        payment_source: {
+            apple_pay: {
+                name: string;
+                card: unknown;
+            };
+        };
+        links?: unknown[];
+    };
+};
+
+/**
  * Apple Pay payment request configuration returned from eligible methods.
  */
 export type ApplePayConfig = {
@@ -160,21 +177,27 @@ export type ApplePayOneTimePaymentSession = {
      * Confirms the order with PayPal after Apple Pay authorization.
      *
      * @param options - Confirmation options including order ID, token, and contact information
-     * @returns Promise that resolves when the order is confirmed
+     * @returns Promise resolving to the confirmed order response
      *
      * @example
      * ```typescript
      * appleSdkSession.onpaymentauthorized = async (event) => {
+     *   const order = await createOrder();
      *   await applePaySession.confirmOrder({
-     *     orderId: createdOrder.orderId,
+     *     orderId: order.orderId,
      *     token: event.payment.token,
      *     billingContact: event.payment.billingContact,
      *     shippingContact: event.payment.shippingContact,
      *   });
+     *   appleSdkSession.completePayment({
+     *     status: ApplePaySession.STATUS_SUCCESS,
+     *   });
      * };
      * ```
      */
-    confirmOrder: (options: ConfirmOrderOptions) => Promise<void>;
+    confirmOrder: (
+        options: ConfirmOrderOptions,
+    ) => Promise<ConfirmOrderResponse>;
 };
 
 /**


.changeset/eleven-women-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
```
EvanReinstein: Would you mind updating this to minor 👍

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,257 @@
...
+    };
+};
+
+/**
+ * Options for validating Apple Pay merchant.
+ */
+export type ValidateMerchantOptions = {
+    validationUrl: string;
+    displayName?: string;
+    domainName?: string;
```
HackTheW2d: I see in `ValidateApplePayMerchantOptions`, `displayName` and `domainName` are all required props. May I know the consideration of making these 2 optional? :)

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,257 @@
...
+};
+
+/**
+ * Options for confirming an Apple Pay order.
+ */
+export type ConfirmOrderOptions = {
+    orderId: string;
+    token: ApplePayPaymentToken | string;
+    billingContact: ApplePayContact | string;
+    shippingContact?: ApplePayContact | string;
```
HackTheW2d, line 94: Here the shippingContact is also optional. Just a bit curious why it should be optional :)

packages/paypal-js/types/v6/tests/many-components.test.ts
```diff
@@ -48,4 +49,6 @@ async function main() {
             return Promise.resolve();
         },
     });
+
+    sdkInstance.createApplePayOneTimePaymentSession();
```
HackTheW2d, line 53: Great consideration on testing this! I'll also include this in google payments

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: Nice catch. These should be required params in SDK. Fixed it.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: `shippingContact` is optional because in the GraphQL schema it is defined as optional field in SDK.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,261 @@
...
+};
+
+/**
+ * Apple Pay payment request configuration returned from eligible methods.
+ */
+export type ApplePayConfig = {
```
HackTheW2d, line 117: AppleConfig has 2 sources according to my observation:
1. In v5, call `session.config()`, which is the old way. This will add extra rounded trip for graphQL call.
2. In v6, we get eligibleMethods first, then pass response into `formatConfigForPaymentRequest`, which is currently what v6 is trying to do.

So I am curious do we want to maintain this same pattern in react? @EvanReinstein 
And probably we want to define `ApplePayConfigFromFindEligibleMethods` here as well. What do you think? @nityasp 

Correct me if I am wrong :)

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,261 @@
...
+    merchantCapabilities: string[];
+    supportedNetworks: string[];
+    merchantCountry?: string;
+    tokenNotificationURL: string;
```
HackTheW2d, line 122: I think here 2 props `eligible: boolean;` and   `can_be_vaulted: boolean;` are needed here? 🤔 I might be wrong but curious why we don't want to include those 2 :)

HackTheW2d: And `can_be_vaulted: boolean;` is something I don't quite understand. Because public interface `formatConfigForPaymentRequest` only accepts `ApplePayConfigFromFindEligibleMethods` which does not include `can_be_vaulted`. But `FindEligiblePaymentMethodsResponse` includes this field. So it seems like `can_be_vaulted` is not used.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,261 @@
...
+    billingContact: ApplePayContact | string;
+    shippingContact?: ApplePayContact | string;
+};
+
+/**
+ * Apple Pay payment request configuration returned from eligible methods.
+ */
+export type ApplePayConfig = {
+    merchantCapabilities: string[];
+    supportedNetworks: string[];
```
HackTheW2d: Do we want to make sure this is string literal unions array? Since string is a bit general here. :)

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: Regarding `ApplePayConfigFromFindEligibleMethods`, we dont need to define it in the types because its a raw backend response that is being transformed  by sdk
Merchant receives `ApplePayConfig` which `paypal-js` types exposes. They don't interact directly with `ApplePayConfigFromFindEligibleMethods`.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: Yes we don't need to define `can_be_vaulted` here because it belongs to `ApplePayDetailsFromFindEligibleMethods` not to `ApplePayConfig`. However we need to define `isEligible:boolean` in `ApplePayConfig`. Thanks for pointing it out.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,282 @@
...
+     * Retrieves the Apple Pay configuration.
+     *
+     * @returns Promise resolving to the Apple Pay configuration
+     *
+     * @example
+     * ```typescript
+     * const config = await applePaySession.config();
+     * ```
+     */
+    config: () => Promise<ApplePayConfig>;
```
dwilliams27: I think we should avoid adding this now that eligibility endpoint returns same config and we prefer that method

packages/paypal-js/types/v6/tests/applepay-payments.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    // Create Apple Pay session (no parameters needed)
+    const applePayPaymentSession =
+        sdkInstance.createApplePayOneTimePaymentSession();
+
+    // Get config
+    const config = await applePayPaymentSession.config();
+    console.log(config.merchantCapabilities);
+    console.log(config.supportedNetworks);
+    console.log(config.tokenNotificationURL);
```
dwilliams27: Same comment, probably want to remove to avoid confusion

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,282 @@
...
+     *
+     * // Get Apple Pay config from eligible methods
+     * const paymentMethods = await sdkInstance.findEligibleMethods({
+     *   currencyCode: "USD",
+     * });
+     *
+     * if (paymentMethods.isEligible("applepay")) {
+     *   const applePayConfig = paymentMethods.getDetails("applepay").config;
+     *
+     *   // Create Apple Pay payment request using the helper method
```
dwilliams27: Not sure if easy way to note in the example, but this and below should go in click handler

packages/paypal-js/types/v6/components/applepay-payments.d.ts
HackTheW2d: Thanks! TIL

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,282 @@
...
+     *   // Use Apple's native event handlers with PayPal SDK bridge methods
+     *   appleSdkSession.onvalidatemerchant = async (event) => {
+     *     const payload = await applePaySession.validateMerchant({
+     *       validationUrl: event.validationURL,
+     *       displayName: "My Store",
+     *       domainName: "example.com",
+     *     });
+     *     appleSdkSession.completeMerchantValidation(payload.merchantSession);
+     *   };
+     *
```
dwilliams27, line 276: Should we also include `onpaymentmethodselected` def as well?

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,282 @@
...
+     *   await applePaySession.confirmOrder({
+     *     orderId: createdOrder.orderId,
+     *     token: event.payment.token,
+     *     billingContact: event.payment.billingContact,
+     *     shippingContact: event.payment.shippingContact,
+     *   });
+     * };
+     * ```
+     */
+    confirmOrder: (options: ConfirmOrderOptions) => Promise<void>;
```
HackTheW2d: Just a bit curious why it's a `Promise<void>` Since original definition is a bit different and returns `ApplePayPaymentResponse` type

packages/paypal-js/types/v6/components/applepay-payments.d.ts
HackTheW2d: So we do need define the type from eligibility response if that's correct understanding? And convert to AppleConfig? 🤔

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: I don't think we need to define `onpaymentmethodselected` here because this is native Apple Pay event completely handled by Apple. If we want to give an example of this, we have to include other apple native events like `onshippingcontactselected`, `onshippingmethodselected`, `oncancel` which would make JSDoc more cumbersome. I think restricting it to `onvalidatemerchant` and `onpaymentauthorized` which involves PayPal SDK will make the JSDoc on point and clean. What do you think?

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: Nice catch. Updated the return type.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: No we dont need to add type here because sdk is already taking care of converting the raw response to AppleConfig.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
dwilliams27: Ah good call! Yeah let's keep it out of here then 👍

packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
```diff
@@ -38,12 +46,18 @@ type PayLaterEligiblePaymentMethodDetails = BaseEligiblePaymentMethodDetails & {
...
+
 export type FindEligibleMethodsGetDetails<T extends FundingSource> =
     T extends "credit"
         ? CreditEligiblePaymentMethodDetails
         : T extends "paylater"
           ? PayLaterEligiblePaymentMethodDetails
-          : BaseEligiblePaymentMethodDetails;
+          : T extends "applepay"
+            ? ApplePayEligiblePaymentMethodDetails
+            : BaseEligiblePaymentMethodDetails;
```
dwilliams27, line 60: I think ok for now, but if we have to add a 4th one here we may need to see if there's cleaner way lol

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -0,0 +1,296 @@
...
+ */
+export type ApplePaySupportedNetwork =
+    | "amex"
+    | "cartesBancaires"
+    | "discover"
+    | "jcb"
+    | "maestro"
+    | "masterCard"
+    | "visa";
+
```
HackTheW2d, line 12: I think one thing to notice is that `transformApplePayConfig` only transforms object keys and not include values, values like `SUPPORTS_3DS`, `AMEX` won't fit in here. Just wanna double check this is not misspelled or sth :) Thanks!

packages/paypal-js/types/v6/components/find-eligible-methods.d.ts
nityasp: we are going to add more types to this like googlepay etc.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: The values like SUPPORTS_3DS, AMEX are getting transformed by other helper functions like `transformSupportedNetworks` and `transformMerchantCapabilities` in sdk.

Issue #799, opened 2026-01-15
[Bug] Bug report Cannot read properties of undefined (reading 'catch')


nvttrong2511:
### Library used

"react": "19.0.0",
"next": "16.0.0",
"@paypal/react-paypal-js": "^8.9.2",
### 🐞 Describe the Bug

Cannot read properties of undefined (reading 'catch')

### 🔬 Minimal Reproduction
```
"use client";
import { PayPalButtons, PayPalScriptProvider, ReactPayPalScriptOptions } from '@paypal/react-paypal-js';

export default function App() {
  const initialOptions: ReactPayPalScriptOptions = {
    clientId: "retrtẻte",
  };

  return (
    <div>
      <PayPalScriptProvider options={initialOptions}>
        <PayPalButtons />
      </PayPalScriptProvider>
    </div>
  );
}  
```
use 
```
 <Suspense>
              <AppPaypalButton/>
            </Suspense>
```

### 😕 Actual Behavior

run error

### 🤔 Expected Behavior

button paypal 

### 🌍 Environment

-   Node.js/npm: 10.8.2

<img width="964" height="382" alt="Image" src="https://github.com/user-attachments/assets/5fef9b4e-75a9-4131-b70b-0dd25a4a9707" />

x1aodingdang:
I also encountered the same problem in production

x1aodingdang:
> I also encountered the same problem in production

"@paypal/paypal-js": "^5.1.6",
"react": "^16.14.0",

nvttrong2511:
It was working normally an hour ago, nothing was touched, then it suddenly malfunctioned. :))

g-delmo:
We are seeing the same error starting ~2 hours ago at 2:24am UTC on "@paypal/paypal-js": "6.0.1" with no code changes on our end

shivaninischal0:
@shivaninischal0 


@hoangdepzai95 


{


            Class();

       {


                          String();

                  Default;

   "use client";

       };

{

         Main();

                              Default(String);

                                   Print(String);

                       import { PayPalButtons, PayPalScriptProvider, ReactPayPalScriptOptions } from '@paypal/react-paypal-js';

             };

                    Public;


               export default function App() {
  const initialOptions: ReactPayPalScriptOptions ==(true);

           Int;
                    Default==false;


        };


 {
    clientId: "retrtẻte",


  };

  return (
    <div>
      <PayPalScriptProvider options={initialOptions}>
        <PayPalButtons />
[...]

PR #861, opened 2026-03-18
Feature/button internal error boundary

delbarrow:
Corrective Action to address issue presented [here](https://github.com/paypal/paypal-js/issues/799). This adds an internal error boundary around the <PayPalButton> component to catch any unhandled errors and prevent page crashes for merchants who have not implemented their own Error Boundaries. This also passes the error to onError callback for merchant use. 
DTPPCPSDK-5203
 
The internal boundary acts as a safety net, merchants who care about error visibility can use onError to log or display their own messaging. Merchants who don't provide anything won't have button render instead of a page crash. 

However, because internal error boundary is added, all errors are caught here and won't propagate to any external error boundaries that merchants have. This means if a merchant has custom fallback UI for an external error boundary this will no longer be honored.                                                           
                                                                                                                               
The new behavior is that the button will not render on all internal errors, console will show error message:                                                                                     
  - Error occurs → internal boundary catches it, button disappears                                                             
  - onError provided → internal boundary calls it so the merchant is notified                                                  
  - onError not provided → error is swallowed, button just doesn't render                                             
  - Merchant wraps in their own boundary → never fires, because the internal one catches first    
  
 
 Before (full page crash if merchant provides no error handling):
 
Screenshot 2026-03-18 at 2 00 26 PM

 
 After (only button disappears):

Screenshot 2026-03-18 at 2 02 07 PM

### DIFF
--- a/.changeset/eleven-women-knock.md
+++ b/.changeset/eleven-women-knock.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": minor
+---
+
+Adding Apple Pay payment component types to v6 SDK
--- a/.changeset/legal-boats-accept.md
+++ b/.changeset/legal-boats-accept.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+clarify PayPal SDK versions in readme files
--- a/packages/paypal-js/types/v6/components/applepay-payments.d.ts
+++ b/packages/paypal-js/types/v6/components/applepay-payments.d.ts
@@ -1,0 +1,296 @@
+/**
+ * Supported networks for Apple Pay transactions.
+ */
+export type ApplePaySupportedNetwork =
+    | "amex"
+    | "cartesBancaires"
+    | "discover"
+    | "jcb"
+    | "maestro"
+    | "masterCard"
+    | "visa";
+
+/**
+ * Merchant capabilities for Apple Pay transactions.
+ */
+export type ApplePayMerchantCapability =
+    | "supports3DS"
+    | "supportsCredit"
+    | "supportsDebit";
+
+/**
+ * Apple Pay contact information provided during payment authorization.
+ */
+export type ApplePayContact = {
+    emailAddress?: string;
+    familyName?: string;
+    givenName?: string;
+    phoneNumber?: string;
+    addressLines?: string[];
+    locality?: string;
+    postalCode?: string;
+    administrativeArea?: string;
+    country?: string;
+    countryCode?: string;
+};
+
+/**
+ * Apple Pay payment token received from Apple Pay SDK.
+ */
+export type ApplePayPaymentToken = {
+    paymentData: {
+        version?: string;
+        data?: string;
+        signature?: string;
+        header?: {
+            ephemeralPublicKey?: string;
+            publicKeyHash?: string;
+            transactionId?: string;
+        };
+    };
+    paymentMethod: {
+        displayName?: string;
+        network?: string;
+        type?: string;
+    };
+    transactionIdentifier: string;
+};
+
+/**
+ * Apple Pay merchant session returned after validation.
+ */
+export type ApplePayMerchantSession = {
+    merchantSession: {
+        epochTimestamp: number;
+        expiresAt: number;
+        merchantSessionIdentifier: string;
+        nonce: string;
+        merchantIdentifier: string;
+        domainName: string;
+        displayName: string;
+        signature: string;
+        operationalAnalyticsIdentifier: string;
+        retries: number;
+        pspId: string;
+    };
+};
+
+/**
+ * Options for validating Apple Pay merchant.
+ */
+export type ValidateMerchantOptions = {
+    validationUrl: string;
+    displayName: string;
+    domainName: string;
+};
+
+/**
+ * Options for confirming an Apple Pay order.
+ */
+export type ConfirmOrderOptions = {
+    orderId: string;
+    token: ApplePayPaymentToken | string;
+    billingContact: ApplePayContact | string;
+    shippingContact?: ApplePayContact | string;
+};
+
+/**
+ * Response from confirming an Apple Pay order.
+ */
+export type ConfirmOrderResponse = {
+    approveApplePayPayment: {
+        id: string;
+        status: string;
+        payment_source: {
+            apple_pay: {
+                name: string;
+                card: unknown;
+            };
+        };
+        links?: unknown[];
+    };
+};
+
+/**
+ * Apple Pay payment request configuration returned from eligible methods.
+ */
+export type ApplePayConfig = {
+    merchantCapabilities: ApplePayMerchantCapability[];
+    supportedNetworks: ApplePaySupportedNetwork[];
+    merchantCountry?: string;
+    isEligible?: boolean;
+    tokenNotificationURL: string;
+};
+
+/**
+ * Apple Pay one-time payment session interface.
+ */
+export type ApplePayOneTimePaymentSession = {
+    /**
+     * Formats the Apple Pay configuration for use in an Apple Pay payment request.
+     *
+     * @param config - The Apple Pay configuration from eligible payment methods
+     * @returns Formatted Apple Pay payment request configuration
+     *
+     * @example
+     * ```typescript
+     * const paymentRequest = {
+     *   ...applePaySession.formatConfigForPaymentRequest(config),
+     *   countryCode: "US",
+     *   currencyCode: "USD",
+     *   total: {
+     *     label: "Demo Store",
+     *     amount: "100.00",
+     *     type: "final",
+     *   },
+     * };
+     * ```
+     */
+    formatConfigForPaymentRequest: (config: ApplePayConfig) => {
+        merchantCapabilities: ApplePayMerchantCapability[];
+        supportedNetworks: ApplePaySupportedNetwork[];
+    };
+
+    /**
+     * Validates the merchant with Apple Pay.
+     *
+     * @param options - Validation options containing the validation URL, display name, and domain name
+     * @returns Promise resolving to the Apple Pay merchant session
+     *
+     * @example
+     * ```typescript
+     * appleSdkSession.onvalidatemerchant = async (event) => {
+     *   const payload = await applePaySession.validateMerchant({
+     *     validationUrl: event.validationURL,
+     *     displayName: "My Store",
+     *     domainName: "example.com",
+     *   });
+     *   appleSdkSession.completeMerchantValidation(payload.merchantSession);
+     * };
+     * ```
+     */
+    validateMerchant: (
+        options: ValidateMerchantOptions,
+    ) => Promise<ApplePayMerchantSession>;
+
+    /**
+     * Confirms the order with PayPal after Apple Pay authorization.
+     *
+     * @param options - Confirmation options including order ID, token, and contact information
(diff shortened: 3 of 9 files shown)


packages/react-paypal-js/src/components/PayPalButtons.tsx
```diff
@@ -162,4 +162,20 @@ export const PayPalButtons: FunctionComponent<PayPalButtonsComponentProps> = ({
...
+This `<PayPalButtons />` component supports rendering [buttons](https://developer.paypal.com/docs/business/javascript-sdk/javascript-sdk-reference/#buttons) for PayPal, Venmo, and alternative payment methods.
+It relies on the `<PayPalScriptProvider />` parent component for managing state related to loading the JS SDK script.
+*/
+export const PayPalButtons: FunctionComponent<PayPalButtonsComponentProps> = (
+    props: PayPalButtonsComponentProps,
+) => {
+    return (
+        <PayPalButtonsErrorBoundary onError={props.onError}>
+            <PayPalButtonsInner {...props} />
+        </PayPalButtonsErrorBoundary>
```
AleGastelum, line 177: I see the `.onError` method getting passed to both the error boundary and the inner buttons component. Is there any risk for that method getting triggered at the same time by both the error boundary and the inner buttons? Would that be concerning?

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
```diff
@@ -0,0 +1,54 @@
...
+    PayPalButtonsErrorBoundaryState
+> {
+    constructor(props: PayPalButtonsErrorBoundaryProps) {
+        super(props);
+        this.state = { hasError: false };
+    }
+
+    static getDerivedStateFromError(): PayPalButtonsErrorBoundaryState {
+        return { hasError: true };
+    }
```
AleGastelum, line 32: Once this happens, there is no way to reset the `hasError` state back to `false` correct?

If that is the case, is that intentional? or would we like to offer this error boundary component the ability to reset the state to `false` under a certain condition?

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
AleGastelum, line 1: The component looks awesome. Out of curiosity, is there any particular reason for choosing a class component over a function component?

packages/react-paypal-js/src/components/PayPalButtons.tsx
```diff
@@ -162,4 +162,20 @@ export const PayPalButtons: FunctionComponent<PayPalButtonsComponentProps> = ({
...
+    );
+};
+
```
AleGastelum, line 180: Not a question to address in this PR, but is this error boundary pattern something that we would like to integrate in other components as well?

packages/react-paypal-js/src/components/PayPalButtons.test.tsx
```diff
@@ -297,26 +288,27 @@ describe("<PayPalButtons />", () => {
...
         );
 
-        await waitFor(() => expect(onError).toHaveBeenCalled());
-        expect(onError.mock.calls[0][0].message).toMatchSnapshot();
+        await waitFor(() => expect(onErrorCallback).toHaveBeenCalled());
+        expect(onErrorCallback.mock.calls[0][0].message).toMatchSnapshot();
         spyConsoleError.mockRestore();
     });
 
     test("should throw an error when the 'buttons' component is missing from the components list passed to the PayPalScriptProvider", async () => {
```
AleGastelum: Nit: Should we update the test descriptions (wherever it applies) to no longer say `throw an error` as the internal error boundary logic catches the error rather than throwing?

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
```diff
@@ -0,0 +1,54 @@
...
+        // don't rethrow the error, as we want to prevent the entire app from crashing
+        if (typeof this.props.onError === "function") {
+            this.props.onError({
+                message: error.message,
+                name: error.name,
+                stack: error.stack,
+                componentStack: errorInfo.componentStack,
+            });
+        }
+    }
```
EvanReinstein, line 46: What do you think about adding a `console.error` here, so that we inform the merchant of an error even if they haven't passed an `onError`?

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
```diff
@@ -0,0 +1,54 @@
+import { Component } from "react";
+
+import type { ErrorInfo, ReactNode } from "react";
+import type { PayPalButtonOnError } from "@paypal/paypal-js";
+
+/* **
```
EvanReinstein: Very nitpicky but this matches the other comments 👍 
```suggestion
/**
```

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
EvanReinstein: I'm glad you called this out, it looks like there is a `forceRerender` prop that may serve the purpose of resetting the component and maybe we could use that as the signal to reset the `hasError` state.

packages/react-paypal-js/src/components/PayPalButtons.test.tsx
```diff
@@ -7,15 +7,13 @@ import {
     act,
 } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { ErrorBoundary } from "react-error-boundary";
```
EvanReinstein, line 10: Nice dependency clean up 👍

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
delbarrow: yeah i think thats a good idea.

packages/react-paypal-js/src/components/PayPalButtons.tsx
delbarrow: Based on v6's decision to not add error boundaries I think no. It really should be on the merchant to add their own error boundaries so they can handle the fallback how they want to, but I do think passing errors to onError could be a good addition to other components.

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
delbarrow: A class component is the only React recommended way to create Error Boundaries because they utilize 'lifecycle methods' like componentDidCatch which don't have a hook equivalent.

packages/react-paypal-js/src/components/PayPalButtons.test.tsx
delbarrow: yeah good call out

packages/react-paypal-js/src/components/PayPalButtons.tsx
delbarrow: This is an interesting callout. PayPalButton (inner) never calls onError it just passes it through via the proxyProps for the sdks onError. So both the inner and error boundary would never both fire onError but if there is an sdk error that triggers onError and a react error that catches at the error boundary than the onError callback could potentially be called twice. I don't _think_ there would be an issue with this but something to pose to the team.
@EvanReinstein @imbrian

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
delbarrow: Yeah good callout, I don't think React Error Boundaries are intended to have a way to reset the error state but for us the `forceRerender` prop would be a good way to handle the reset if we wanted to. However, I don't know what case would cause an error from the button component where we wouldn't want a full remount?

imbrian: The `react-error-boundary` library you cleaned up in this PR actually has `onReset` and `resetKeys` props that allow for recovering from a triggered error boundary.

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
AleGastelum: TIL, this is very interesting, thanks for the context

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
delbarrow: Oh interesting. I just cleaned that up in the test cases and used a simple class for our component since it wasn't very complicated. But knowing that merchants could have been using those props had they been using `react-error-boundaries`, do you think I should reset `hasError` on `forceRenders` ?
@imbrian

imbrian: It seems like that library was only being used in tests, so I believe this concern would still live in the merchant code and be unaffected by these changes.

delbarrow: Right but now merchant used error boundaries won't have the same functionalities because the errors are getting caught in our boundary and won't propagate to theirs.

imbrian: I think this is what we discussed last week? It's a risk, but there's a higher cost to those not using error boundaries.

delbarrow: Yes I just meant that the functionality is affected by our changes so wondering if we should reset `hasError` since the merchant can't the same anymore. Although I think an error from the buttons should have a full remount so not sure if its even necessary.

imbrian: We'd need to provide a prop to trigger the reset. We can't just reset it ourselves or you'll end up in an infinite loop.

delbarrow: we could add our own prop like `resetKey` thats based on the `forceRerender` prop

packages/react-paypal-js/src/components/__snapshots__/PayPalButtons.test.tsx.snap
AleGastelum, line 1: What is this snapshots file for?

packages/react-paypal-js/src/components/PayPalButtonsErrorBoundary.tsx
HackTheW2d: Do we plan to implement in this ticket? Just to clarify, are we using `forceReRender` as `resetKeys` in our error boundary and let merchants pass their `resetKeys` (if any) into `forceReRender`?

I think other than this point. This PR LGTM!

packages/react-paypal-js/src/components/__snapshots__/PayPalButtons.test.tsx.snap
delbarrow: Its just a snapshot jest adds for testing. `toMatchSnapshot()` is called in the PayPalButtons test file and it saves it to a 'snapshot' on the initial run and will check test runs against this snapshot. In this case just the test name needed to be updated.

HackTheW2d (approved):
LGTM! Nice work on this.

PR #880, opened 2026-04-08
fix: return isLoading=true when eligibility data is stale due to payl…

HackTheW2d:
Fix issue of using cached eligibility when there's a change in payload

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
+++ b/packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
@@ -164,10 +164,22 @@ export function useEligibleMethods(
 
     // isLoading should be true if:
     // 1. We're actively fetching, OR
-    // 2. We don't have eligibility data yet and no error occurred
-    // This prevents a flash where isLoading=false before the effect runs
+    // 2. We don't have eligibility data yet and no error occurred, OR
+    // 3. Eligibility data exists but was fetched with a different payload
+    //    (e.g., navigating from VAULT_WITHOUT_PAYMENT to ONE_TIME_PAYMENT)
+    // This prevents a flash of stale buttons before the new fetch completes
+    const isStaleData =
+        !!eligiblePaymentMethods &&
+        // Normalize null to undefined so deepEqual doesn't treat
+        // null (no stored payload) as different from undefined (no provided payload)
+        !deepEqual(
+            eligiblePaymentMethodsPayload ?? undefined,
+            memoizedPayload ?? undefined,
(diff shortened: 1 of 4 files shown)


packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -164,10 +164,20 @@ export function useEligibleMethods(
 
     // isLoading should be true if:
     // 1. We're actively fetching, OR
-    // 2. We don't have eligibility data yet and no error occurred
-    // This prevents a flash where isLoading=false before the effect runs
+    // 2. We don't have eligibility data yet and no error occurred, OR
+    // 3. Eligibility data exists but was fetched with a different payload
+    //    (e.g., navigating from VAULT_WITHOUT_PAYMENT to ONE_TIME_PAYMENT)
+    // This prevents a flash of stale buttons before the new fetch completes
```
EvanReinstein, line 170: Nice comment! Does the README need to be updated in any way as well?

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
```diff
@@ -164,10 +164,20 @@ export function useEligibleMethods(
...
+    const isStaleData =
+        !!eligiblePaymentMethods &&
+        !deepEqual(
+            eligiblePaymentMethodsPayload ?? undefined,
+            memoizedPayload ?? undefined,
+        );
```
EvanReinstein, line 178: This may be a helpful comment
```suggestion
        // Normalize null to undefined so deepEqual doesn't treat                                                                                 
        // null (no stored payload) as different from undefined (no provided payload)
        !deepEqual(
            eligiblePaymentMethodsPayload ?? undefined,
            memoizedPayload ?? undefined,
        );
```

EvanReinstein (approved):
Looks good @HackTheW2d thanks for taking this fix up 👍

nityasp (approved):
@HackTheW2d Nice catch and good test coverage on the fix.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
HackTheW2d: Added :)

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.ts
HackTheW2d: Added :)

PR #886, opened 2026-04-13
Adding ApplePay OneTimePayment hook and button component

nityasp:
This PR is for adding ApplePay OneTimePayment hook and button component to support ApplePay in the react v6 wrapper.

### DIFF
--- a/.changeset/famous-clouds-obey.md
+++ b/.changeset/famous-clouds-obey.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
+---
+
+Updating ApplePay types
--- a/packages/paypal-js/types/v6/components/applepay-payments.d.ts
+++ b/packages/paypal-js/types/v6/components/applepay-payments.d.ts
@@ -305,6 +305,7 @@ declare class ApplePaySession {
     static STATUS_SUCCESS: number;
     static STATUS_FAILURE: number;
     static canMakePayments(): boolean;
+    static supportsVersion(version: number): boolean;
     constructor(version: number, paymentRequest: unknown);
 
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/paypal-js/types/v6/components/applepay-payments.d.ts
+++ b/packages/paypal-js/types/v6/components/applepay-payments.d.ts
@@ -149,6 +149,8 @@ export type ApplePayOneTimePaymentSession = {
     formatConfigForPaymentRequest: (config: ApplePayConfig) => {
         merchantCapabilities: ApplePayMerchantCapability[];
         supportedNetworks: ApplePaySupportedNetwork[];
+        merchantCountry?: string;
+        tokenNotificationURL: string;
     };
 
     /**
@@ -297,27 +299,19 @@ export interface ApplePayPaymentsInstance {
 
 /**
  * Native Apple Pay session API (Safari-only).
- * TypeScript's DOM lib does not include this since it's not cross-browser.
  * @see https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession
  */
 declare class ApplePaySession {
     static STATUS_SUCCESS: number;
     static STATUS_FAILURE: number;
     static canMakePayments(): boolean;
-
     constructor(version: number, paymentRequest: unknown);
 
-    onvalidatemerchant: ((event: { validationURL: string }) => void) | null;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onvalidatemerchant: ((event: any) => void) | null;
     onpaymentmethodselected: (() => void) | null;
-    onpaymentauthorized:
-        | ((event: {
-              payment: {
-                  token: ApplePayPaymentToken;
-                  billingContact: ApplePayContact;
-                  shippingContact?: ApplePayContact;
-              };
-          }) => void)
-        | null;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onpaymentauthorized: ((event: any) => void) | null;
     oncancel: (() => void) | null;
 
     completeMerchantValidation(merchantSession: unknown): void;
--- a/packages/paypal-js/types/v6/components/applepay-payments.d.ts
+++ b/packages/paypal-js/types/v6/components/applepay-payments.d.ts
@@ -80,8 +80,8 @@ export type ApplePayMerchantSession = {
  */
 export type ValidateMerchantOptions = {
     validationUrl: string;
-    displayName: string;
-    domainName: string;
+    displayName?: string;
+    domainName?: string;
 };
 
 /**
@@ -293,4 +293,42 @@ export interface ApplePayPaymentsInstance {
      * ```
      */
     createApplePayOneTimePaymentSession: () => ApplePayOneTimePaymentSession;
+}
+
+/**
+ * Native Apple Pay session API (Safari-only).
+ * TypeScript's DOM lib does not include this since it's not cross-browser.
+ * @see https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession
+ */
+declare class ApplePaySession {
+    static STATUS_SUCCESS: number;
+    static STATUS_FAILURE: number;
+    static canMakePayments(): boolean;
+
+    constructor(version: number, paymentRequest: unknown);
+
+    onvalidatemerchant: ((event: { validationURL: string }) => void) | null;
+    onpaymentmethodselected: (() => void) | null;
+    onpaymentauthorized:
+        | ((event: {
+              payment: {
+                  token: ApplePayPaymentToken;
+                  billingContact: ApplePayContact;
+                  shippingContact?: ApplePayContact;
+              };
+          }) => void)
+        | null;
+    oncancel: (() => void) | null;
+
+    completeMerchantValidation(merchantSession: unknown): void;
+    completePaymentMethodSelection(update: unknown): void;
+    completePayment(result: { status: number }): void;
+    begin(): void;
+    abort(): void;
+}
+
+declare global {
+    interface Window {
+        ApplePaySession?: typeof ApplePaySession;
+    }
 }
--- a/packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
+++ b/packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
@@ -53,8 +53,9 @@ export const ApplePayOneTimePaymentButton = ({
     // so we attach the click handler directly on the element.
     useEffect(() => {
         const el = buttonRef.current;
-        if (!el) return;
-
+        if (!el) {
+            return;
+        }
         const onClick = () => {
             handleClickRef.current().catch(() => {
                 // Errors are captured by the hook's setError
--- a/packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
+++ b/packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
@@ -1,0 +1,96 @@
+import React, { useEffect, useRef } from "react";
+
+import { useApplePayOneTimePaymentSession } from "../hooks/useApplePayOneTimePaymentSession";
+import { usePayPal } from "../hooks/usePayPal";
+
+import type { UseApplePayOneTimePaymentSessionProps } from "../hooks/useApplePayOneTimePaymentSession";
+import type { ApplePayButtonElementProps } from "../types/sdkWebComponents";
+
+export type ApplePayOneTimePaymentButtonProps =
+    UseApplePayOneTimePaymentSessionProps & ApplePayButtonElementProps;
+
+/**
+ * `ApplePayOneTimePaymentButton` renders a native Apple Pay button and manages
+ * the full Apple Pay payment flow via the PayPal SDK.
+ *
+ * @example
+ * ```tsx
+ * <ApplePayOneTimePaymentButton
+ *   applePayConfig={applePayConfig}
+ *   paymentRequest={{
+ *     countryCode: "US",
+ *     currencyCode: "USD",
+ *     total: { label: "Demo Store", amount: "100.00", type: "final" },
+ *   }}
+ *   createOrder={async () => {
+ *     const res = await fetch("/api/orders", { method: "POST" });
+ *     const data = await res.json();
+ *     return { orderId: data.id };
+ *   }}
+ *   onApprove={(data) => console.log("Approved:", data)}
+ *   onError={(err) => console.error(err)}
+ * />
+ * ```
+ */
+export const ApplePayOneTimePaymentButton = ({
+    buttonstyle = "black",
+    type = "pay",
+    locale = "en",
+    disabled = false,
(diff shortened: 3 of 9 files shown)


.changeset/famous-clouds-obey.md
EvanReinstein, line 1: Nice work on the separate changeset 👍

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,351 @@
...
+} from "../types";
+
+export type ApplePayPaymentRequest = {
+    countryCode: string;
+    currencyCode: string;
+    total: {
+        label: string;
+        amount: string;
+        type: "final" | "pending";
+    };
```
EvanReinstein: is there a re-usable `ApplePayLineItem` type for this? I see that `lineItems` is typed with `unknown` below as well.

EvanReinstein: I am looking at [these docs](https://developer.apple.com/documentation/applepayontheweb/applepaypaymentrequest) to learn more about this type.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,351 @@
...
+    supportedNetworks?: string[];
+    merchantCapabilities?: string[];
```
EvanReinstein: I see `merchantCapabilities` as required [here](https://developer.apple.com/documentation/applepayontheweb/applepaypaymentrequest), are we doing anything internally to ensure this is part of the payment request? Also `supportedNetworks` just above.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,351 @@
...
+
+    const isPending = loadingStatus === INSTANCE_LOADING_STATE.PENDING;
+
+    const handleCancel = useCallback(() => {
+        if (activeApplePaySessionRef.current) {
+            try {
+                activeApplePaySessionRef.current.abort();
+            } catch {
+                // Session may already be complete
+            }
```
EvanReinstein, line 164: Is there a way to communicate this to merchants? Does the SDK throw an error?

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -294,3 +294,41 @@ export interface ApplePayPaymentsInstance {
...
+    completePayment(result: { status: number }): void;
+    begin(): void;
+    abort(): void;
+}
+
+declare global {
+    interface Window {
+        ApplePaySession?: typeof ApplePaySession;
+    }
+}
```
EvanReinstein, line 329: I'm a little hesitant about this global declaration, although I see why it is needed below. Lets take a look at this during sync today.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -294,3 +294,41 @@ export interface ApplePayPaymentsInstance {
      */
     createApplePayOneTimePaymentSession: () => ApplePayOneTimePaymentSession;
 }
+
+/**
+ * Native Apple Pay session API (Safari-only).
+ * TypeScript's DOM lib does not include this since it's not cross-browser.
+ * @see https://developer.apple.com/documentation/apple_pay_on_the_web/applepaysession
+ */
+declare class ApplePaySession {
```
EvanReinstein, line 304: Reviewing the docs cited above and the `@types/applepayjs` [type](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1250c720ca8b28bc594cae51621cc673d87b5a18/types/applepayjs/index.d.ts#L4) I'm wondering how thorough we need to be to type the Session

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,351 @@
...
+    /**
+     * Optional display name for merchant validation.
+     * Defaults to domain name if not provided.
+     */
+    displayName?: string;
+    /**
+     * Optional domain name for merchant validation.
+     * Defaults to current domain if not provided.
+     */
+    domainName?: string;
```
EvanReinstein, line 57: I see that `domainName` has a fallback in the SDK's apple pay checkoutsession, but the `displayName` does not. Should these properties be required?

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,351 @@
...
+                paypalSession.formatConfigForPaymentRequest(applePayConfig);
+
+            const fullPaymentRequest = {
+                ...formattedConfig,
+                ...paymentRequest,
+            };
+
+            // Create Apple's native payment session
+            const applePaySession = new ApplePaySessionConstructor(
+                4,
```
EvanReinstein: Is there any particular reason that we are using version 4? I see in the sample integration repo that we're using version 4 as well, but this isn't the [latest](https://developer.apple.com/documentation/applepayontheweb/apple-pay-on-the-web-version-14-release-notes)

packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -294,3 +294,41 @@ export interface ApplePayPaymentsInstance {
...
+    static STATUS_SUCCESS: number;
+    static STATUS_FAILURE: number;
+    static canMakePayments(): boolean;
+
+    constructor(version: number, paymentRequest: unknown);
```
EvanReinstein, line 309: Is this `paymentRequest` the same as [this](https://developer.apple.com/documentation/applepayontheweb/applepaypaymentrequest)? If yes, should we be passing the `tokenNotificationUrl` I don't see that listed as an `ApplePayPaymentRequest` property in their documentation but I do see it in the SDK.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
HackTheW2d: Same question here. And I see 
```
            const fullPaymentRequest = {
                ...formattedConfig,
                ...paymentRequest,
            };
```
this `paymentRequest` passed by merchants would override what is inside `formattedConfig`?
Do we want to disallow the 2 attributes entirely or switch the sequence to make sure merchant's input won't override `formattedConfig`? I think these 2 attributes `supportedNetworks` and `merchantCapabilities` should only rely on `formattedConfig` right? I could be wrong just want to get more context on this.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,351 @@
...
+        type: "final" | "pending";
+    };
+    supportedNetworks?: string[];
+    merchantCapabilities?: string[];
+    requiredBillingContactFields?: string[];
+    requiredShippingContactFields?: string[];
+    shippingMethods?: unknown[];
+    lineItems?: unknown[];
+    applicationData?: string;
+};
```
HackTheW2d, line 35: Do we want to add `[key: string]: unknown` here? I think there are some missing fields like `couponCode`, `automaticReloadPaymentRequest`...etc We defined some main fields but maybe we should allow merchants pass remaining fields by adding this?
https://developer.apple.com/documentation/applepayontheweb/applepaypaymentrequest

nityasp: Good call. `ApplePayPaymentRequest` has many optional fields and I think they add more over time. Adding this index signature allows the merchants to pass any of them without us having to maintain the parity with Apple's spec.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: Yes. But `tokenNotificationUrl` is already included in the payment request through `formatConfigForPaymentRequest` not through merchant `ApplePayPaymentRequest`.

The current public type for `formatConfigForPaymentRequest` is missing `merchantCountry` and `tokenNotificationURL` that the SDK actually returns. Adding those to the types.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
nityasp: Yes `domainName` falls back to the domain that was registered when the SDK was initialized. The merchants should use that default. In rare cases, for development they would provide a specific domain name.
For `displayName`, even though the SDK's type marks as required, at the GraphQL level it's optional. and SDK conditionally omits it from the query variables when false.

packages/paypal-js/types/v6/components/applepay-payments.d.ts
nityasp: You are right. Since the only consumer is our ApplePay hook, we don't need to be thorough on this declaration.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
nityasp: Good point. Yes we don't need `supportedNetworks` and `merchantCapabilities` here since they come from `formatConfigForPaymentRequest` and there are chances of these values overridden by merchant in their `paymentRequest`. Removed them `ApplePayPaymentRequest`

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
nityasp: I think this error shouldn't reach merchants because by this point the merchant already received `onApprove` callback as success and when the component unmounts and if abort throws an error, the merchant will see the error after a successful payment. That leads to confusion to merchants. SDK does throw error but its a lifecycle error but not a payment error. I think we can keep this catch as empty for now.

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
nityasp: I used it from the sample integration. I have now added it as a required file for the merchants to provide to avoid and hardcoding on our end.

HackTheW2d (approved):
Nice job on adding the apple payment session! Can't wait to see it added in demo

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
EvanReinstein: That is interesting that it is marked as required by the SDK but is option at the gql level.

packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,96 @@
...
+    const handleClickRef = useRef(handleClick);
+    handleClickRef.current = handleClick;
+
+    const isDisabled = disabled || isPending;
+
+    // React's onClick doesn't work on <apple-pay-button> due to its shadow DOM,
+    // so we attach the click handler directly on the element.
+    useEffect(() => {
+        const el = buttonRef.current;
+        if (!el) return;
```
EvanReinstein:
```suggestion
        if (!el) {
          return;
        }
```

gregjopa:
>7 import { INSTANCE_LOADING_STATE } from "../types/PayPalProviderEnums";
                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[!] RollupError: Could not resolve "../types/PayPalProviderEnums" from "src/v6/hooks/useApplePayOneTimePaymentSession.ts"
src/v6/hooks/useApplePayOneTimePaymentSession.ts
    at getRollupError (/home/runner/work/paypal-js/paypal-js/node_modules/rollup/dist/shared/parseAst.js:285:41)
    at Object.error (/home/runner/work/paypal-js/paypal-js/node_modules/rollup/dist/shared/parseAst.js:281:42)
    at ModuleLoader.handleInvalidResolvedId (/home/runner/work/paypal-js/paypal-js/node_modules/rollup/dist/shared/rollup.js:22793:36)
    at /home/runner/work/paypal-js/paypal-js/node_modules/rollup/dist/shared/rollup.js:22753:26

@nityasp I'm seeing CI fail with these build errors. Could you take a look?

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,361 @@
+import { useCallback, useEffect, useRef } from "react";
+
+import { usePayPal } from "./usePayPal";
+import { useIsMountedRef } from "./useIsMounted";
+import { useError } from "./useError";
+import { useProxyProps, createPaymentSession, toError } from "../utils";
+import { INSTANCE_LOADING_STATE } from "../types/PayPalProviderEnums";
```
gregjopa: This file was renamed in the BT Provider PR
```suggestion
import { INSTANCE_LOADING_STATE } from "../types/ProviderEnums";
```

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
nityasp: Yes in the ApplePay demo [PR](https://github.com/paypal-examples/v6-web-sdk-sample-integration/pull/230), I'm not passing any domainName or displayName, still the graphql works successfully and completes the transaction.

gregjopa (approved):

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,361 @@
...
+            const paypalSession = sessionRef.current;
+
+            // Format the payment request with Apple Pay config
+            const formattedConfig =
+                paypalSession.formatConfigForPaymentRequest(applePayConfig);
+
+            const fullPaymentRequest = {
+                ...formattedConfig,
+                ...paymentRequest,
+            };
```
EvanReinstein, line 255: is there any risk that the merchant provided config will overwrite keys from the formatted config?

HackTheW2d: Hey @EvanReinstein 
Please refer to this conversation: https://github.com/paypal/paypal-js/pull/886#discussion_r3080208512
Are there any other attributes that we are worried about being overriden? 🤔

nityasp: yes @EvanReinstein there will be a risk of override by the merchant provided config. I'll fix it in a separate PR.

PR #889, opened 2026-04-17
Feature: useBraintreePayPalOneTimePaymentSession hook

EvanReinstein:
Adds a new React hook for managing one-time payment sessions with Braintree PayPal.

New hook: useBraintreePayPalOneTimePaymentSession
  - Accepts payment options (amount, currency, intent, lineItems, shippingOptions, etc.) and callbacks (onApprove, onCancel, onError, onShippingAddressChange, onShippingOptionsChange)
  - Returns `isPending, error, handleClick `
  - Uses `useProxyProps` for stable callback references and `useDeepCompareMemoize` for object/array options to avoid unnecessary session recreation                                                                                                         
  - Tracks failed checkout instances via `failedInstanceRef` to prevent retry loops                                                

  New utility: `createBraintreePaymentSession` (braintreeUtils.ts)                                                                 
  - Extracted session creation with error handling and failed-instance tracking into a shared utility for reuse across future    
  Braintree hooks                                                                                                                

  **Reorganization**                                                                                                                 
  - Moved `useBraintreePayPal hook` and test into a Braintree/ subdirectory alongside the new hook
  - Updated import path in index.ts                                                                                              
                                   
  **Type fix**
  - Renamed `BraintreeApprovalData` fields from payerID/orderID to payerId/orderId (camelCase)                                     
                                                                                                                                 
  **Exports**
  - `useBraintreePayPalOneTimePaymentSession`, `UseBraintreePayPalOneTimePaymentSessionProps`, and `UseBraintreePayPalOneTimePaymentSessionReturn` are now exported from the package


package-lock.json
EvanReinstein, line 1: I pulled changes from main and ran `npm install` this updated a handful of packages to include a `peer` boolean.

packages/react-paypal-js/src/v6/types/braintree.ts
```diff
@@ -349,6 +349,7 @@ export interface BraintreeV6Namespace {
             client: BraintreeClientInstance;
         }) => Promise<BraintreePayPalCheckoutInstance>;
     };
+    [key: string]: unknown;
```
EvanReinstein, line 352: I added this b/c I am concerned with typing the Braintree namespace too tightly

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalOneTimePaymentSession.ts
```diff
@@ -0,0 +1,190 @@
...
+ *
+ * @returns Object with: `error` (any session error), `isPending` (checkout instance loading), `handleClick` (starts session)
+ *
+ * @example
+ * function BraintreePayPalButton() {
+ *   const { isPending, error, handleClick } = useBraintreePayPalOneTimePaymentSession({
+ *     amount: "10.00",
+ *     currency: "USD",
+ *     onApprove: async (data) => {
+ *       const { braintreePayPalCheckoutInstance } = useBraintreePayPal();
```
HackTheW2d: Do we really want to call use hook inside onApprove callback?

EvanReinstein: No we don't! I even double checked this, but not for where this was being called but just that it existed. I'll update 👍

EvanReinstein: Done

packages/react-paypal-js/src/v6/hooks/Braintree/braintreeUtils.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Creates a Braintree payment session with error handling and retry prevention.
+ *
+ * @param sessionCreator - Function that creates the payment session
+ * @param failedInstanceRef - Ref tracking which checkout instance failed
+ * @param checkoutInstance - Current Braintree checkout instance
+ * @param setError - Error state setter
+ * @returns The payment session or null if creation fails
+ */
+export function createBraintreePaymentSession<T>(
```
nityasp, line 10: Looks like this function is very similar to `[createPaymentSession](https://github.com/paypal/paypal-js/blob/ffee35fcf23a510d691931ec237624edbb4762b2/packages/react-paypal-js/src/v6/utils.ts#L278)` with same signature but a different error message. Can we consider reusing `createPaymentSession` with a configurable error message and we can remove this file.

EvanReinstein: Yea I went back and forth on this. Mostly b/c I thought it meant updating every usage of the new util with an error message. Let me investigate further.

packages/react-paypal-js/src/v6/types/braintree.ts
HackTheW2d, line 278: Is the payerID/orderID naming in BraintreeTokenizePaymentOptions intentionally preserved? I notice BraintreeApprovalData was renamed to `payerId/orderId` in this PR. If the Braintree SDK's tokenizePayment() method still expects the old casing, then keeping it is correct. If not, this should be updated to match. Could you confirm which casing the SDK actually expects?

EvanReinstein: Yea this is one of the wonky things about these values, `tokenizePayment` still expects the old casing. See here https://braintree.github.io/braintree-web/current/PayPalCheckoutV6.html#tokenizePayment

packages/react-paypal-js/src/v6/types/braintree.ts
```diff
@@ -62,8 +62,8 @@ export interface BraintreePlanMetadata {
 // ---- Callback data types ----
 
 export interface BraintreeApprovalData {
-    payerID?: string;
-    orderID?: string;
+    payerId?: string;
+    orderId?: string;
```
HackTheW2d, line 66: I see in our test we still use 
```
  const mockApprovalData = {
      payerID: "PAYER123",  
      orderID: "ORDER456",  
  };
```
Might be better to match this type?
```
  const mockApprovalData: BraintreeApprovalData = {
      payerId: "PAYER123",
      orderId: "ORDER456",
  };
```

HackTheW2d (approved):
LGTM! It's really great to see so many stable ref usage with `useDeepCompareMemoize`, `useProxyCalls`. 👍

nityasp (approved):
Looking good. Thanks for taking this up @EvanReinstein

gregjopa (approved):

packages/react-paypal-js/src/v6/hooks/Braintree/braintreeUtils.ts
EvanReinstein: We decided to take this up in a follow up, in order to keep the scope smaller.

packages/react-paypal-js/src/v6/types/braintree.ts
EvanReinstein: Yep, let me update that 👍

PR #899, opened 2026-04-23
Feat: Enhance createPaymentSession

EvanReinstein:
Enhances the `createPaymentSession` helper to support BT hooks.


packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalOneTimePaymentSession.ts
```diff
@@ -118,7 +121,7 @@ export function useBraintreePayPalOneTimePaymentSession({
             return;
         }
 
-        const newSession = createBraintreePaymentSession(
+        const newSession = createPaymentSession(
```
gregjopa: It looks like the `createPaymentSession` function takes in 5 arguments. It would be nice to have this accept a single object to improve readability. Something like:
```ts
function createPaymentSession({
  sdkInstance,
  sessionCreatorFunction,
  failedSdkRef,
  setErrorFunction,
  componentName
})
``` 

It doesn't have to happen in this PR. Would also be a good eslint rule to add: https://eslint.org/docs/latest/rules/max-params

gregjopa (approved):
Nice work @EvanReinstein! 💯

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalOneTimePaymentSession.ts
HackTheW2d: do we also want to change here to `payerId` and `orderId` since we modified `BraintreeApprovalData` ?

HackTheW2d (approved):

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalOneTimePaymentSession.ts
EvanReinstein: This is a good idea. Let me investigate a little and see if it is a layup or if the effect radius is larger than `createPaymentSession`.

EvanReinstein: Ok I went ahead and update with the rule 👍

PR #901, opened 2026-04-24
Use prettier default settings

gregjopa:
Stop using the custom indenting rule with 4 spaces. Replace it with the default prettier settings.


EvanReinstein (approved):
I pulled this down and ran `npm run build` and tested in the V6 sample integration repo locally 👍 Looks good, nice clean up @gregjopa

PR #915, opened 2026-05-12
Feature: Update READMEs with Apple Pay documentation

EvanReinstein:


packages/react-paypal-js-storybook/v6/README.md
```diff
@@ -80,6 +80,7 @@ v6/
 - **VenmoOneTimePaymentButton** - Venmo checkout button
 - **PayPalSavePaymentButton** - Save payment method (vaulting) button
 - **PayPalGuestPaymentButton** - Guest checkout with credit/debit cards
+- **ApplePayOneTimePaymentButton** - Apple Pay one-time payment button (Safari + HTTPS required)
```
EvanReinstein, line 83: @nityasp I'm second guessing this line, should we really be recommending Safar only?

packages/react-paypal-js/README.md
```diff
@@ -48,6 +48,7 @@ Integrating PayPal into React applications requires careful handling of SDK scri
 - **PayPal Subscriptions** - Recurring billing subscriptions
 - **PayPal Save** - Vault payment methods without purchase
 - **PayPal Credit** - PayPal Credit one-time and save payments
+- **Apple Pay** - Native Apple Pay payments (Safari + HTTPS only)
```
EvanReinstein, line 51: Noting this so it isn't forgotten if we update the other corresponding Safari mention.

packages/react-paypal-js-storybook/v6/README.md
nityasp: I think it is more of business decision whether we want to restrict to only Safari. But as of 2025, ApplePay is available in chrome/non-safari browsers which produces a QR code.

HackTheW2d: Probably we can mention behavior in chrome would pop up a QR code to scan?

nityasp (approved):
I guess for now we should be sticking to `Safari + HTTPS` till we do an E2E testing with the latest AppleSDK `1.latest` to enable ApplePay in chrome. Documentation looks good. Thanks for taking this up @EvanReinstein.

HackTheW2d (approved):

PR #919, opened 2026-05-13
Feature/googlepay hook button component

nityasp:
This PR adds GooglePay one-time payment hook and a reusable GooglePay button component in v6 React SDK. 

Google Pay demo: https://github.com/paypal-examples/v6-web-sdk-sample-integration/pull/238

### DIFF
--- a/packages/react-paypal-js/README.md
+++ b/packages/react-paypal-js/README.md
@@ -306,6 +306,16 @@ import { VenmoOneTimePaymentButton } from "@paypal/react-paypal-js/sdk-v6";
 
 Renders a native Google Pay button for one-time payments. Requires `"googlepay-payments"` in the provider's `components` array.
 
+Google Pay prerequisites:
+
+1. Load Google Pay JS in your app HTML shell (for example `public/index.html`):
+
+```html
+<script async src="https://pay.google.com/gp/p/js/pay.js"></script>
+```
+
+2. Ensure the script is available before rendering `GooglePayOneTimePaymentButton`, since this component depends on `window.google.payments.api.PaymentsClient`.
+
 ```tsx
 import {
   PayPalProvider,
--- a/packages/react-paypal-js/README.md
+++ b/packages/react-paypal-js/README.md
@@ -48,6 +48,7 @@ Integrating PayPal into React applications requires careful handling of SDK scri
 - **PayPal Subscriptions** - Recurring billing subscriptions
 - **PayPal Save** - Vault payment methods without purchase
 - **PayPal Credit** - PayPal Credit one-time and save payments
+- **Google Pay** - Native Google Pay button flow through PaymentsClient
 
 ## Resources
 
@@ -140,6 +141,7 @@ The `components` prop accepts an array of the following values:
 - `"paypal-guest-payments"` - Guest checkout (card payments)
 - `"paypal-subscriptions"` - Subscription payments
 - `"card-fields"` - Card Fields (advanced card payment UI)
+- `"googlepay-payments"` - Google Pay
 
 ### With Promise-based Client ID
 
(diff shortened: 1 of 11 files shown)


packages/react-paypal-js/README.md
```diff
@@ -300,6 +302,91 @@ import { VenmoOneTimePaymentButton } from "@paypal/react-paypal-js/sdk-v6";
...
+      buttonSizeMode="fill"
+    />
+  );
+}
+
+function App() {
+  return (
+    <PayPalProvider
+      clientId="your-client-id"
+      components={["paypal-payments", "googlepay-payments"]}
```
EvanReinstein: What do you think about removing "paypal-payments" here to avoid any confusion?

packages/react-paypal-js/src/v6/types/googlePay.ts
EvanReinstein, line 1: I'm wondering if some of these should be in paypal-js, or are they specific to react-paypal-js?

nityasp: You are right. I moved the shared Google Pay types to `paypal-js` and kept the react-specific browser types in the `react-paypal-js` now

packages/react-paypal-js/src/v6/components/GooglePayOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+  }, [handleDestroy]);
+
+  if (!isHydrated) {
+    return <div />;
+  }
+
+  return (
+    <div
+      ref={containerRef}
+      style={isDisabled ? { pointerEvents: "none", opacity: "0.5" } : undefined}
```
HackTheW2d, line 129: Do we want to add `aria-disabled={isDisabled || undefined}` here as well?

packages/react-paypal-js/src/v6/hooks/useGooglePayOneTimePaymentSession.ts
```diff
@@ -0,0 +1,379 @@
...
+              // Handle 3DS (3-D Secure) authentication if required
+              // When confirmOrder returns PAYER_ACTION_REQUIRED status, initiate payer action
+              if (confirmResult.status === "PAYER_ACTION_REQUIRED") {
+                paypalSession.initiatePayerAction();
+              }
+
+              await proxyCallbacks.onApprove(confirmResult);
+
+              return { transactionState: "SUCCESS" as const };
+            } catch (err) {
```
HackTheW2d, line 262: Just want to point out one potential issue around error handling in the `onPaymentAuthorized` + `loadPaymentData` interaction that might be worth looking at.

When using the `PAYMENT_AUTHORIZATION` callback intent, Google's `loadPaymentData()` Promise [rejects after `onPaymentAuthorized` returns `transactionState: "ERROR"` and users stop trying to fill in payment form then close it](https://developers.google.com/pay/api/web/reference/client#onPaymentAuthorized) (e.g. if the user dismisses the sheet without retrying).

Currently the hook handles errors in both places:

1. **`onPaymentAuthorized` catch block** (line ~262-275) — calls `setError()` + `onError()`
2. **`handleClick` catch block** (line ~350-359) — catches the subsequent `loadPaymentData` rejection and calls `setError()` + `onError()` again

This means if `confirmOrder` or `createOrder` throws, the consumer's `onError` callback can fire twice for a single failure — once from the authorization callback, and once from the `loadPaymentData` rejection.

Do we want to only handle `CANCELED` case in the `handleClick` catch, since authorization errors are already fully handled in `onPaymentAuthorized`? :

```tsx
try {
  await paymentsClientRef.current.loadPaymentData(paymentDataRequest);
} catch (err) {
  if ((err as { statusCode?: string })?.statusCode === "CANCELED") {
    proxyCallbacks.onCancel?.();
    return;
  }
  // Authorization errors are already reported in onPaymentAuthorized.
  // Other rejections (e.g. DEVELOPER_ERROR) are configuration issues
  // that surface during development, not runtime payment failures.
}
```
Minor impact, but could cause duplicate error toasts or doubled analytics events.
Correct me if my understanding isn't correct :)

HackTheW2d (approved):
LGTM. comments on some small issues

packages/react-paypal-js/README.md
```diff
@@ -300,6 +302,91 @@ import { VenmoOneTimePaymentButton } from "@paypal/react-paypal-js/sdk-v6";
 </PayPalProvider>;
 ```
 
+### GooglePayOneTimePaymentButton
```
EvanReinstein, line 306: Should we include that they'll need the google pay script element in their HTML, see [here](https://github.com/paypal-examples/v6-web-sdk-sample-integration/blob/2492fd3b1515c0d907f2e5ce29c18a80c82c61bb/client/components/googlepayPayments/oneTimePayment/html/src/index.html#L21).

I need to double check whether this is being done in the Apple Pay README update.

packages/react-paypal-js/src/v6/hooks/useGooglePayOneTimePaymentSession.ts
EvanReinstein: @nityasp It looks like we still have the double error issue here. Would you mind updating?

packages/react-paypal-js/src/v6/components/GooglePayOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,136 @@
...
+    handleDestroy,
+    createGooglePayButton,
+  } = useGooglePayOneTimePaymentSession(hookProps);
+  const { isHydrated } = usePayPal();
+  const containerRef = useRef<HTMLDivElement>(null);
+  const buttonMountedRef = useRef(false);
+  const handleClickRef = useRef(handleClick);
+  handleClickRef.current = handleClick;
+
+  const isDisabled = disabled || isPending || error !== null;
```
EvanReinstein: Apple Pay does not include `error !== null` in its `isDisabled` value, should we update that there as well?

packages/react-paypal-js/src/v6/hooks/useGooglePayOneTimePaymentSession.ts
nityasp: Removed double-error behavior.

packages/react-paypal-js/README.md
nityasp: Added Google Pay pre-req to add the script tag to load google sdk.

packages/react-paypal-js/src/v6/components/GooglePayOneTimePaymentButton.tsx
nityasp: Thanks for pointing it out. Actually we don't need `error !== null` in neither GooglePay nor ApplePay because this check can block the retries after errors are recovered. Hence removed this check for GooglePay.

packages/react-paypal-js/README.md
EvanReinstein: Hey I'm not seeing this update, is it located in a js doc comment?

packages/react-paypal-js/src/v6/types/googlePay.ts
```diff
@@ -0,0 +1,49 @@
+import type {
+  GooglePayButtonOptions,
+  GooglePayIsReadyToPayRequest,
+  GooglePayPaymentDataRequest,
+  GooglePayPaymentMethodData,
+} from "@paypal/paypal-js/sdk-v6";
+
+declare global {
+  interface Window {
```
EvanReinstein, line 9: This is the 2nd or 3rd declare global. I wonder if we should consolidate them somehow. Nothing to do here, but I'll add this to my to do list.

packages/react-paypal-js/README.md
EvanReinstein: Oh, I see it now thanks! 👍

EvanReinstein (approved):
Great work, thanks @nityasp 🚀

PR #925, opened 2026-05-20
Fix to prevent ApplePaySession completePayment from being called twice

nityasp:
The fix prevents ApplePay Session `completePayment` from being called twice.

Previously, the hook completed the Apple Pay session with `STATUS_SUCCESS` before `onApprove` finished. If `onApprove` later threw, the error path attempted to call `completePayment` again with `STATUS_FAILURE` on the same `ApplePaySession` which can throw `InvalidAccessError`.

This change updates the hook to call `onApprove` before marking Apple Pay session successful and guards `completePayment` so it can only run once per session.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.test.ts
@@ -511,6 +511,36 @@ describe("useApplePayOneTimePaymentSession", () => {
       });
     });
 
+    test("should complete with STATUS_FAILURE when onApprove throws", async () => {
+      const onApproveError = new Error("Capture failed");
+      defaultProps.onApprove = jest.fn().mockRejectedValue(onApproveError);
+
+      const { result } = renderHook(() =>
+        useApplePayOneTimePaymentSession(defaultProps),
+      );
+
+      await act(async () => {
+        await result.current.handleClick();
+      });
+
+      expect(capturedApplePaySession).not.toBeNull();
+      await act(async () => {
+        await capturedApplePaySession!.onpaymentauthorized?.({
+          payment: {
+            token: {},
+            billingContact: {},
+          },
+        });
+      });
+
+      expect(result.current.error).toEqual(onApproveError);
+      expect(defaultProps.onError).toHaveBeenCalledWith(onApproveError);
+      expect(capturedApplePaySession!.completePayment).toHaveBeenCalledTimes(1);
+      expect(capturedApplePaySession!.completePayment).toHaveBeenCalledWith({
+        status: MockApplePaySession.STATUS_FAILURE,
+      });
+    });
+
     test("should clear error when SDK instance becomes available", () => {
       const { rerender, result } = renderHook(() =>
         useApplePayOneTimePaymentSession(defaultProps),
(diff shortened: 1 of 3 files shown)


packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.test.ts
```diff
@@ -511,6 +511,36 @@ describe("useApplePayOneTimePaymentSession", () => {
...
+      });
+
+      expect(result.current.error).toEqual(onApproveError);
+      expect(defaultProps.onError).toHaveBeenCalledWith(onApproveError);
+      expect(capturedApplePaySession!.completePayment).toHaveBeenCalledTimes(1);
+      expect(capturedApplePaySession!.completePayment).toHaveBeenCalledWith({
+        status: MockApplePaySession.STATUS_FAILURE,
+      });
+    });
+
```
HackTheW2d, line 580: Could we add a test case that `onApprove()` succeed but `completePaymentOnce` mocked to fail and then `onError` is called and the whole test case does not have any unhandled rejection?

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -306,20 +325,16 @@ export function useApplePayOneTimePaymentSession({
...
-          // Complete the Apple Pay session successfully
-          applePaySession.completePayment({
-            status: ApplePaySessionConstructor.STATUS_SUCCESS,
-          });
-
           // Call onApprove callback
           await proxyCallbacks.onApprove(confirmResult);
+
+          // Complete the Apple Pay session successfully
+          completePaymentOnce(ApplePaySessionConstructor.STATUS_SUCCESS);
```
HackTheW2d, line 340: I see we switched order here. 
Please note there's a 30s limit on [onpaymentauthorized](https://developer.apple.com/documentation/applepayontheweb/applepaysession/onpaymentauthorized) before calling `completePayment()`. Now we have 3 awaits ahead. I trust our `createOrder()`, `confirmOrder()` and `captureOrder()` can finish within seconds. But maybe it is worth noting in JSDoc to let merchants know that their `onApprove()` shouldn't be too time consuming? Or there's better solutions?

For eg,
```
  /**
   * Callback invoked when the payment is successfully approved.
   * Runs before the Apple Pay sheet is dismissed, so avoid long-running
   * async work — the total time from user authorization to completion
   * must stay within Apple Pay's 30-second window.
   */
  onApprove: (data: ConfirmOrderResponse) => void | Promise<void>;

```

EvanReinstein (approved):

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
nityasp: Yes along with the JSDoc, I added an optional post completion callback `onApproveCompleted` which can take care of non-critical tasks that helps the timeout risk in `onpaymentAuthorized`  by reducing what happens inside `onApprove`. Let me know your thoughts on this @HackTheW2d

HackTheW2d: Brilliant thought! Thanks for the work.

HackTheW2d (approved):
Nice fix!

PR #936, opened 2026-06-01
Feature: BT-PayPal Pay Later hook

EvanReinstein:
## Summary

Adds Braintree PayPal **Pay Later (BNPL)** support to `@paypal/react-paypal-js` (v6), along with a companion **eligibility** hook and the provider-context plumbing needed to cache eligibility results. Also tightens error reporting and types across the existing Braintree session hooks.

### New hooks
- **`useBraintreePayPalPayLaterSession`** — manages a Pay Later session, mirroring the existing one-time-payment / billing-agreement / checkout-with-vault hooks. Returns `{ error, isPending, handleClick }`, recreates the session when data options or the checkout instance change (deep-compared), skips recreation on callback-only changes, and won't retry a failed instance until it's replaced.
- **`useBraintreeEligibleMethods`** — calls `findEligibleMethods` on the shared checkout instance and caches the result on provider context. Fetches are deduped by `(instance, options)` so multiple mounts / remounts with the same options reuse the cached result; refetches when options or the instance change; recovers from transient failures when the consumer corrects the input; and is safe under React 18 StrictMode double-mount and post-unmount resolution.

### Provider / context plumbing
- New `SET_ELIGIBILITY` reducer action; `eligiblePaymentMethods` and `eligiblePaymentMethodsPayload` added to `BraintreePayPalState`.
- New **internal** `BraintreeDispatchContext` + `useBraintreePayPalDispatch` hook so hooks can write eligibility back into provider state. Not exported to external consumers.

### Error handling
- Provider-level (context) failures are now surfaced **separately and labeled** (`Braintree provider error: …`, with the original error as `cause`), distinct from "checkout instance not available" — so consumers can tell which layer failed.
- Per-hook instance-unavailable messages are now specific (e.g. `Braintree Pay Later checkout instance not available`, `Braintree One-Time Payment …`, `Braintree Billing Agreement …`, `Braintree Checkout With Vault …`).

### Types (`types/braintree.ts`)
- `getDetails` is now generically typed via `FindEligibleMethodsGetDetails<T>` from `@paypal/paypal-js/sdk-v6`.
- New `BraintreeEligibleFundingSource` union (`"paypal" | "paylater" | "credit"`).
- New typed `BraintreeShippingAddressOverride` (replaces `Record<string, unknown>`).
- Added `shippingCallbackUrl`, `shippingAddressOverride`, and `contactPreference` to the one-time / pay-later / checkout-with-vault option types.

### Exports & changeset
- Exports both new hooks and their `Props` / `Return` types from `v6/index.ts`.
- Adds a `minor` changeset for `@paypal/react-paypal-js`.

## Testing
- New unit tests for `useBraintreePayPalPayLaterSession` and `useBraintreeEligibleMethods` (init, refetch, caching, StrictMode, unmount, error labeling).
- Reducer tests for the `SET_ELIGIBILITY` action and `RESET_STATE`.
- Updated existing Braintree session-hook tests for the new labeled error messages and state shape.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreeEligibleMethods.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreeEligibleMethods.test.ts
@@ -1,8 +1,12 @@
 import React, { useReducer } from "react";
 import { renderHook, act } from "@testing-library/react-hooks";
+import { render, waitFor } from "@testing-library/react";
 
 import { expectCurrentErrorValue } from "../useErrorTestUtil";
-import { useBraintreeEligibleMethods } from "./useBraintreeEligibleMethods";
+import {
+  useBraintreeEligibleMethods,
+  UseBraintreeEligibleMethodsReturn,
+} from "./useBraintreeEligibleMethods";
 import {
   BraintreePayPalContext,
   braintreeInitialState,
@@ -79,10 +83,11 @@ interface ProviderHandle {
 function makeProvider(
   initialOverrides: Partial<BraintreePayPalState> = {},
 ): ProviderHandle {
-  const handle: ProviderHandle = {
-    Wrapper: () => null,
-    dispatch: () => {},
-  };
+  // `currentDispatch` is rebound on every Wrapper render. The returned
+  // `dispatch` forwards to it so callers can destructure `dispatch` from the
+  // handle BEFORE the Wrapper has mounted (when the real reducer dispatch is
+  // not yet known) and still hit the live dispatch at call time.
+  let currentDispatch: (action: BraintreeAction) => void = () => {};
 
   const Wrapper: TestWrapper = ({ children }) => {
     const [state, dispatch] = useReducer(braintreeReducer, {
@@ -91,7 +96,7 @@ function makeProvider(
       isHydrated: true,
       ...initialOverrides,
     });
-    handle.dispatch = dispatch;
+    currentDispatch = dispatch;
     return React.createElement(
       BraintreeDispatchContext.Provider,
       { value: dispatch },
@@ -103,8 +108,10 @@ function makeProvider(
     );
   };
 
-  handle.Wrapper = Wrapper;
-  return handle;
+  return {
+    Wrapper,
+    dispatch: (action) => currentDispatch(action),
+  };
 }
 
 describe("useBraintreeEligibleMethods", () => {
@@ -257,7 +264,7 @@ describe("useBraintreeEligibleMethods", () => {
       expect(firstInstance.findEligibleMethods).toHaveBeenCalledTimes(1);
 
       const secondInstance = createMockCheckoutInstance();
-      act(() => {
+      await act(async () => {
         dispatch({
           type: BRAINTREE_DISPATCH_ACTION.SET_INSTANCE,
           value: secondInstance as unknown as BraintreePayPalCheckoutInstance,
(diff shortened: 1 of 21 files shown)


packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreeEligibleMethods.test.ts
```diff
@@ -0,0 +1,276 @@
...
+  describe("initialization", () => {
+    test("should return isPending=true while provider instance is pending", () => {
+      mockBraintreeContext({ loadingStatus: INSTANCE_LOADING_STATE.PENDING });
+
+      const { result } = renderHook(() =>
+        useBraintreeEligibleMethods(defaultProps),
+      );
+
+      expect(result.current.isPending).toBe(true);
+      expect(result.current.eligibility).toBeNull();
```
HackTheW2d: Is this `eligibility` or `eligibleMethods`? 🤔 
same elsewhere

EvanReinstein: Thank you! Yes I did update the eligibility value name and need to do the same in the tests. I am giving this another once over 👍

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreeEligibleMethods.ts
```diff
@@ -0,0 +1,135 @@
...
+    setEligibleMethods(null);
+    setError(null);
+
+    braintreePayPalCheckoutInstance
+      .findEligibleMethods(memoizedOptions)
+      .then((result) => {
+        if (!isSubscribed || !isMountedRef.current) {
+          return;
+        }
+        setEligibleMethods(result);
```
HackTheW2d:
```suggestion
        setEligibleMethods(result);
        setError(null);
```

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
```diff
@@ -0,0 +1,206 @@
...
+  useEffect(() => {
+    // Reset failed instance tracking when checkout instance changes
+    if (failedInstanceRef.current !== braintreePayPalCheckoutInstance) {
+      failedInstanceRef.current = null;
+    }
+
+    if (braintreePayPalCheckoutInstance) {
+      setError(null);
+    } else if (loadingStatus !== INSTANCE_LOADING_STATE.PENDING) {
+      setError(new Error("Braintree checkout instance not available"));
```
HackTheW2d: This seems a bit general to me. Might hard to debug when error occurs.
Can we consider use:
```
const { braintreePayPalCheckoutInstance, loadingStatus, error: contextError } =
  useBraintreePayPal();
```

....
```
    setError(
      contextError
        ? new Error(`Braintree provider error: ${contextError.message}`, { cause: contextError })
        : new Error("Braintree checkout instance not available"),
    );
  }
}, [braintreePayPalCheckoutInstance, loadingStatus, contextError, setError]);
```

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreeEligibleMethods.test.ts
```diff
@@ -0,0 +1,276 @@
...
+      await act(async () => {
+        resolveFetch(createMockEligibility());
+        await pending;
+      });
+
+      // Eligibility should remain null because the post-unmount update was skipped
+      expect(result.current.eligibility).toBeNull();
+    });
+  });
+});
```
HackTheW2d, line 527: Do you mind add one more test to verify this case?

```
test("should retry with new options after a failure (transient errors must recover)", async () => {
  const checkoutInstance = {
    findEligibleMethods: jest
      .fn()
      // First call: simulate a transient failure
      .mockRejectedValueOnce(new Error("network blip"))
      // Second call: succeed
      .mockResolvedValueOnce(createMockEligibility()),
  };

  mockBraintreeContext({ braintreePayPalCheckoutInstance: checkoutInstance });

  const { rerender, waitForNextUpdate } = renderHook(
    ({ currency }) =>
      useBraintreeEligibleMethods({ ...defaultProps, currency }),
    { initialProps: { currency: "US" } },
  );

  await waitForNextUpdate();
  expect(checkoutInstance.findEligibleMethods).toHaveBeenCalledTimes(1);

  // User changes the currency — hook MUST retry
  rerender({ currency: "USD" });

  await waitForNextUpdate();
  expect(checkoutInstance.findEligibleMethods).toHaveBeenCalledTimes(2);
  expect(checkoutInstance.findEligibleMethods).toHaveBeenLastCalledWith(
    expect.objectContaining({ currency: "20.00" }),
  );
});
```

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
```diff
@@ -0,0 +1,206 @@
...
+import { useError } from "../useError";
+import {
+  createPaymentSession,
+  useProxyProps,
+  useDeepCompareMemoize,
+} from "../../utils";
+import { INSTANCE_LOADING_STATE } from "../../types/ProviderEnums";
+
+import type {
+  BraintreePayLaterSessionOptions,
```
HackTheW2d, line 14: I notice `BraintreePayLaterSessionOptions` has some other attributes:
`shippingAddressOverride`, `shippingCallbackUrl` which is not included in our `BraintreePayLaterSessionOptions`. Any reason why we won't include those?

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
EvanReinstein: Yea great idea!

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreeEligibleMethods.test.ts
EvanReinstein: Done 👍

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
EvanReinstein: Oh interesting, I wonder if those are new. I will update

EvanReinstein: Ok looks like there were a few type properties missing across the payment session types. I've updated with the newest ones 👍

HackTheW2d (approved):
Nice work!

Issue #933, opened 2026-05-31
[Bug] PayPalOneTimePaymentButton type requires "presentationMode" but in docs shown as optional


eddiejaoude:
### Library used

`react-paypal-js`

### 🐞 Describe the Bug

The documentation examples for PayPalOneTimePaymentButton imply presentationMode is optional (or defaulted to auto), but the published TypeScript types require presentationMode to be explicitly provided. As a result, the official example fails type-checking in strict TypeScript projects.

### 🔬 Minimal Reproduction

Follow the first example in your README.

### 😕 Actual Behavior

<img width="745" height="585" alt="Image" src="https://github.com/user-attachments/assets/8aa142fd-48db-41f8-abc2-ea4354ac991b" />

### 🤔 Expected Behavior

Two options:
1. docs updated to include `presentationMode` in the examples
2. remove requirement type

### 🌍 Environment

- Node.js/npm: v22
- OS: Mac
- Browser: Chrome

EvanReinstein:
@eddiejaoude Thanks for reaching out with this oversight in the README documentation and opening a PR to update. We'll review the full README and ensure that required props are included in the other examples as well in order to make integration easier.

eddiejaoude:
Thank you 🙏

EvanReinstein:
Hi @eddiejaoude I have merged 2 PRs in the last few days to address the issue that you brought up. I will update here once we've release new versions of `paypal-js` and `react-paypal-js`.

EvanReinstein:
We have released `paypal-js v9.8.0` and `react-paypal-js v9.3.0` that fixes this issue. Thank you @eddiejaoude for you collaboration and patience!

eddiejaoude:
Awesome! Thank you for the quick turnaround on this 🎉 .

PR #937, opened 2026-06-01
Feat: Make presentationMode optional for v6 hooks and button components

EvanReinstein:
This PR loosens the v6 Button and Hook prop types strictness for the key `presentationMode`. This property is currently required based on the types in `paypal-js`. With this change each hook will default to `auto` and the button components will no longer carry the requirement of including a `presentationMode` prop.

This is an alternative solution to the issue brought up [here](https://github.com/paypal/paypal-js/issues/933).

### DIFF
--- a/.changeset/funny-hands-refuse.md
+++ b/.changeset/funny-hands-refuse.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": minor
+---
+
+Makes presentationMode optional for v6 buttons and hooks, and sets a default value for each hook.
--- a/packages/react-paypal-js/README.md
+++ b/packages/react-paypal-js/README.md
@@ -260,17 +260,17 @@ import { PayPalOneTimePaymentButton } from "@paypal/react-paypal-js/sdk-v6";
 
 **Props:**
 
-| Prop               | Type                                                         | Description                                            |
-| ------------------ | ------------------------------------------------------------ | ------------------------------------------------------ |
-| `orderId`          | `string`                                                     | Static order ID (alternative to `createOrder`)         |
-| `createOrder`      | `() => Promise<{ orderId: string }>`                         | Async function to create an order                      |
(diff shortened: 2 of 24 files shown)


.changeset/funny-hands-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@paypal/react-paypal-js": minor
+---
+
+Makes presentationMode optional for v6 buttons and hooks, and sets a default value for each hook.
```
HackTheW2d, line 5: Just want to raise a potential follow up:
I see in our [types definition](https://github.com/paypal/paypal-js/blob/3d3b60e73a8b926696a9f2d6a5f02e9c66e027a2/packages/paypal-js/types/v6/components/base-component.d.ts#L92)

we put `= { presentationMode: "auto" }` in our type definition, rather than at runtime code. This might introduce some compile errors under strict mode. 
such as:
`TS1015  "Parameter cannot have question mark and initializer."`

But this error has been covered by `"skipLibCheck": true ` in our tsconfig.

Nothing to be done just want to raise awareness

packages/react-paypal-js/README.md
```diff
@@ -1535,7 +1534,6 @@ function CustomPayPalSaveButton() {
       const { vaultSetupToken } = await createVaultToken();
       return { vaultSetupToken };
     },
-    presentationMode: "popup",
```
HackTheW2d, line 1538: Is this example's purpose for teaching "how to configure with customized parameters"?
Deleting this line would make this into a basic example with purpose of "work with least config".

Can we consider this:
`presentationMode: "popup", // optional; defaults to "auto"`

HackTheW2d (approved):
Great work!

gregjopa (approved):

packages/react-paypal-js/README.md
EvanReinstein: I am going to leave this as is for now. We want to encourage integrations using "auto", if folks want to change the presentation mode they can review the other parts of the README and proceed accordingly

PR #940, opened 2026-06-01
Fix: Make payment session .start options optional

EvanReinstein:


packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -89,7 +89,7 @@ export type CreateOrderCallback = () => CreateOrderPromise;
  */
 export type BasePaymentSession = {
   start: (
-    options: PresentationModeOptionsForAuto,
+    options?: PresentationModeOptionsForAuto,
```
gregjopa: What do you think about having this express the default value?
```suggestion
    options?: PresentationModeOptionsForAuto = { presentationMode: "auto" },
```

EvanReinstein: Done 👍

packages/paypal-js/types/v6/components/paypal-guest-payments.d.ts
```diff
@@ -38,7 +38,7 @@ export type PayPalGuestOneTimePaymentSession = Omit<
   "start"
 > & {
   start: (
-    presentationModeOptions: PayPalGuestPresentationModeOptions,
+    presentationModeOptions?: PayPalGuestPresentationModeOptions,
```
gregjopa: Could you default the rest of these to default to auto too just like the base-component?

gregjopa (approved):

packages/paypal-js/types/v6/components/paypal-guest-payments.d.ts
EvanReinstein: Done 👍

PR #946, opened 2026-06-04
fix: require explicit v6 sdk environment

### LINKED ISSUE #945: [Bug] PayPalProvider / loadCoreSdkScript silently defaults to sandbox when `environment` is omitted
### Library used

both `paypal-js` and `react-paypal-js` (v6 SDK)

### Versions

- `@paypal/react-paypal-js@9.2.0`
- `@paypal/paypal-js@9.7.0`

### 🐞 Describe the bug

The v6 SDK loader silently defaults to the **sandbox** environment when the `environment` option is omitted. Because the option is documented as optional, typed as optional, and omitted from every `PayPalProvider` example in the published JSDoc, the most natural way to follow the docs produces an integration that loads the **sandbox** SDK in production — with no error, no warning, and no type complaint. The buyer-facing checkout silently talks to `https://www.sandbox.paypal.com`.

This is easy to miss because the merchant's (live) `clientId` is unrelated to which environment the SDK loads — the `client-id` does **not** select the environment in v6, only `environment` does. So a developer can have a correct production `clientId` and still ship a sandbox checkout.

### 🔬 Root cause (traced through the source)

**1. The loader defaults to sandbox via a `=== "production"` fall-through.** There is no explicit default assignment — anything that is not the exact string `"production"` (including `undefined`) selects sandbox:

```js
// @paypal/paypal-js@9.7.0/dist/v6/esm/paypal-js.js
50  function loadCoreSdkScript(options) {
54    validateArguments(options);
58    var environment = options.environment, ... ;          // read straight from options, no default
64    var baseURL = environment === "production"
65        ? "https://www.paypal.com"
66        : "https://www.sandbox.paypal.com";                // ← undefined lands here
67    var url = new URL("/web-sdk/v6/core", baseURL);        // becomes the <script src>
```

**2. Validation does not catch the omission.** `validateArguments` only rejects a *truthy, invalid* value; `undefined` short-circuits and passes silently, so nothing fails loudly:

```js
// same file
102  if (environment && environment !== "production" && environment !== "sandbox") {
105    throw new Error('The "environment" option must be either "production" or "sandbox"');
106  }
```

**3. `PayPalProvider` passes the option straight through with no default.** The only mention of `environment` in the v6 wrapper is the pass-through; `environment` is not a named prop, so it arrives via `...scriptOptions` and is `undefined` unless the consumer supplies it:

```js
// @paypal/react-paypal-js@9.2.0/dist/v6/esm/react-paypal-js.js
1717  const PayPalProvider = ({ clientMetadataId, clientToken, clientId,
1721    components = [...], locale, pageType, ...,
1730    ...scriptOptions                                    // 'environment' falls in here
1731  }) => {
1738    const loadCoreScriptOptions = useRef({ ...scriptOptions,
1740      dataSdkIntegrationSource: "react-paypal-js" });   // no environment default added
1754    await loadCoreSdkScript({
1755      environment: loadCoreScriptOptions.current.environment,  // undefined unless caller passed it
        ...
```

**4. The type system permits the omission.** `environment` is optional, so omitting it is valid TypeScript:

```ts
// @paypal/paypal-js@9.7.0/types/v6/index.d.ts
239  export interface LoadCoreSdkScriptOptions extends CoreSdkScriptDataAttributes {
240    environment?: "production" | "sandbox";
241  }
```

`PayPalProvider`'s props inherit this via `Omit<LoadCoreSdkScriptOptions, "dataSdkIntegrationSource">`, so the provider's `environment` is optional too.

**5. The documented examples omit it.** Every `@example` in `PayPalProvider.d.ts` (and the README quickstart) shows only `clientId`/`clientToken`, `components`, and `pageType` — never `environment`. Copying any example verbatim yields a sandbox-in-production integration:

```tsx
<PayPalProvider
  clientId="YOUR_CLIENT_ID"
  components={["paypal-payments"]}
  pageType="checkout"
>
```

### 😕 Actual behavior

Omitting `environment` (as all documented examples do) loads `https://www.sandbox.paypal.com/web-sdk/v6/core` in every environment, including production — silently, regardless of whether the `clientId` is a live credential.

### 🤔 Expected behavior

It should be impossible to *accidentally* ship sandbox to production. Defaulting to the safe-for-testing-but-wrong-for-prod environment when the option is simply forgotten is a footgun, especially since the value is baked at build time (e.g. via a public/inlined env var in most bundlers) and a misconfiguration is not obvious at a glance.

### ✅ Proposed fix

Make the environment an explicit, required choice rather than a silent default:

1. **Docs** — document `environment` as a **required** option for `PayPalProvider` / `loadCoreSdkScript`, with `"production"` and `"sandbox"` as the allowed values and an explicit note that the `clientId` does not select the environment.
2. **Examples** — show `environment` **set** in every `PayPalProvider` example (JSDoc `@example` blocks and README quickstart), e.g. `environment="production"` / `environment={process.env.X}`.
3. **TypeScript** — make `environment` **required** (non-optional) on `LoadCoreSdkScriptOptions` (and therefore on the provider props), so omitting it is a compile-time error rather than a silent sandbox fallback.

(Related: #568 previously asked for the `environment` option to be documented at all. This issue is the production-safety flip side of the same gap — not just "how do I reach sandbox?" but "how do I avoid silently *staying* in sandbox?")

### Repro

```tsx
import { PayPalProvider } from "@paypal/react-paypal-js/sdk-v6";

// Following the documented example verbatim — note: no `environment`.
<PayPalProvider clientId={LIVE_CLIENT_ID} components={["paypal-payments"]} pageType="checkout">
  {/* ... */}
</PayPalProvider>
```

Observe in DevTools → Network that the core SDK loads from `https://www.sandbox.paypal.com/web-sdk/v6/core`, despite a live `clientId`.

---

*This issue was written by Claude Code, with the assistance and verification of a human.*

jdjioe5-cpu:
## Summary

Closes #945.

This makes the V6 SDK environment an explicit choice instead of silently falling back to sandbox when `environment` is omitted.

Changes:
- require `environment` in `LoadCoreSdkScriptOptions`
- throw a clear runtime error when `loadCoreSdkScript()` is called without `environment`
- update V6 loader tests from "default sandbox" to "explicit environment required"
- add `environment="sandbox"` to `PayPalProvider` docs/examples and mark the prop as required
- add a changeset for `@paypal/paypal-js` and `@paypal/react-paypal-js`

## Why

In V6, the `clientId` does not select sandbox vs production. If `environment` is omitted, the current loader falls through to `https://www.sandbox.paypal.com`, which can make a production integration silently load the sandbox SDK.

## Validation

- `npx vitest run src/v6/index.test.ts` from `packages/paypal-js`
- `npm run typecheck --workspace=@paypal/paypal-js`
- `npm run typecheck --workspace=@paypal/react-paypal-js`
- `npm run test --workspace=@paypal/react-paypal-js -- src/v6/components/PayPalProvider.test.tsx --runInBand`
- `npx prettier --check .changeset/wise-seals-choose.md packages/paypal-js/README.md packages/paypal-js/src/v6/index.ts packages/paypal-js/src/v6/index.test.ts packages/paypal-js/types/v6/index.d.ts packages/react-paypal-js/README.md packages/react-paypal-js/src/v6/components/PayPalProvider.tsx`
- `git diff --check`

Note: `npm run build --workspace=@paypal/paypal-js` fails on Windows before Rollup because the package script uses `rm -rf`. I verified the needed local build by running `npx rollup --config --bundleConfigAsCjs` from `packages/paypal-js`.

### DIFF
--- a/.changeset/wise-seals-choose.md
+++ b/.changeset/wise-seals-choose.md
@@ -1,0 +1,6 @@
+---
+"@paypal/paypal-js": major
+"@paypal/react-paypal-js": major
+---
+
+Require an explicit V6 SDK `environment` option so integrations do not silently load the sandbox SDK when the option is omitted.
--- a/packages/paypal-js/README.md
+++ b/packages/paypal-js/README.md
@@ -299,18 +299,18 @@ Use `loadCoreSdkScript()` to asynchronously load the V6 SDK:
 import { loadCoreSdkScript } from "@paypal/paypal-js/sdk-v6";
 
 const paypal = await loadCoreSdkScript({
-  environment: "sandbox", // "sandbox" | "production"
+  environment: "sandbox", // required: "sandbox" | "production"
   debug: true, // optional
 });
 ```
 
 **Options:**
 
-| Option          | Type                          | Description             |
-| --------------- | ----------------------------- | ----------------------- |
-| `environment`   | `"sandbox"` \| `"production"` | Target environment      |
-| `debug`         | `boolean`                     | Enable debug mode       |
-| `dataNamespace` | `string`                      | Custom global namespace |
+| Option          | Type                          | Required | Description                                                                        |
+| --------------- | ----------------------------- | -------- | ---------------------------------------------------------------------------------- |
+| `environment`   | `"sandbox"` \| `"production"` | Yes      | Target environment. The `clientId` does not determine which environment is loaded. |
+| `debug`         | `boolean`                     | No       | Enable debug mode                                                                  |
+| `dataNamespace` | `string`                      | No       | Custom global namespace                                                            |
 
 ### Creating an SDK Instance
 
--- a/packages/paypal-js/src/v6/index.test.ts
+++ b/packages/paypal-js/src/v6/index.test.ts
@@ -23,16 +23,13 @@ describe("loadCoreSdkScript()", () => {
       });
   });
 
-  test("should default to using the sandbox environment", async () => {
-    const result = await loadCoreSdkScript();
-    expect(scriptAppendChildSpy).toHaveBeenCalledTimes(1);
-    const scriptElement = scriptAppendChildSpy.mock.calls[0][0];
-    expect(scriptElement.src).toBe(
-      "https://www.sandbox.paypal.com/web-sdk/v6/core",
+  test("should require an explicit environment", async () => {
+    expect(async () => {
+      await loadCoreSdkScript();
+    }).rejects.toThrowError(
+      'The "environment" option is required and must be either "production" or "sandbox"',
     );
-    expect(scriptElement.getAttribute("data-loading-state")).toBe("resolved");
-    expect(result).toBeDefined();
-    expect(window.paypal).toBeDefined();
+    expect(scriptAppendChildSpy).toHaveBeenCalledTimes(0);
   });
 
   test("should support options for using production environment", async () => {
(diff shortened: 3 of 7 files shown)


EvanReinstein:
Closing this PR, we will use the solution [here](https://github.com/paypal/paypal-js/pull/947). Thank you for your attention!

PR #959, opened 2026-06-11
Feat: BT-PayPal Pay Later button component

EvanReinstein:
This PR adds the BT-PayPal Pay Later static UI button component. This can be tested locally using `npm link` or `npm pack` in order to serve the package locally to the V6 Web SDK with Braintree sample integration. The [PR](https://github.com/paypal-examples/v6-web-sdk-with-braintree-sdk-sample-integration/pull/44) updates the Braintree-PayPal prebuilt React app and can be used to test the Pay Later button.

### DIFF
--- a/packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
@@ -23,7 +23,7 @@ describe("BraintreePayPalPayLaterButton", () => {
     jest.clearAllMocks();
     mockGetDetails.mockReturnValue({
       countryCode: "US",
-      productCode: "PAY_LATER",
+      productCode: "PAYLATER",
     });
     mockUseBraintreePayPalPayLaterSession.mockReturnValue({
       error: null,
@@ -94,7 +94,7 @@ describe("BraintreePayPalPayLaterButton", () => {
     const button = container.querySelector("paypal-pay-later-button");
     expect(mockGetDetails).toHaveBeenCalledWith("paylater");
     expect(button).toHaveAttribute("countryCode", "US");
-    expect(button).toHaveAttribute("productCode", "PAY_LATER");
+    expect(button).toHaveAttribute("productCode", "PAYLATER");
   });
 
   it("should render without eligibility details when not available", () => {
--- a/packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
@@ -23,7 +23,7 @@ describe("BraintreePayPalPayLaterButton", () => {
     jest.clearAllMocks();
     mockGetDetails.mockReturnValue({
       countryCode: "US",
-      productCode: "PAY_LATER_LONG_TERM",
+      productCode: "PAY_LATER",
     });
     mockUseBraintreePayPalPayLaterSession.mockReturnValue({
       error: null,
@@ -94,7 +94,7 @@ describe("BraintreePayPalPayLaterButton", () => {
     const button = container.querySelector("paypal-pay-later-button");
     expect(mockGetDetails).toHaveBeenCalledWith("paylater");
     expect(button).toHaveAttribute("countryCode", "US");
-    expect(button).toHaveAttribute("productCode", "PAY_LATER_LONG_TERM");
+    expect(button).toHaveAttribute("productCode", "PAY_LATER");
   });
 
   it("should render without eligibility details when not available", () => {
--- a/packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
+++ b/packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
@@ -1,0 +1,219 @@
+import React from "react";
+import { render, fireEvent } from "@testing-library/react";
+
+import { BraintreePayPalPayLaterButton } from "./BraintreePayPalPayLaterButton";
+import { useBraintreePayPalPayLaterSession } from "../../hooks/Braintree/useBraintreePayPalPayLaterSession";
+import { useBraintreePayPal } from "../../hooks/Braintree/useBraintreePayPal";
+
+jest.mock("../../hooks/Braintree/useBraintreePayPalPayLaterSession", () => ({
+  useBraintreePayPalPayLaterSession: jest.fn(),
+}));
+jest.mock("../../hooks/Braintree/useBraintreePayPal", () => ({
+  useBraintreePayPal: jest.fn(),
+}));
+
+describe("BraintreePayPalPayLaterButton", () => {
+  const mockHandleClick = jest.fn();
+  const mockGetDetails = jest.fn();
+  const mockUseBraintreePayPalPayLaterSession =
+    useBraintreePayPalPayLaterSession as jest.Mock;
+  const mockUseBraintreePayPal = useBraintreePayPal as jest.Mock;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockGetDetails.mockReturnValue({
+      countryCode: "US",
+      productCode: "PAY_LATER_LONG_TERM",
+    });
+    mockUseBraintreePayPalPayLaterSession.mockReturnValue({
+      error: null,
+      isPending: false,
+      handleClick: mockHandleClick,
+    });
+    mockUseBraintreePayPal.mockReturnValue({
+      isHydrated: true,
+      eligiblePaymentMethods: { getDetails: mockGetDetails },
+    });
+  });
+
+  it("should render paypal-pay-later-button when hydrated", () => {
+    const { container } = render(
+      <BraintreePayPalPayLaterButton
+        amount="100.00"
+        currency="USD"
+        onApprove={() => Promise.resolve()}
+      />,
+    );
+    expect(
+      container.querySelector("paypal-pay-later-button"),
+    ).toBeInTheDocument();
+  });
+
+  it("should render a div when not hydrated", () => {
+    mockUseBraintreePayPal.mockReturnValue({
+      isHydrated: false,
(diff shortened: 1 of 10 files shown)


packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+ *
+ *   return (
+ *     <BraintreePayPalPayLaterButton
+ *       amount="100"
+ *       currency="USD"
+ *       onApprove={handleOnApprove}
+ *     />
+ *   );
+ * }
+ */
```
HackTheW2d, line 60: Do you mind modify this comment?
Can alleviate silent fails. Because for `<paypal-pay-later-button ... />` I think `countryCode` or `productCode` is necessary. I see we have added those in this component wrapper. But if it's unavailable for merchant, it could have silent fail
```
/**
 * `BraintreePayPalPayLaterButton` is a prebuilt button.
 *
 * **Requires eligibility data.** This component reads `countryCode` and
 * `productCode` from `useBraintreePayPal().eligiblePaymentMethods`, which is
 * populated by `useBraintreeEligibleMethods`. **Without eligibility, the button
 * renders with `display: none` and is invisible.** Wait for eligibility before
 * rendering:
 *
 * @example
 * function CheckoutButtons() {
 *   const { eligiblePaymentMethods, isLoading } = useBraintreeEligibleMethods({
 *     amount, currency, countryCode: "US", paymentFlow: "ONE_TIME_PAYMENT",
 *   });
 *   if (isLoading) return <Spinner />;
 *   if (!eligiblePaymentMethods?.isEligible("paylater")) return null;
 *   return <BraintreePayPalPayLaterButton ... />;
 * }
 */
```

EvanReinstein: Thanks for calling out this explicitness. I'll update here and include this type of change as an item to audit for in the documentation audit task for next sprint.

EvanReinstein: I ended up pushing more changes to update the readme and other hooks/components' js doc comments.

packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const mockGetDetails = jest.fn();
+  const mockUseBraintreePayPalPayLaterSession =
+    useBraintreePayPalPayLaterSession as jest.Mock;
+  const mockUseBraintreePayPal = useBraintreePayPal as jest.Mock;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockGetDetails.mockReturnValue({
+      countryCode: "US",
+      productCode: "PAY_LATER_LONG_TERM",
```
HackTheW2d: Do we have this `productCode` for PayLater?
`PAY_LATER_LONG_TERM` seems only exist in paypalMessages

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
HackTheW2d, line 104: Here it seems `contactPreference` were missing.
`BraintreePayLaterSessionOptions` contains `contactPreference` and alias as `UseBraintreePayPalPayLaterSessionProps` here. But destructuring here didn't include. if merchants add this prop in the button button component, it could have no impact. Please confirm this and other props as well :)

packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  useEffect(() => {
+    if (error) {
+      console.error(error);
+    }
+  }, [error]);
+
+  return isHydrated ? (
+    <paypal-pay-later-button
+      onClick={handleClick}
```
HackTheW2d, line 79: I think the inner event for this web component should be `pay-later-click`? But this event is not listened to when bubbling up. And we choose to use `onClick` Event here. Nothing to change, just want to draw some attention. I am not sure in the future if the sdk will ban using `onClick` and use only `pay-later-click` event in the future plan?

EvanReinstein: I see `onClick` being used in the PayPal pay later button, I don't think we'll opt for a different web component property.

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
EvanReinstein: Good catch, i've updated and will push the change.

packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.test.tsx
EvanReinstein: Oh interesting, claude is generating some odd stuff here.

packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+ *   if (isLoading) return <Spinner />;
+ *
+ *   const handleOnApprove = async (data) => {
+ *     const { nonce } = await braintreePayPalCheckoutInstance.tokenizePayment(data);
+ *     // Send nonce to your server to complete the transaction
+ *   };
+ *
+ *   if (!eligiblePaymentMethods?.isEligible("paylater")) {
+ *     return null;
+ *   }
```
HackTheW2d, line 49: Sorry. My example is a bit wrong, `useBraintreeEligibleMethods` is different it should be 
```
if (!eligiblePaymentMethods?.paylater) {
  return null;
}
```
cause I see `BraintreeEligibilityResult` with type of 
```
export interface BraintreeEligibilityResult {
  paypal: boolean;
  paylater: boolean;
  credit: boolean;
  getDetails: <T extends BraintreeEligibleFundingSource>(
    methodName: T,
  ) => FindEligibleMethodsGetDetails<T>;
}
```

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
```diff
@@ -34,29 +34,36 @@ export interface UseBraintreePayPalPayLaterSessionReturn {
...
  *
  *   return (
- *     <paypal-button
- *       type="pay"
- *       fundingSource="paylater"
+ *     <paypal-pay-later-button
  *       onClick={() => handleClick()}
  *       disabled={isPending}
+ *       country={payLaterDetails?.countryCode}
+ *       product={payLaterDetails?.productCode}
```
HackTheW2d: this should be `countryCode` and `productCode` right?

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalPayLaterSession.ts
```diff
@@ -34,29 +34,36 @@ export interface UseBraintreePayPalPayLaterSessionReturn {
  * @returns Object with: `error` (any session error), `isPending` (checkout instance loading), `handleClick` (starts session)
  *
  * @example
- * // Custom button using the hook directly with a <paypal-button> web component
+ * // Custom button using the hook directly with a <paypal-pay-later-button> web component
  * function PayPalPayLaterButton(props: UseBraintreePayPalPayLaterSessionProps) {
  *   const { isPending, handleClick } = useBraintreePayPalPayLaterSession(props);
+ *   const { isLoading, eligiblePaymentMethods } = useBraintreeEligibleMethods();
```
HackTheW2d: I think for `useBraintreeEligibleMethods`, `currency` is a required prop right?

```
export interface BraintreeFindEligibleMethodsOptions {
  amount?: string;
  currency: string;
  countryCode?: string;
  paymentFlow?:
    | "ONE_TIME_PAYMENT"
    | "VAULT_WITH_PAYMENT"
    | "VAULT_WITHOUT_PAYMENT"
    | "RECURRING_PAYMENT";
}

```

packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * `productCode` from `useBraintreePayPal().eligiblePaymentMethods`, which is
+ * populated by `useBraintreeEligibleMethods`. **Without eligibility, the button
+ * renders with `display: none` and is invisible.** Wait for eligibility before
+ * rendering.
+ *
+ * For full control over the button UI, use the {@link useBraintreePayPalPayLaterSession}
+ * hook directly instead.
+ *
+ * @example
+ * function CheckoutButtons() {
```
HackTheW2d, line 31:
```suggestion
* function CheckoutButtons({ amount, currency }: { amount: string; currency: string }) {
```

EvanReinstein: Yea I went back and forth on this, mostly b/c the way that `amount` is determined is up to the integrating developer to sort out. I'll update the js doc comment but I'm not going to prescribe that these values are passed as props.

packages/react-paypal-js/src/v6/components/Braintree/BraintreePayPalPayLaterButton.tsx
EvanReinstein: Great catch, I'd assumed that everything was the same across BT and PayPal JS SDK eligibility helpers. But that is incorrect! The `isEligible` is abstracted away. I'll update these examples.

Screenshot 2026-06-15 at 9 11 16 AM

HackTheW2d (approved):
Great work!

PR #974, opened 2026-06-23
fix buyer country in BCDC button

HackTheW2d:
Current `buyerCountry` is incorrect JSX key. The correct one to use should be `buyer-country` which though is not following naming conventions in react. So we try to deal with this issue by encouraging passing `buyerCountry` via `usePayPalGuestPaymentSession` or `PayPalGuestPaymentButton`. And if merchants decide to use web-component alone, this should be passed as `<paypal-basic-card-button buyer-country="US" ...>`


EvanReinstein (approved):
Great fix @HackTheW2d !

PR #982, opened 2026-06-29
Feature: BT-PayPal Messages

EvanReinstein:
This PR adds a wrapper for the Braintree Web SDK with PayPal JS SDK `createMessages` method. This can be tested locally using the changes in this [PR](https://github.com/paypal-examples/v6-web-sdk-with-braintree-sdk-sample-integration/pull/49).


packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.ts
```diff
@@ -0,0 +1,212 @@
...
+      failedInstanceRef.current?.instance === braintreePayPalCheckoutInstance &&
+      failedInstanceRef.current?.buyerCountry === buyerCountry &&
+      failedInstanceRef.current?.currencyCode === currencyCode
+    ) {
+      return;
+    }
+
+    let isSubscribed = true;
+    setIsCreating(true);
+
```
HackTheW2d, line 125:
```suggestion
    setError(null);
```

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (!isMountedRef.current) {
+        return;
+      }
+
+      if (!messages) {
+        setError(new Error("PayPal Messages instance not available"));
+        return;
+      }
+
+      const result = await messages.fetchContent(options);
```
HackTheW2d, line 177: Please make sure it's null rather than empty `MessageContent` object.

packages/react-paypal-js/src/v6/types/braintree.ts
```diff
@@ -340,6 +340,33 @@ export interface BraintreePaymentSession {
...
+export interface BraintreeMessagesOptions {
+  buyerCountry?: string;
+  currencyCode?: string;
+}
+
+// Content returned by `fetchContent`. Exposes `update` so the amount displayed
+// in the <paypal-message> can be changed (e.g. when cart quantity changes)
+// without re-fetching. Extra fields are passed through to `setContent`.
+export interface BraintreeMessageContent {
+  update: (options: { amount?: string }) => void;
```
HackTheW2d: Do we want to return void for this function? Is this on purpose because we want to teach merchants to fire and forget. Or it could be `Promise<BraintreeMessageContent | null>;` because Braintree also uses `paypal-message`. I could be wrong.

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.ts
```diff
@@ -0,0 +1,212 @@
...
+  ]);
+
+  const handleFetchContent = useCallback(
+    async (options?: BraintreeFetchMessageContentOptions) => {
+      if (!isMountedRef.current) {
+        return;
+      }
+
+      if (!messages) {
+        setError(new Error("PayPal Messages instance not available"));
```
HackTheW2d: Do we want to use `Braintree PayPal Messages instance not available` to be consistent with line 96?

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.ts
HackTheW2d: Since it's returning a sentinel value that will collapse, I wouldn't worry too much about checking null here.
```
      try {
        const result = await messages.fetchContent(options);

        if (!isMountedRef.current) {
          return;
        }

        // The PayPal Messages runtime currently resolves API/no-content failures
        // with an empty MessageContent sentinel so <paypal-message> can collapse
        // itself. Keep the null branch only for SDK/type compatibility.
        if (result == null) {
          setError(new Error("Failed to fetch PayPal Messages content"));
          return;
        }

        return result;
      } catch (err) {
        if (isMountedRef.current) {
          // fetchContent normally resolves API/no-content failures; this only
          // covers unexpected rejections such as validation, callback, or runtime errors.
          setError(err);
        }
      }
```

HackTheW2d (approved):
Nice work! Just some minor comments

packages/react-paypal-js/src/v6/types/braintree.ts
```diff
@@ -340,6 +340,38 @@ export interface BraintreePaymentSession {
...
+  buyerCountry?: string;
+  currencyCode?: string;
+}
+
+// Content returned by `fetchContent`. Exposes `update` so the amount displayed
+// in the <paypal-message> can be changed (e.g. when cart quantity changes)
+// without re-fetching. Extra fields are passed through to `setContent`.
+export interface BraintreeMessageContent {
+  update: (options?: {
+    amount?: string;
```
HackTheW2d: May I ask why `options?` is only allowed to pass `amount?`. The real update seems to support all the other params like `buyerCountry` etc. Or is it because Braintree only supports this?

packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.ts
```diff
@@ -0,0 +1,218 @@
...
+    async (options?: BraintreeFetchMessageContentOptions) => {
+      if (!isMountedRef.current) {
+        return;
+      }
+
+      if (!messages) {
+        setError(new Error("Braintree PayPal Messages instance not available"));
+        return;
+      }
+
```
HackTheW2d, line 174:
```suggestion

      setError(null);
```

HackTheW2d (approved):
Thanks for helping addressing comments! Just some minor comments

PR #986, opened 2026-06-30
fix apple pay disable property

HackTheW2d:
Screenshot 2026-06-30 at 9 55 38 AM
Screenshot 2026-06-30 at 9 55 32 AM

Remove the disabled attribute on web component. Add a customized attribute `isDisabled`, consistent with how we dealt with google payment button

### DIFF
--- a/packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
+++ b/packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
@@ -7,7 +7,8 @@ import type { UseApplePayOneTimePaymentSessionProps } from "../hooks/useApplePay
 import type { ApplePayButtonElementProps } from "../types/sdkWebComponents";
 
 export type ApplePayOneTimePaymentButtonProps =
-  UseApplePayOneTimePaymentSessionProps & ApplePayButtonElementProps;
+  UseApplePayOneTimePaymentSessionProps &
+    Omit<ApplePayButtonElementProps, "disabled">;
 
 /**
  * `ApplePayOneTimePaymentButton` renders a native Apple Pay button and manages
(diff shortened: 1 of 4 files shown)


packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
```diff
@@ -84,13 +95,16 @@ export const ApplePayOneTimePaymentButton = ({
   }
 
   return (
-    <div className={className}>
+    <div
+      className={className}
+      aria-disabled={isDisabled || undefined}
+      style={isDisabled ? { pointerEvents: "none", opacity: "0.5" } : undefined}
```
HackTheW2d: @EvanReinstein We put our own `isDisabled` here rather than get rid of it for consistency.(same with google payment) Let me know if you think it's not applicable to apple payment :)

EvanReinstein: I think that adding this layer is not needed. We should let merchants handle the UI as they see fit, especially considering Apple handles disabling the button behind the scenes

EvanReinstein: So i think we can remove the isDisabled logic 👍

EvanReinstein (approved):
Thanks for catching and putting in this fix. One comment around `isDisabled` but this should be good to merge after removing that.

packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
HackTheW2d: Removed :)

gregjopa (approved):

packages/react-paypal-js/src/v6/components/ApplePayOneTimePaymentButton.tsx
gregjopa: I agree with removing this `isDisabled` logic all together. We should not have that abstraction on top of ApplePay or GooglePay buttons

PR #926, opened 2026-05-21
feat(react-paypal-js): Add LPM Generic React wrapper with Storybook support

Mathanjeevika:
# PR Description — `feature/lpm-react-wrapper` → `main`
# Jira: DTLAMBO-597

---

**[DTLAMBO-597] feat(lpm): add LPM React wrapper for all 50 payment methods**
Jira: https://paypal.atlassian.net/browse/DTLAMBO-597

> **Update:** This PR now covers **all 50 LPMs** from the [Manual Validation
> tracker](https://paypal.atlassian.net/wiki/spaces/~7120205034c011404f436ba8507cce803f2ff2/pages/2988863925/LPM+Manual+Validation+Process+for+PayPal+React+Wrapper+Testing).
> The final 16 (previously excluded) were added in commit
> `feat(lpm): add remaining 16 LPMs and simplify story variants`. They are pure
> registry-driven additions — no new component logic.

### New files

- **`packages/react-paypal-js/src/v6/config/lpmRegistry.ts`**
  Single source of truth for all LPMs. Each entry declares `component`,
  `buttonTag`, `sessionMethod`, `displayName`, `fields` (SDK-rendered iframes),
  and `sessionFields` (session-level inputs such as `phone`, `billingAddress`).
  Now covers all 50 LPMs.

- **`packages/react-paypal-js/src/v6/hooks/useLPMOneTimePaymentSession.ts`**
  Generic hook — resolves the LPM session method from `LPM_REGISTRY`, calls
  `sdkInstance[sessionMethod](callbacks)`, and handles React lifecycle
  (create on SDK load, destroy on unmount, retry on error).

- **`packages/react-paypal-js/src/v6/components/LPMOneTimePaymentButton.tsx`**
  Generic button component — renders `<{lpm}-button>` web component wired to
  `handleClick`, `isPending`, and `error` from `useLPMOneTimePaymentSession`.

- **`packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx`**
  Factory module with two exports:
  - `createEnhancedLPMHook(lpm, fields, buttonTag)` — creates a named
    `use*OneTimePaymentSession` hook that returns everything from
    `useLPMOneTimePaymentSession` plus a **stable** `NameField`/`EmailField`
    component per field. Field components are created once via `useRef` and use a
    pub/sub listener pattern so the SDK iframe survives session changes without
    unmount/remount.
  - `createLPMButtonComponent(buttonTag, displayName)` — creates a standalone
[...]

PR #992, opened 2026-07-13
feat(paypal-js): add v6 Local Payment Methods (LPM) types

Mathanjeevika:
## Summary
- Adds TypeScript type declarations for the v6 SDK's Local Payment Methods (LPM) surface: `LPMComponents`, `LPMSessionMethodName`, `LPMOneTimePaymentSession`, and related types in `packages/paypal-js/types/v6/components/lpm-payments.d.ts`.
- Wires these types into `packages/paypal-js/types/v6/index.d.ts`.

## Context
This is PR 1 of 3 split out from #926 per review feedback requesting the original monolithic PR be broken up. This PR contains only the `paypal-js` type declarations, with no runtime code changes. It's a foundational dependency for the React wrapper split out in a follow-up PR.

## Test plan
- [x] `npm run typecheck` passes
- [x] `packages/paypal-js/src/v6/index.test.ts` updated and passing
- [x] Changeset included (`.changeset/lpm-paypal-js-types.md`)

### DIFF
--- a/packages/paypal-js/src/v6/index.test.ts
+++ b/packages/paypal-js/src/v6/index.test.ts
@@ -150,8 +150,8 @@ describe("loadCoreSdkScript()", () => {
 
   test("should error when environment is omitted", async () => {
     await expect(async () => {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      await loadCoreSdkScript({} as any);
+      // @ts-expect-error invalid arguments
+      await loadCoreSdkScript({});
     }).rejects.toThrow(
       'The "environment" option is required and must be either "production" or "sandbox"',
     );
@@ -159,8 +159,8 @@ describe("loadCoreSdkScript()", () => {
 
   test("should error when environment is explicitly undefined", async () => {
     await expect(async () => {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      await loadCoreSdkScript({ environment: undefined } as any);
+      // @ts-expect-error invalid arguments
+      await loadCoreSdkScript({ environment: undefined });
     }).rejects.toThrow(
       'The "environment" option is required and must be either "production" or "sandbox"',
     );
--- a/packages/paypal-js/src/v6/index.test.ts
+++ b/packages/paypal-js/src/v6/index.test.ts
@@ -150,8 +150,8 @@ describe("loadCoreSdkScript()", () => {
 
   test("should error when environment is omitted", async () => {
     await expect(async () => {
-      // @ts-expect-error environment is required
-      await loadCoreSdkScript({});
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      await loadCoreSdkScript({} as any);
     }).rejects.toThrow(
       'The "environment" option is required and must be either "production" or "sandbox"',
     );
@@ -159,8 +159,8 @@ describe("loadCoreSdkScript()", () => {
 
   test("should error when environment is explicitly undefined", async () => {
     await expect(async () => {
-      // @ts-expect-error environment is required
-      await loadCoreSdkScript({ environment: undefined });
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      await loadCoreSdkScript({ environment: undefined } as any);
     }).rejects.toThrow(
       'The "environment" option is required and must be either "production" or "sandbox"',
     );
--- a/packages/paypal-js/src/v6/index.test.ts
+++ b/packages/paypal-js/src/v6/index.test.ts
@@ -186,8 +186,7 @@ describe("loadCoreSdkScript()", () => {
       );
 
       expect(result).toBeDefined();
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      expect(window[customNamespace as any]).toBeDefined();
+      expect(customNamespace in window).toBe(true);
     });
 
     test("should error when dataNamespace is an empty string", async () => {
--- a/packages/paypal-js/types/v6/components/lpm-payments.d.ts
+++ b/packages/paypal-js/types/v6/components/lpm-payments.d.ts
@@ -1,0 +1,189 @@
+import {
+  BasePaymentSessionOptions,
+  BasePaymentSession,
+  PresentationModeOptionsForPopup,
+  PresentationModeOptionsForAuto,
+} from "./base-component";
+
+export type LPMOneTimePaymentSessionOptions = BasePaymentSessionOptions & {
+  orderId?: string;
+};
+
+export type LPMPresentationModeOptions =
+  | PresentationModeOptionsForAuto
+  | PresentationModeOptionsForPopup;
+
+/** Merchant-provided session-level inputs required by specific LPMs. */
+export type LPMSessionFieldPhone = {
+  countryCode: string;
+  nationalNumber: string;
+};
+
+export type LPMSessionFieldBillingAddress = {
+  addressLine1: string;
+  addressLine2?: string;
+  adminArea1: string;
+  adminArea2?: string;
+  postalCode: string;
+  countryCode: string;
+};
+
+export type LPMSessionFieldTaxInfo = {
+  taxId: string;
+  taxIdType: string;
+};
+
+/**
+ * Optional session-level inputs collected by the merchant and forwarded to the
+ * SDK when starting an LPM payment session. Only a subset of fields is required
+ * per LPM (see `LPMConfig.sessionFields` in the registry).
+ */
+export type LPMSessionFields = {
+  phone?: LPMSessionFieldPhone;
+  billingAddress?: LPMSessionFieldBillingAddress;
+  taxInfo?: LPMSessionFieldTaxInfo;
+  expiryDate?: string;
+  dateOfBirth?: string;
+  numberOfInstallments?: number;
+};
+
+/**
+ * Options passed to `LPMOneTimePaymentSession.start()`.
+ * Extends the presentation-mode options with optional merchant-collected
+ * session fields (phone, billingAddress, taxInfo, etc.) required by some LPMs.
+ */
+export type LPMStartOptions = LPMPresentationModeOptions & LPMSessionFields;
+
+export type LPMOneTimePaymentSessionPromise = Promise<{
+  orderId: string;
+  phone?: { countryCode: string; nationalNumber: string };
+  billingAddress?: {
+    addressLine1: string;
+    addressLine2?: string;
+    adminArea1: string;
+    adminArea2?: string;
+    postalCode: string;
+    countryCode: string;
+  };
+  taxInfo?: { taxId: string; taxIdType: string };
+  expiryDate?: string;
+}>;
+
+export type LPMOneTimePaymentSession = Omit<BasePaymentSession, "start"> & {
+  start: (
+    options: LPMStartOptions,
+    paymentSessionPromise?: LPMOneTimePaymentSessionPromise,
+  ) => Promise<void>;
+  createPaymentFields: (options: { type: string; value?: string }) => HTMLElement;
+  validate: () => Promise<boolean>;
+};
+
+export type LPMComponents =
+  | "ideal-payments"
+  | "bancontact-payments"
+  | "eps-payments"
+  | "blik-payments"
+  | "mybank-payments"
+  | "trustly-payments"
+  | "p24-payments"
+  | "multibanco-payments"
+  | "bizum-payments"
+  | "swish-payments"
+  | "klarna-payments"
+  | "twint-payments"
+  | "wechatpay-payments"
+  | "afterpay-payments"
+  | "oxxopay-payments"
+  | "boletobancario-payments"
+  | "verkkopankki-payments"
+  | "payu-payments"
+  | "paysafecard-payments"
+  | "mbway-payments"
+  | "satispay-payments"
+  | "wero-payments"
+  | "floa-payments"
+  | "scalapay-payments"
+  | "grabpay-payments"
+  | "pix-international-payments"
+  | "sepa-payments"
+  | "crypto-payments"
+  | "doku-payments"
+  | "dragonpay-payments"
+  | "estoniabank-payments"
+  | "fpx-payments"
+  | "gopay-payments"
+  | "alipay-payments"
+  | "indomaret-payments"
+  | "indonesiabanks-payments"
+  | "kredivo-payments"
+  | "linkaja-payments"
+  | "ovo-payments"
+  | "paysera-payments"
+  | "skrill-payments"
+  | "thailandbanks-payments"
+  | "blikpaylater-payments"
+  | "alfamart-payments"
+  | "zip-payments"
+  | "bancomatpay-payments"
+  | "latviabanks-payments"
+  | "fiuu-cash-payments"
+  | "lithuaniabanks-payments"
+  | "jeniuspay-payments";
+
+export type LPMSessionMethodName =
+  | "createIdealOneTimePaymentSession"
+  | "createBancontactOneTimePaymentSession"
+  | "createEpsOneTimePaymentSession"
+  | "createBlikOneTimePaymentSession"
+  | "createMyBankOneTimePaymentSession"
+  | "createTrustlyOneTimePaymentSession"
+  | "createP24OneTimePaymentSession"
+  | "createMultibancoOneTimePaymentSession"
+  | "createBizumOneTimePaymentSession"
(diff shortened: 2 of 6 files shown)


packages/paypal-js/src/v6/index.test.ts
```diff
@@ -150,17 +150,17 @@ describe("loadCoreSdkScript()", () => {
 
   test("should error when environment is omitted", async () => {
     await expect(async () => {
-      // @ts-expect-error environment is required
-      await loadCoreSdkScript({});
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      await loadCoreSdkScript({} as any);
```
EvanReinstein: Does this need to be added? Same with the `as any` below. I don't think we need to cast this in our tests.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -237,7 +244,7 @@ interface CoreSdkScriptDataAttributes {
 }
 
 export interface LoadCoreSdkScriptOptions extends CoreSdkScriptDataAttributes {
-  environment: "production" | "sandbox";
+  environment?: "production" | "sandbox";
```
EvanReinstein: This is accidentally reverting a prior change.
```suggestion
  environment: "production" | "sandbox";
```

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,189 @@
...
+  | "gopay-payments"
+  | "alipay-payments"
+  | "indomaret-payments"
+  | "indonesiabanks-payments"
+  | "kredivo-payments"
+  | "linkaja-payments"
+  | "ovo-payments"
+  | "paysera-payments"
+  | "skrill-payments"
+  | "thailandbanks-payments"
```
EvanReinstein: slightly different component name, which is odd b/c other component names do not have the hyphen `indonesiabanks-payment` for example.
```suggestion
  | "thailand-banks-payments"
```

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,189 @@
...
+    addressLine1: string;
+    addressLine2?: string;
+    adminArea1: string;
+    adminArea2?: string;
+    postalCode: string;
+    countryCode: string;
+  };
+  taxInfo?: { taxId: string; taxIdType: string };
+  expiryDate?: string;
+}>;
```
EvanReinstein: Can we re-use the types that are defined above? Maybe this type `LPMSessionFields` could be re-used more broadly?

```
export type LPMOneTimePaymentSessionPromise = Promise<{
  orderId: string;
  phone?: LPMSessionFieldPhone;
  billingAddress?: LPMSessionFieldBillingAddress;
  taxInfo?: LPMSessionFieldTaxInfo;
  expiryDate?: string;
}>;
```

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,189 @@
...
+  | "createBlikPayLaterOneTimePaymentSession"
+  | "createAlfamartOneTimePaymentSession"
+  | "createZipOneTimePaymentSession"
+  | "createBancomatPayOneTimePaymentSession"
+  | "createLatviaBanksOneTimePaymentSession"
+  | "createFIUUOneTimePaymentSession"
+  | "createLithuaniaBanksOneTimePaymentSession"
+  | "createJeniuspayOneTimePaymentSession";
+
+export type LPMPaymentsInstance = {
```
EvanReinstein, line 145: Can you add a JS Doc comment that is in line with the other instance interfaces/types? `VenmoPaymentsInstance` is a good example.

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,189 @@
...
+  taxInfo?: { taxId: string; taxIdType: string };
+  expiryDate?: string;
+}>;
+
+export type LPMOneTimePaymentSession = Omit<BasePaymentSession, "start"> & {
+  start: (
+    options: LPMStartOptions,
+    paymentSessionPromise?: LPMOneTimePaymentSessionPromise,
+  ) => Promise<void>;
+  createPaymentFields: (options: { type: string; value?: string }) => HTMLElement;
```
EvanReinstein: I see `createPaymentFields` `option` typed more tightly internally. Is there a reason to leave the `style` key off of the options here? Also should the `type` be more tightly typed to the actual values `"email" | "name" | "tax_id" | "tax_id_type"` ?

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+  dateOfBirth?: string;
+  numberOfInstallments?: number;
+};
+
+/**
+ * Options passed to `LPMOneTimePaymentSession.start()`.
+ * Extends the presentation-mode options with optional merchant-collected
+ * session fields (phone, billingAddress, taxInfo, etc.) required by some LPMs.
+ */
+export type LPMStartOptions = LPMPresentationModeOptions & LPMSessionFields;
```
EvanReinstein: Why are the LPM session fields being added to the LPMStartOptions? Looking at the internal code, the first argument for `.start` is `options` and can include either `presentationMode` or `fullPageOverlay`, why is `options` being expanded to include these other properties?

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Optional session-level inputs collected by the merchant and forwarded to the
+ * SDK when starting an LPM payment session. Only a subset of fields is required
+ * per LPM (see `LPMConfig.sessionFields` in the registry).
+ */
+export type LPMSessionFields = {
+  phone?: LPMSessionFieldPhone;
+  billingAddress?: LPMSessionFieldBillingAddress;
+  taxInfo?: LPMSessionFieldTaxInfo;
+  expiryDate?: string;
```
EvanReinstein: Looking more closely at the internal code it looks like expiryDate is shaped like this
```
expiryDate: {expiry_date: string} <-- this is the shape of ExpiryDate type
```

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+ */
+export type LPMStartOptions = LPMPresentationModeOptions & LPMSessionFields;
+
+export type LPMOneTimePaymentSessionPromise = Promise<{
+  orderId: string;
+  phone?: LPMSessionFieldPhone;
+  billingAddress?: LPMSessionFieldBillingAddress;
+  taxInfo?: LPMSessionFieldTaxInfo;
+  expiryDate?: string;
+}>;
```
EvanReinstein: It looks like this is also missing date of birth and number of installments fields, shouldn't we just re-use `LPMSessionFields` here?

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+  PresentationModeOptionsForPopup,
+  PresentationModeOptionsForAuto,
+} from "./base-component";
+
+export type LPMOneTimePaymentSessionOptions = BasePaymentSessionOptions & {
+  orderId?: string;
+};
+
+export type LPMPresentationModeOptions =
+  | PresentationModeOptionsForAuto
```
HackTheW2d: Can you please check in sdk that we really have "auto" for LPM session?

packages/paypal-js/types/v6/components/lpm-payments.d.ts
HackTheW2d, line 1: Can we also add `packages/paypal-js/types/v6/tests/lpm-payments.test.ts`? Every other v6 component family has a dedicated compile-time type test under `packages/paypal-js/types/v6/tests/` (applepay-payments.test.ts, googlepay-payments.test.ts, venmo-payments.test.ts, etc.), but this PR adds lpm-payments.d.ts with no corresponding lpm-payments.test.ts.
This matters beyond convention: these type tests are exactly the mechanism that would catch the signature issues already raised on this PR. For eg, a test that exercises `createInstance(["ideal-payments"]) `-> `createIdealOneTimePaymentSession` -> `start(...)` would surface, at compile time, that session fields belong in the promise (2nd arg) (The problem @EvanReinstein Just pointed out). 
Could we add `types/v6/tests/lpm-payments.test.ts` modeled on `googlepay-payments.test.ts`, including a couple of `// @ts-expect-error assertions` for the invalid cases (e.g. "auto" presentation mode, session fields passed in the first start argument)? That turns these from review catches into CI-enforced guarantees.

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+};
+
+/**
+ * Options passed to `LPMOneTimePaymentSession.start()`.
+ * Extends the presentation-mode options with optional merchant-collected
+ * session fields (phone, billingAddress, taxInfo, etc.) required by some LPMs.
+ */
+export type LPMStartOptions = LPMPresentationModeOptions & LPMSessionFields;
+
+export type LPMOneTimePaymentSessionPromise = Promise<{
```
HackTheW2d: I saw we have `dateOfBirth` and `numberOfInstallments` in `LPMSessionFields`. Should them also be included here?

packages/paypal-js/types/v6/components/lpm-payments.d.ts
HackTheW2d: Good catch @EvanReinstein
Also I think since we have `?` to make each method type optional. This would lead to code like:
```
const sdk = await createInstance({ components: ["ideal-payments"] });
sdk.createIdealOneTimePaymentSession?.({ onApprove });
// Note: ? here is a must because it's of type of `(() => ...) | undefined`
```

Wonder if we can consider make this more precise by:
```
type RequestedLPMMethods<T> = {
  [K in LPMSessionMethodName as MethodToComponent[K] extends T[number] ? K : never]:
    (options: LPMOneTimePaymentSessionOptions) => LPMOneTimePaymentSession;
};
```
The only con here is we need to maintain `MethodToComponent`. But this follows what we did in other payment methods such as `sdk.createApplePayOneTimePaymentSession()`.

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+
+/** Merchant-provided session-level inputs required by specific LPMs. */
+export type LPMSessionFieldPhone = {
+  countryCode: string;
+  nationalNumber: string;
+};
+
+export type LPMSessionFieldBillingAddress = {
+  addressLine1: string;
+  addressLine2?: string;
```
HackTheW2d: May I ask why we make `addressLine2` and `adminArea2` optional? I see the internal `BillingAddress` marks these all required props.

EvanReinstein:
@Mathanjeevika looks like there is a build failure, please take a look.

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,188 @@
...
+  countryCode: string;
+};
+
+export type LPMSessionFieldTaxInfo = {
+  taxId: string;
+  taxIdType: string;
+};
+
+export type LPMSessionFieldExpiryDate = {
+  expiry_date: string;
```
HackTheW2d, line 34: While all other fields are camel cases, Please confirm here `expiry_date` is correct in runtime.

packages/paypal-js/types/v6/index.d.ts
```diff
@@ -163,7 +167,10 @@ export type SdkInstance<T extends readonly Components[]> = BaseInstance &
   ("applepay-payments" extends T[number] ? ApplePayPaymentsInstance : unknown) &
   ("googlepay-payments" extends T[number]
     ? GooglePayPaymentsInstance
-    : unknown);
+    : unknown) &
+  ([Extract<T[number], LPMComponents>] extends [never]
+    ? unknown
+    : LPMPaymentsInstance);
```
HackTheW2d: This gates the *entire* `LPMPaymentsInstance` (all 50 optional methods) on the presence of *any* LPM component. So requesting `["ideal-payments"]` surfaces all 50 `create…?` methods in autocomplete, whereas `applepay-payments`/`googlepay-payments` narrow precisely to their own interface.

Root cause: `LPMComponents` and `LPMSessionMethodName` are two independent, positionally-aligned unions with no type-level link, so TS can't derive a method from a component. That alignment is also an unenforced hand-maintained invariant (drift risk).

A single source-of-truth map fixes both the narrowing *and* the drift risk:

```ts
// 1) one map = the only place the 50 pairs are written
type LPMComponentToSessionMethod = {
  "ideal-payments": "createIdealOneTimePaymentSession";
  "bancontact-payments": "createBancontactOneTimePaymentSession";
  // …50 entries…
  "jeniuspay-payments": "createJeniuspayOneTimePaymentSession";
};

// 2) derive both unions (no more parallel lists to keep in sync)
export type LPMComponents = keyof LPMComponentToSessionMethod;
export type LPMSessionMethodName =
  LPMComponentToSessionMethod[keyof LPMComponentToSessionMethod];

// 3) key-remapping: emit only the methods for the requested components
export type LPMInstanceFor<T extends readonly Components[]> = {
  [C in Extract<T[number], LPMComponents>
    as LPMComponentToSessionMethod[C]]?: (
    options: LPMOneTimePaymentSessionOptions,
  ) => LPMOneTimePaymentSession;
};

// 4) SdkInstance — no conditional needed; empty extract → {} → no-op
//    ... & LPMInstanceFor<T>;
```

Result: `["ideal-payments"]` yields only `createIdealOneTimePaymentSession?`.

Open to other suggestions! Please correct me if I get it wrong!

HackTheW2d (approved):
LGTM! Thanks for all the work and helping addressing those comments

packages/paypal-js/types/v6/components/lpm-payments.d.ts
```diff
@@ -0,0 +1,166 @@
...
+ * `Extract`, and an empty extraction resolves to `{}` (a no-op intersection).
+ */
+export type LPMInstanceFor<T extends readonly string[]> = {
+  [C in Extract<
+    T[number],
+    LPMComponents
+  > as LPMComponentToSessionMethod[C]]?: (
+    options: LPMOneTimePaymentSessionOptions,
+  ) => LPMOneTimePaymentSession;
+};
```
EvanReinstein, line 166: We want to remove `?` b/c at runtime the component should always be loaded
```suggestion
export type LPMInstanceFor<T extends readonly string[]> = {
  [C in Extract<
    T[number],
    LPMComponents
  > as LPMComponentToSessionMethod[C]]: (
    options: LPMOneTimePaymentSessionOptions,
  ) => LPMOneTimePaymentSession;
};
```

packages/paypal-js/types/v6/tests/lpm-payments.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  });
+
+  const idealSession: LPMOneTimePaymentSession | undefined =
+    sdkInstance.createIdealOneTimePaymentSession?.({
+      onApprove: async () => undefined,
+    });
+
+  if (!idealSession) {
+    return;
+  }
```
EvanReinstein, line 39: You should be able to update this once you've made the change to `LPMInstanceFor` that I noted above.
```suggestion
  const idealSession: LPMOneTimePaymentSession | undefined =
    sdkInstance.createIdealOneTimePaymentSession.({
      onApprove: async () => undefined,
    });
```

EvanReinstein (approved):
This looks good @Mathanjeevika Thanks for your patience. I have 2 related comments that are outstanding, once those are addressed this can be merged.

Mathanjeevika:
@EvanReinstein  @HackTheW2d  All your comments were addressed Shall we merge the PR?

EvanReinstein:
> @EvanReinstein @HackTheW2d All your comments were addressed Shall we merge the PR?

I am seeing my comments regarding undefined components that are still not updated. Please make the changes discussed [here](https://github.com/paypal/paypal-js/pull/992#discussion_r3805759656) and [here](https://github.com/paypal/paypal-js/pull/992#discussion_r3805852786) FYI @Mathanjeevika

PR #993, opened 2026-07-13
feat(react-paypal-js): add v6 Local Payment Methods (LPM) React wrapper core

Mathanjeevika:
## Summary
- Adds a generic, registry-driven React wrapper for all 50 v6 SDK Local Payment Methods (LPM), exposed via a new `@paypal/react-paypal-js/sdk-v6/local-payment-methods` subpath so the default `@paypal/react-paypal-js` export is unaffected.
- Supports three usage patterns: an all-in-one named button (e.g. `BancontactOneTimePaymentButton`), a split hook + standalone button (`useBancontactOneTimePaymentSession` + `BancontactPaymentButton`), and a generic dynamic-selector button (`LPMOneTimePaymentButton`).
- New files: `src/v6/config/lpmRegistry.ts` (registry), `src/v6/hooks/useLPMOneTimePaymentSession.ts`, `src/v6/components/{LPMOneTimePaymentButton,LPMPaymentProvider}.tsx`, `src/v6/lpmExports.ts` (subpath entry), plus a new Rollup build target and `exports` subpath.

## Context
This is PR 2 of 3 split out from #926 per review feedback requesting the original monolithic PR be broken up:
1. #992 — `paypal-js` v6 LPM type declarations
2. **this PR** — `react-paypal-js` core LPM wrapper
3. (follow-up) — Storybook stories for LPM components

**This PR depends on #992 merging first.** The LPM types (`LPMComponents`, `LPMSessionMethodName`, `LPMOneTimePaymentSession`, etc.) are re-exported from `@paypal/paypal-js/sdk-v6` and consumed here via `src/v6/types/index.ts`'s `export type * from "@paypal/paypal-js/sdk-v6"`. Until #992 merges, `npm run typecheck` in this package will fail against `main` — this is expected and resolves once #992 lands (verified locally by overlaying #992's type declarations, which produces a clean `tsc --noEmit`).

## Known pre-existing issue
`src/v6/hooks/useLPMOneTimePaymentSession.test.ts` hangs under `jest` in this environment. This is **not** introduced by this PR — verified by reproducing the identical hang on the original, unmodified `feature/lpm-react-wrapper` branch running the same test file in isolation. Flagging for visibility; happy to dig into it further if it reproduces in CI.

## Test plan
- [x] `npx tsc --noEmit` passes once #992's types are present
- [x] `npx eslint` clean
- [ ] `npx jest` — blocked by the pre-existing hang noted above (other new LPM test files, e.g. `lpmRegistry.test.ts`, pass individually)
- [x] Changeset included (`.changeset/lpm-react-wrapper.md`)


EvanReinstein:
One thing we'll need to do is update the react-paypal-js README with integration code snippets and appropriate re-directs to integration resources (v6 sample integration repo or dev docs).

packages/react-paypal-js/src/v6/config/lpmRegistry.ts
```diff
@@ -0,0 +1,471 @@
...
+    buttonTag: "doku-button",
+    sessionMethod: "createDOKUOneTimePaymentSession",
+    displayName: "DOKU",
+    fields: ["name", "email"],
+    sessionFields: ["phone"],
+    testBuyerCountry: "ID",
+  },
+  estonia: {
+    component: "estoniabank-payments",
+    buttonTag: "estoniabank-button",
```
EvanReinstein:
```suggestion
    buttonTag: "estonia-button",
```

packages/react-paypal-js/src/v6/config/lpmRegistry.ts
```diff
@@ -0,0 +1,471 @@
...
+    component: "indomaret-payments",
+    buttonTag: "indomaret-button",
+    sessionMethod: "createIndomaretOneTimePaymentSession",
+    displayName: "Indomaret",
+    fields: ["name", "email"],
+    sessionFields: [],
+    testBuyerCountry: "ID",
+  },
+  thailandBanks: {
+    component: "thailandbanks-payments",
```
EvanReinstein:
```suggestion
    component: "thailand-banks-payments",
```

packages/react-paypal-js/src/v6/config/lpmRegistry.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    const TAG_EXCEPTIONS: Record<string, string> = {
+      "fiuu-cash-payments": "fiuu-button",
+    };
+    for (const config of Object.values(LPM_REGISTRY)) {
+      const expectedTag =
+        TAG_EXCEPTIONS[config.component] ??
+        config.component.replace("-payments", "-button");
+      expect(config.buttonTag).toBe(expectedTag);
+    }
+  });
```
EvanReinstein, line 46: This test is using the registry's own component value to compute the expected tag, which is why it didn't catch the estonia button bug that I pointed out elsewhere. This needs an independent list, like checking if config.buttonTag is a key in SDKWebComponents.

packages/react-paypal-js/src/v6/config/lpmRegistry.test.ts
```diff
@@ -0,0 +1,104 @@
...
+          expect(config.sessionFields).toContain("dateOfBirth");
+          expect(config.sessionFields).toContain("numberOfInstallments");
+        }
+      }
+    }
+  });
+
+  test("has exactly 50 LPM entries", () => {
+    expect(Object.keys(LPM_REGISTRY)).toHaveLength(50);
+  });
```
EvanReinstein: What is the purpose of this test?

packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Creates a stable SDK-iframe field component for a given field type.
+ *
+ * This is a module-level factory (not defined inside the hook render) so
+ * React concurrent-mode is fully safe and no ESLint suppression is required.
+ * The returned component is created once per `createEnhancedLPMHook` call via
+ * `useRef` and will not unmount/remount when the session changes — instead it
+ * uses a version counter driven by a pub/sub listener on `sessionListenersRef`.
+ */
```
EvanReinstein, line 176: It looks like this implementation goes against the decision from the ADR to create an LPMSessionContext in place of the pub/sub listener set. Can you review the ADR decisions and update the implementation here?

packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx
```diff
@@ -0,0 +1,272 @@
...
+      listeners.add(bumpVersion);
+      return () => {
+        listeners.delete(bumpVersion);
+      };
+    }, [bumpVersion]);
+
+    // Mount (or re-mount) the SDK iframe whenever the session version changes,
+    // i.e. whenever a new session becomes available via the pub/sub mechanism.
+    // sessionRef.current is accessed imperatively; fieldType is a stable
+    // closure variable from the factory — neither belongs in the deps array.
```
EvanReinstein: We decided against this pub/sub mechanism in favor of leveraging Context, please review the finalized ADR 👍

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -101,4 +101,111 @@ export type SDKWebComponents = {
...
+  "gopay-button": ButtonProps;
+  "alipay-button": ButtonProps;
+  "indomaret-button": ButtonProps;
+  "indonesiabanks-button": ButtonProps;
+  "kredivo-button": ButtonProps;
+  "linkaja-button": ButtonProps;
+  "ovo-button": ButtonProps;
+  "paysera-button": ButtonProps;
+  "skrill-button": ButtonProps;
+  "thailand-banks-button": ButtonProps;
```
HackTheW2d: Is this supposed to be `thailandbanks-button`? See internal implementation of this custom web components.

packages/paypal-js/types/v6/tests/lpm-payments.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  const startOptions: LPMStartOptions = { presentationMode: "popup" };
+
+  await idealSession.start(
+    startOptions,
+    Promise.resolve({ orderId: "ORDER-123", ...sessionFields }),
+  );
+
+  // LPM sessions do not support "auto" presentation mode.
+  // @ts-expect-error "auto" is not a valid LPM presentation mode
+  await idealSession.start({ presentationMode: "auto" });
```
HackTheW2d, line 55: https://github.com/paypal/paypal-js/pull/992#discussion_r3623195653
Same discussion here

packages/react-paypal-js/src/v6/lpmExports.test.tsx
```diff
@@ -0,0 +1,319 @@
...
+    const result = captureHookResult(() =>
+      useIdealOneTimePaymentSession({
+        presentationMode: "popup",
+        createOrder: jest.fn().mockResolvedValue({ orderId: "test" }),
+        onApprove: jest.fn().mockResolvedValue(undefined),
+      }),
+    );
+
+    expect(typeof result.NameField).toBe("function");
+    expect((result.NameField as unknown as React.FC).displayName).toBe("NameField");
```
HackTheW2d: Wonder if this is best practice since name field is unknown to typescript.
can we try:
```
// this converts "name" to "NameField"
type FieldComponents<Fields extends readonly LPMFieldType[]> = {
  [F in Fields[number] as `${Capitalize<F>}Field`]: React.FC<LPMFieldComponentProps>;
};

export type LPMEnhancedHookReturn<Fields extends readonly LPMFieldType[]> =
  LPMPaymentSessionReturn &
  { LPMSessionProvider: React.FC<{ children: React.ReactNode }> } &
  FieldComponents<Fields>;
```

```
export function createEnhancedLPMHook<const Fields extends readonly LPMFieldType[]>(
  lpm: LPMName, fieldTypes: Fields,
): (props: NamedHookProps) => LPMEnhancedHookReturn<Fields> { ... }
```

packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx
```diff
@@ -0,0 +1,314 @@
...
+ * export const IdealPaymentButton =
+ *   createLPMButtonComponent("ideal-button", "IdealPaymentButton");
+ *
+ * // Merchant usage:
+ * const { LPMSessionProvider } = useIdealOneTimePaymentSession({ ... });
+ * <LPMSessionProvider>
+ *   <IdealPaymentButton type="pay" />
+ * </LPMSessionProvider>
+ */
+export function createLPMButtonComponent(
```
HackTheW2d, line 137: Do we need to check `isHydrated` here?

packages/react-paypal-js/README.md
```diff
@@ -800,6 +800,121 @@ export default function App() {
 - `applePayConfig` is required and must be obtained from `useEligibleMethods()`
 - `onApprove` receives `ConfirmOrderResponse` — capture the order using `data.approveApplePayPayment.id`
 
+## Local Payment Methods (LPM)
+
+`react-paypal-js` ships pre-built components and hooks for 45+ regional Local Payment Methods (LPMs) — iDEAL, Bancontact, BLIK, Pix, Klarna, and more — each imported by name from the dedicated `@paypal/react-paypal-js/sdk-v6-lpm` subpath. LPMs are bundled separately from the core SDK to keep the main bundle small, since most integrations only need a handful of region-specific methods.
```
HackTheW2d: It seems that `sdk-v6-lpm` only appeared in README but mismatch the real path in `package.json`:
` ./sdk-v6/local-payment-methods`
Can we be consistent on this?

packages/react-paypal-js/src/v6/config/lpmRegistry.ts
```diff
@@ -0,0 +1,473 @@
...
+
+export const LPM_REGISTRY = {
+  ideal: {
+    component: "ideal-payments",
+    buttonTag: "ideal-button",
+    sessionMethod: "createIdealOneTimePaymentSession",
+    displayName: "iDEAL",
+    fields: ["name"],
+    sessionFields: [],
+    testBuyerCountry: "NL",
```
HackTheW2d, line 33: Should we move these `testBuyerCountry` to somewhere else? This will be bundled into package merchants consume. for eg, `__fixtures__/testBuyerCountry.ts: { ideal: "NL", bancontact: "BE", … }`

packages/react-paypal-js/src/v6/hooks/useLPMOneTimePaymentSession.ts
```diff
@@ -0,0 +1,197 @@
...
+            `Ensure "${config.component}" is included in the SDK components array.`,
+        ),
+      );
+      return;
+    }
+
+    const newSession = createPaymentSession({
+      sessionCreator: () =>
+        createSession({
+          orderId,
```
HackTheW2d, line 111: Seems like `orderId` is not used when creating session, it's set later on by promise. And used when `start()` is called. May I know why we pass `orderId` here? It adds one more dependency for this `useEffect` loop 🤔

packages/react-paypal-js/src/v6/components/LPMOneTimePaymentButton.tsx
```diff
@@ -0,0 +1,109 @@
...
+        const value = fieldValues?.[fieldType];
+        const fieldElement = session.createPaymentFields(
+          value !== undefined
+            ? { type: fieldType, value }
+            : { type: fieldType },
+        );
+        container.appendChild(fieldElement);
+      }
+    });
+  }, [session, config.fields, fieldValues]);
```
HackTheW2d, line 94: Please note that `fieldValues` is an object, and if merchants use inline like this:
`<IdealOneTimePaymentButton fieldValues={{ name: "John" }} />` This will cause re-render. Can we add more documentation to teach usage like:
```
const values = { name: "John" };
<IdealOneTimePaymentButton fieldValues={values} />
```
or we could have a new component to resolve this:
```
// src/v6/components/LPMPaymentField.tsx
export function LPMPaymentField({ session, type, value, style, className }: {
  session: LPMOneTimePaymentSession | null;
  type: LPMFieldType; value?: string;
  style?: React.CSSProperties; className?: string;
}) {
  const ref = useRef<HTMLDivElement>(null);
  useEffect(() => {
    const el = ref.current;
    if (!session?.createPaymentFields || !el) return;
    el.innerHTML = "";
    el.appendChild(session.createPaymentFields(value !== undefined ? { type, value } : { type }));
  }, [session, type, value]);   // these are all primitive vars
  return <div ref={ref} style={style} className={className} />;
}
```

and pass value like this:
` config.fields.map(f => <LPMPaymentField session={session} type={f} value={fieldValues?.[f]} />)`

packages/react-paypal-js/README.md
HackTheW2d: Can you please check comments here as well? @Mathanjeevika I could be wrong

HackTheW2d:
Can you please check why the build has failed? @Mathanjeevika

Mathanjeevika:
> Can you please check why the build has failed? @Mathanjeevika

@HackTheW2d  The build issue has been resolved now. Can you review the PR one more time.

packages/react-paypal-js/src/v6/index.ts
```diff
@@ -63,9 +63,68 @@ export {
   type ApplePayOneTimePaymentButtonProps,
 } from "./components/ApplePayOneTimePaymentButton";
 export {
-  GooglePayOneTimePaymentButton,
-  type GooglePayOneTimePaymentButtonProps,
-} from "./components/GooglePayOneTimePaymentButton";
```
HackTheW2d, line 68: This is a breaking change, please make sure these types exporting are not deleted.

packages/react-paypal-js/src/v6/index.ts
```diff
@@ -160,10 +217,109 @@ export {
   useApplePayOneTimePaymentSession,
   type UseApplePayOneTimePaymentSessionProps,
 } from "./hooks/useApplePayOneTimePaymentSession";
+// Factory-generated named LPM hooks
 export {
-  useGooglePayOneTimePaymentSession,
-  type UseGooglePayOneTimePaymentSessionProps,
-} from "./hooks/useGooglePayOneTimePaymentSession";
```
HackTheW2d, line 166: Same as above

packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx
```diff
@@ -0,0 +1,327 @@
...
+    containerClassName,
+    value,
+  }: LPMFieldComponentProps): JSX.Element {
+    const session = useContext(LPMSessionContext);
+    const containerRef = useRef<HTMLDivElement>(null);
+
+    React.useEffect(() => {
+      const container = containerRef.current;
+      const s = session;
+      if (!s?.createPaymentFields || !container) return;
```
EvanReinstein:
```suggestion
      if (!s?.createPaymentFields || !container) {
        return;
      }
```

packages/react-paypal-js/src/v6/lpmExports.test.tsx
```diff
@@ -0,0 +1,319 @@
...
+    // Capture hook result + render NameField wrapped in LPMSessionProvider
+    let capturedResult: ReturnType<typeof useIdealOneTimePaymentSession> | null = null;
+
+    function TestTree() {
+      capturedResult = useIdealOneTimePaymentSession({
+        presentationMode: "popup",
+        createOrder: jest.fn().mockResolvedValue({ orderId: "t" }),
+        onApprove: jest.fn().mockResolvedValue(undefined),
+      });
+      if (!capturedResult) return null;
```
EvanReinstein:
```suggestion
      if (!capturedResult) {
        return null;
      }
```

packages/react-paypal-js/MANUAL_VERIFICATION_LPM.md
EvanReinstein, line 1: I think the purpose of this markdown is really great, but we don't need to include this in the package. Rather it is something that I'd like to put into a confluence for our testing purposes. Please remove this file from the PR.

.changeset/lpm-add-remaining-16.md
EvanReinstein, line 1: There are 3 changesets currently. Can you consolidate that to one that covers the overall feature?

packages/react-paypal-js/src/v6/index.ts
```diff
@@ -63,9 +63,68 @@ export {
...
-  GooglePayOneTimePaymentButton,
-  type GooglePayOneTimePaymentButtonProps,
-} from "./components/GooglePayOneTimePaymentButton";
+  LPMOneTimePaymentButton,
+  type LPMOneTimePaymentButtonProps,
+} from "./components/LPMOneTimePaymentButton";
+export { LPM_REGISTRY } from "./config/lpmRegistry";
+export type { LPMName } from "./config/lpmRegistry";
+// Factory-generated named LPM button components and hooks (Approach 2)
+export {
```
EvanReinstein: I think that we can remove this list of exported LPMs here, these should only be exported via lpmExports.ts

packages/react-paypal-js/src/v6/index.ts
```diff
@@ -160,10 +217,109 @@ export {
   useApplePayOneTimePaymentSession,
   type UseApplePayOneTimePaymentSessionProps,
 } from "./hooks/useApplePayOneTimePaymentSession";
+// Factory-generated named LPM hooks
 export {
```
EvanReinstein, line 163: Same comment as above, all LPM exports should go through lpmExports.ts in order to adhere to the `/sdk-v6/local-payment-methods` import path restriction.

packages/react-paypal-js/src/v6/types/sdkWebComponents.ts
```diff
@@ -101,4 +101,155 @@ export type SDKWebComponents = {
   "paypal-basic-card-container": HTMLAttributes<HTMLElement>;
   "paypal-message": PayPalMessagesElement;
   "apple-pay-button": ApplePayButtonElementProps;
+  "ideal-button": ButtonProps;
```
EvanReinstein, line 104: It looks like 22 LPM button components were added to this list that do not yet exist in the SDK, can these be removed until after the related internal work has completed?

packages/react-paypal-js/src/v6/config/lpmRegistry.ts
```diff
@@ -0,0 +1,481 @@
...
+
+export type SessionFieldType =
+  | "phone"
+  | "billingAddress"
+  | "taxInfo"
+  | "expiryDate"
+  | "dateOfBirth"
+  | "numberOfInstallments"
+  | "bic"
+  | "identification";
```
EvanReinstein: These fields are not included in the LPMSessionFields type in paypal-js, see [here](https://github.com/paypal/paypal-js/blob/f791da2f598588476b196559eb74a5870458ecdd/packages/paypal-js/types/v6/components/lpm-payments.d.ts#L37-L49). Are these being added as field types to the SDK in the future? If possible I'd like to make sure that this PR matches what is currently in the SDK and is released to production.

packages/react-paypal-js/src/v6/lpmExports.ts
```diff
@@ -0,0 +1,333 @@
...
+export type UseThailandBanksOneTimePaymentSessionProps = NamedLPMHookProps;
+export type UseAlfamartOneTimePaymentSessionProps = NamedLPMHookProps;
+export type UseZipOneTimePaymentSessionProps = NamedLPMHookProps;
+export type UseLatviaBanksOneTimePaymentSessionProps = NamedLPMHookProps;
+export type UseFiuuOneTimePaymentSessionProps = NamedLPMHookProps;
+export type UseLithuaniaBanksOneTimePaymentSessionProps = NamedLPMHookProps;
+
+// ─── Named standalone button components ───────────────────────────────────────
+// Accept a `paymentSession` prop (return value of the corresponding named hook).
+// Can be placed anywhere in the layout — no Provider wrapping required.
```
EvanReinstein: This is a stale comment describing the old pub/sub design
```suggestion
// Read session state from LPMSessionHandleContext — render inside the
// LPMSessionProvider returned by the corresponding use*OneTimePaymentSession hook.
```

package-lock.json
EvanReinstein, line 1: It looks like there was an incidental `npm install` rather than `npm ci` Can you revert this package-lock change please?

packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx
EvanReinstein: @Mathanjeevika Please update the comment to accurately reflect the strategy chosen in our internal ADR 👍

packages/react-paypal-js/src/v6/lpmExports.ts
```diff
@@ -0,0 +1,333 @@
...
+
+// Provider + context utilities re-exported from this subpath. LPM consumers MUST
+// obtain PayPalProvider/usePayPal/INSTANCE_LOADING_STATE from here (not from
+// ./sdk-v6): the LPM subpath is a separate bundle with its own React context
+// instance, so a PayPalProvider imported from the main ./sdk-v6 bundle would not
+// satisfy the LPM hooks' usePayPal, throwing "usePayPal must be used within a
+// PayPalProvider".
+export { PayPalProvider } from "./components/PayPalProvider";
+export { usePayPal } from "./hooks/usePayPal";
+export { INSTANCE_LOADING_STATE } from "./types/ProviderEnums";
```
EvanReinstein, line 213: Ok I think this is worth discussing further. Technically this is completely correct due to the sub-path strategy we are taking. The LPMs depend on createContext and if they are paired with a `PayPalProvider` that is imported from outside the LPM subpath, the contexts won't be shared correctly due to being separate bundles.

```
import { PayPalProvider, usePayPal } from "@paypal/react-paypal-js/sdk-v6"
import {useThailandBanksOneTimePaymentSession} from "@paypal/react-paypal-js/sdk-v6/local-payment-methods"

usePayPal --> /sdk-v6 bundle context
useThailandBanksOTPSession --> needs /sdk-v6/local-payment-methods bundle context
```

To further complicate this, I already left a handful of comments in `index.ts` to clean up/remove all the LPM-related exports that are being done there. This is in order to adhere to the LPM sub-path strategy and keep these exports out of the main bundle. This creates a problem because the only way to use PayPal and LPM buttons together would be to ensure they are imported through the same bundle and can share the same context.

@HackTheW2d I don't think we need to resolve this for this PR specifically. But we'll need a fast follow before this is published.

packages/react-paypal-js/src/v6/lpmExports.ts
HackTheW2d: This [PR](https://github.com/paypal-examples/v6-web-sdk-sample-integration/pull/274) needs to be fixed as well. It still uses `paymentSession`

packages/react-paypal-js/src/v6/components/LPMOneTimePaymentButton.tsx
EvanReinstein: It looks like documentation was added to the button prop types to teach passing a stable ref.

packages/react-paypal-js/src/v6/components/LPMPaymentProvider.tsx
EvanReinstein: Done

EvanReinstein (approved):
This looks good to me 🚀

PR #1013, opened 2026-07-24
Update server side eligibility hydration example

EvanReinstein:
This PR updates the README with more details for making an eligibility request server side using the v6 server-side eligibility hook.

This PR also updates the condition to skip fetching eligibility if the provider already has eligibility response from the SDK (and the eligibility payload is unchanged).


nityasp (approved):
Looks good.

PR #1014, opened 2026-07-27
Fix: v6 Server Eligibility Method 

EvanReinstein:
The v6 server-side eligibility function is exported as useFetchEligibleMethods, but it's not a React hook — it's a server-only async function called with await in server components / loaders. The use prefix falsely trips eslint-plugin-react-hooks (rules-of-hooks / no-unnecessary-use-prefix) in consumer projects, surfacing bogus "React Hook cannot be called in an async function / conditionally" errors.

This PR renames the function and fixes two adjacent naming/placement issues, all non-breaking via @deprecated aliases:

- useFetchEligibleMethods --> fetchEligibleMethods, moved out of hooks/ into src/v6/server/
- Client hook types UseFetchEligibleMethodsOptions / UseFetchEligibleMethodsResult --> UseEligibleMethodsOptions / UseEligibleMethodsResult (matches useEligibleMethods)
- All previous names remain exported as @deprecated aliases (removed in next major)

### DIFF
--- a/.changeset/rename-fetch-eligible-methods.md
+++ b/.changeset/rename-fetch-eligible-methods.md
@@ -1,0 +1,15 @@
+---
+"@paypal/react-paypal-js": minor
+---
+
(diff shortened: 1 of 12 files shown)


nbierdeman (approved):

nityasp (approved):

HackTheW2d (approved):
Looks good! Thanks for helping addressing this issue!

PR #1016, opened 2026-07-28
Prevent redundant client-side fetch when hydrating eligible methods

nityasp:
**What caused the issue?**

PayPalProvider's hydration effect currently protects against a duplicate fetch `deepEqual` dedup check in `useEligibleMethods` that only helps after `eligiblePaymentMethods` is already in context, it does nothing on the very first render, before hydration has happened. So even when `eligibleMethodsResponse` was passed to `PayPalProvider`, the child had no signal that hydration was in flight, saw an empty context, and fired its own client-side request anyway, defeating the purpose of prefetching server-side.

**To reproduce the issue:**

Run the `nextjs-paypal-template` draft [PR](https://github.com/paypal-examples/nextjs-paypal-template/pull/1) with the latest `react-paypal-js` version.
image

**The fix:**
1. `PayPalProviderContext.tsx` / `PayPalProvider.tsx`: added an `isEligibilityHydrationPending` flag to context (true whenever `eligibleMethodsResponse` is set but hasn't hydrated into state yet), giving children a way to know hydration is in progress. Also widened `eligibleMethodsResponse` to accept a `Promise`, and made the hydration effect await it before dispatching.
2. `useEligibleMethods.ts`: added a guard that skips the fetch while `isEligibilityHydrationPending` is true but only for calls with no explicit payload, since hydrated data is always stored with `payload: null` and can never satisfy a payload-specific call anyway.

**To test this fix:**
1. Build the react-paypal-js library with this fix.
2. Link it into the `nextjs-paypal-template` draft [PR](https://github.com/paypal-examples/nextjs-paypal-template/pull/1)
3. Run the `nextjs-paypal-template` app (`npm run dev`)
image


package-lock.json
EvanReinstein, line 1: Looks like an `npm install` rather than `npm ci` here. FYI I just updated my local nvm default to `22.22.3` in accordance with other repo updates.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.test.ts
EvanReinstein, line 1: Should we have a test that checks that no request is made when `isEligibilityHydrationPending` is true?

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -31,7 +31,9 @@ type PayPalProviderPropsBase = Omit<
 > &
   Omit<LoadCoreSdkScriptOptions, "dataSdkIntegrationSource"> & {
     components?: Components[];
-    eligibleMethodsResponse?: FindEligiblePaymentMethodsResponse;
+    eligibleMethodsResponse?:
+      | FindEligiblePaymentMethodsResponse
+      | Promise<FindEligiblePaymentMethodsResponse>;
```
EvanReinstein: This is interesting. Accepting a promise is a pattern that we use for the client token and ID, but I think that we should align with the SDK's expected type here, which is just a promise. I'd rather not deviate from what the SDK does/expects unless we truly need to.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -331,29 +333,49 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
 
   useEffect(() => {
     const sdkInstance = state.sdkInstance;
-    if (!sdkInstance) {
+    if (!sdkInstance || !eligibleMethodsResponse) {
```
EvanReinstein: I think that this is a type check at its core, but I think also that this will do an early return in the case that the eligibility API response is not passed to the provider, right? So even a truthy SDK instance without a response would do an early return.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -331,29 +333,49 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+    hydrate();
+
+    return () => {
+      isSubscribed = false;
+    };
   }, [state.sdkInstance, eligibleMethodsResponse, setError]);
 
+  // True until eligibleMethodsResponse has hydrated into state.
+  const isEligibilityHydrationPending =
+    !!eligibleMethodsResponse && !state.eligiblePaymentMethods && !state.error;
```
EvanReinstein: What if a response is never passed? If hydration is perpetually true, then won't that effect the ability of the hook to make a request?

Based on the condition in `useEligibleMethods` an undefined payload will prevent a request, but I think that technically you can call the eligibility API without a payload 🤔

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: Doing a more thorough review I think that not going with the Promise pattern will also clean up a potential race condition that comes about from a weird integration:
- api response is passed to provider AND a child component calls the eligibility hook w/ or w/o a payload.

technically we wouldn't recommend this type of integration but narrowing the type of API response the provider can accept will remove this issue.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: I'm wondering if using a Reducer based strategy that explicitly knows when eligibility hydration is occurring would be the right move instead of this inferred eligibility hydration that is effected by child component eligibility hook calls.

packages/react-paypal-js/src/v6/hooks/useEligibleMethods.test.ts
HackTheW2d: Should we test some core fix functionality?
eg.
```

describe("hydration pending guard (isEligibilityHydrationPending)", () => {
  function createSdk() {
    return {
      findEligibleMethods: jest.fn().mockResolvedValue({}),
    } as unknown as PayPalState["sdkInstance"];
  }

  test("does NOT fetch when hydration is pending and no payload is given", () => {
    const sdkInstance = createSdk();

    renderHook(() => useEligibleMethods(), {
      wrapper: createWrapper({
        sdkInstance,
        eligiblePaymentMethods: null,
        isEligibilityHydrationPending: true,
      }),
    });

    expect(sdkInstance.findEligibleMethods).not.toHaveBeenCalled();
  });

  test("still fetches immediately when a payload is given, even while hydration is pending", () => {
    const sdkInstance = createSdk();

    renderHook(
      () =>
        useEligibleMethods({
          payload: { purchase_units: [{ amount: { currency_code: "USD" } }] },
        }),
      {
        wrapper: createWrapper({
          sdkInstance,
          eligiblePaymentMethods: null,
          isEligibilityHydrationPending: true,
        }),
      },
    );

    expect(sdkInstance.findEligibleMethods).toHaveBeenCalledTimes(1);
  });

  test("fetches once the pending flag flips to false without new data arriving", () => {
    const sdkInstance = createSdk();
    const dispatch = jest.fn();

    function DynamicWrapper({
      isEligibilityHydrationPending,
      children,
    }: {
      isEligibilityHydrationPending: boolean;
      children: React.ReactNode;
    }) {
      const fullContext: PayPalState = {
        sdkInstance,
        eligiblePaymentMethods: null,
        eligiblePaymentMethodsPayload: null,
        loadingStatus: INSTANCE_LOADING_STATE.RESOLVED,
        error: null,
        isHydrated: true,
        isEligibilityHydrationPending,
      };
      return React.createElement(
        PayPalDispatchContext.Provider,
        { value: dispatch },
        React.createElement(PayPalContext.Provider, { value: fullContext }, children),
      );
    }

    const { rerender } = renderHook(() => useEligibleMethods(), {
      wrapper: DynamicWrapper,
      initialProps: { isEligibilityHydrationPending: true },
    });

    expect(sdkInstance.findEligibleMethods).not.toHaveBeenCalled();

    rerender({ isEligibilityHydrationPending: false });

    expect(sdkInstance.findEligibleMethods).toHaveBeenCalledTimes(1);
  });
});

```

packages/react-paypal-js/src/v6/context/PayPalProviderContext.test.ts
HackTheW2d, line 1: Should we include more cover tests here? since we change response type from object to include promise

```
test("hydrates eligibility from an unresolved Promise passed as eligibleMethodsResponse", async () => {
  let resolvePromise!: (value: typeof TEST_ELIGIBILITY_HOOK_RESULT) => void;
  const eligibleMethodsResponse = new Promise<typeof TEST_ELIGIBILITY_HOOK_RESULT>(
    (resolve) => {
      resolvePromise = resolve;
    },
  );

  const { state } = await renderProvider({ eligibleMethodsResponse });

  // Promise has not been resolved, hydration must stay pending。
  expect(state.isEligibilityHydrationPending).toBe(true);
  expect(state.eligiblePaymentMethods).toBe(null);

  await act(async () => {
    resolvePromise(TEST_ELIGIBILITY_HOOK_RESULT);
    await Promise.resolve();
  });

  await waitFor(() => {
    expect(state.eligiblePaymentMethods).toEqual(TEST_SDK_ELIGIBILITY_RESULT);
    expect(state.isEligibilityHydrationPending).toBe(false);
  });
});

test("surfaces an error when the eligibleMethodsResponse Promise rejects", async () => {
  const rejection = new Error("upstream eligibility fetch failed");

  const { state } = await renderProvider({
    eligibleMethodsResponse: Promise.reject(rejection),
  });

  await waitFor(() => expectRejectedState(state, rejection));
});

```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
HackTheW2d: What if a promise never gets resolved? Should we set a hydration timeout. If the promise resolves before timeout, we hydrate as usual. But what if timeout has reached, we should make the find-eligible-methods call. What do you think?

eg.


```
const [hydrationTimedOut, setHydrationTimedOut] = useState(false);

useEffect(() => {
  const sdkInstance = state.sdkInstance;
  if (!sdkInstance || !eligibleMethodsResponse) {
    return;
  }

  let isSubscribed = true;
  setHydrationTimedOut(false); // a new instance/promise, a new timer

  const HYDRATION_TIMEOUT_MS = 10_000;
  const timedOut = Symbol("hydration-timeout");

  const hydrate = async () => {
    try {
      const result = await Promise.race([
        eligibleMethodsResponse,
        new Promise<typeof timedOut>((resolve) =>
          setTimeout(() => resolve(timedOut), HYDRATION_TIMEOUT_MS),
        ),
      ]);

      if (!isSubscribed) {
        return;
      }

      if (result === timedOut) {
        // timeout has reached
        // let's make a call from client, rather than keep waiting
        setHydrationTimedOut(true);
        return;
      }

      const eligiblePaymentMethods = sdkInstance.hydrateEligibleMethods(result);
      dispatch({
        type: INSTANCE_DISPATCH_ACTION.SET_ELIGIBILITY,
        value: { eligiblePaymentMethods, payload: null },
      });
    } catch (error) {
      if (!isSubscribed) return;
      setError(error);
      dispatch({ type: INSTANCE_DISPATCH_ACTION.SET_ERROR, value: toError(error) });
    }
  };

  hydrate();

  return () => {
    isSubscribed = false;
  };
}, [state.sdkInstance, eligibleMethodsResponse, setError]);

const isEligibilityHydrationPending =
  !!eligibleMethodsResponse &&
  !state.eligiblePaymentMethods &&
  !state.error &&
  !hydrationTimedOut;
```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein: On second glance I misunderstood this early return. It is scoped to the hydration useEffect so there won't be any issue with the situation i described above 👍🏻

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -354,6 +352,21 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
   }, [state.sdkInstance, eligibleMethodsResponse, setError]);
 
+  // Dispatched during render (not in an effect) so children see the
+  // correct status on the same render, before their own effects run.
+  let eligibilityHydrationStatus = INSTANCE_LOADING_STATE.RESOLVED;
+  if (eligibleMethodsResponse && !state.eligiblePaymentMethods) {
+    eligibilityHydrationStatus = state.error
+      ? INSTANCE_LOADING_STATE.REJECTED
+      : INSTANCE_LOADING_STATE.PENDING;
+  }
```
EvanReinstein: Here we are still tying eligibility status to whether or not state.eligiblePaymentMethods exists and inferring the pending status. Rather we should add a hydration only boolean to the reducer state that only flips via the success path after hydration succeeds (see above).

```
// see state.isEligibilityHydrated
let eligibilityHydrationStatus = INSTANCE_LOADING_STATE.RESOLVED;
if (eligibleMethodsResponse && !state.isEligibilityHydrated && !state.error) {
  eligibilityHydrationStatus = INSTANCE_LOADING_STATE.PENDING;
} else if (eligibleMethodsResponse && state.error) {
  eligibilityHydrationStatus = INSTANCE_LOADING_STATE.REJECTED;
}
```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -354,6 +352,21 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
+  if (eligibilityHydrationStatus !== state.eligibilityHydrationStatus) {
+    dispatch({
+      type: INSTANCE_DISPATCH_ACTION.SET_ELIGIBILITY_HYDRATION_STATUS,
+      value: eligibilityHydrationStatus,
+    });
```
HackTheW2d, line 370: It seems like `state.eligibilityHydrationStatus` is only used here to compare with its last value. Can we possibly apply the same pattern like what we have in `eligiblePaymentMethodsRef`?
eg,.
```
const prevEligibilityHydrationStatusRef = useRef(eligibilityHydrationStatus);

let eligibilityHydrationStatus = INSTANCE_LOADING_STATE.RESOLVED;
if (eligibleMethodsResponse && !state.eligiblePaymentMethods) {
  eligibilityHydrationStatus = state.error
    ? INSTANCE_LOADING_STATE.REJECTED
    : INSTANCE_LOADING_STATE.PENDING;
}
prevEligibilityHydrationStatusRef.current = eligibilityHydrationStatus;
```

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
EvanReinstein, line 168: I'm realizing that we probably should memoize or put this in a stable ref for merchants, otherwise it will trigger a re-render.

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
```diff
@@ -331,20 +331,18 @@ export const PayPalProvider: React.FC<PayPalProviderProps> = ({
...
-          value: { eligiblePaymentMethods, payload: null },
-        });
-      }
+      const eligiblePaymentMethods = sdkInstance.hydrateEligibleMethods(
+        eligibleMethodsResponse,
+      );
+      dispatch({
+        type: INSTANCE_DISPATCH_ACTION.SET_ELIGIBILITY,
+        value: { eligiblePaymentMethods, payload: null },
+      });
```
EvanReinstein, line 348: Following up on my comment below, we should leverage a HYDRATE_ELIGIBILITY action rather than re-use the current.

```
const eligiblePaymentMethods = sdkInstance.hydrateEligibleMethods(
  eligibleMethodsResponse,
);
dispatch({
  type: INSTANCE_DISPATCH_ACTION.HYDRATE_ELIGIBILITY,
  value: { eligiblePaymentMethods },
});

// reducer case here:
case INSTANCE_DISPATCH_ACTION.HYDRATE_ELIGIBILITY:
  return {
    ...state,
    eligiblePaymentMethods: action.value.eligiblePaymentMethods,
    eligiblePaymentMethodsPayload: null,
    isEligibilityHydrated: true,
  };
```

This keeps all the logic for hydration isolated in the reducer as well

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
nityasp: `eligiblePaymentMethodsRef` is only read later in an effect, so a repeat render just overwrites it. Here this check decides whether to dispatch right now, so a stale ref could cause the real render to skip the update. So comparing against state avoid that.
Also without the dispatch call, `state.eligibilityHydrationStatus` never actually gets updated.

HackTheW2d (approved):
Brilliant work on fixing this @nityasp 
Thanks for addressing this issue! 🚀

packages/react-paypal-js/src/v6/components/PayPalProvider.tsx
nityasp: nice catch!!

HackTheW2d (approved):

package-lock.json
EvanReinstein: @nityasp Please don't forget to revert this change before merge

EvanReinstein (approved):
This LGTM, thanks @nityasp !

package-lock.json
nityasp: reverted

package-lock.json
```diff
@@ -27077,7 +27077,7 @@
     },
     "packages/paypal-js": {
       "name": "@paypal/paypal-js",
-      "version": "10.1.0",
+      "version": "10.0.3",
```
EvanReinstein: Now i'm seeing more, different churn

PR #1031, opened 2026-08-19
feat: v6 Google Pay 3DS

### LINKED ISSUE #952: [Feature] Implement 3DS authentication flow for Googlepay
### 🚀 Feature Proposal

Complete the Googlepay flow and implement the 3DS auth flow.
Right now the method `initiatePayerAction` of `GooglePayOneTimePaymentSession` is just a placeholder.
From https://github.com/paypal/paypal-js/blob/6175fd3a751e38e6677d6a27ef14b39427592d02/packages/paypal-js/types/v6/components/googlepay-payments.d.ts#L375

```typescript
...
  /**
   * Initiates 3DS authentication flow (placeholder)
   *
   * @remarks
   * This method is currently a placeholder for future 3DS support.
   *
   * @internal
   */
  initiatePayerAction(): void;
...
```

### Motivation

3DS is used a lot, having a payment type that does not implement it make the payment itself unusable in production.

### Example

Something like:

```typescript
 const response = await session.confirmOrder({
   orderId: "ORDER-ID",
   paymentMethodData: paymentData.paymentMethodData,
 }); 

 if (response.status === "PAYER_ACTION_REQUIRED") {
   await session.initiatePayerAction({orderId: "ORDER-ID"});
   // Check the payment liability shift using a server endpoint
   // and proceed accordingly
 }
```

EvanReinstein:
## Add 3D Secure (SCA) support to Google Pay

Closes #952

`GooglePayOneTimePaymentSession.initiatePayerAction` was a no-op placeholder, which left Google Pay unusable for any order requiring 3DS in production. This implements the real payer-action flow end-to-end, matching the API shape proposed in the issue.

### `@paypal/paypal-js` — types
- Replace the placeholder `initiatePayerAction(): void` with its real signature: `initiatePayerAction(options: { orderId: string }): Promise<InitiatePayerActionResponse>` — resolves with the 3DS `liabilityShift`, rejects on buyer cancel or auth failure.
- Add and export `LiabilityShiftType` (`"UNKNOWN" | "NO" | "YES" | "POSSIBLE"`), `InitiatePayerActionOptions`, and `InitiatePayerActionResponse`.
- Update the v6 type-test to exercise the new signature.

### `@paypal/react-paypal-js` — hook & component
- `useGooglePayOneTimePaymentSession` now drives the full 3DS flow. When `confirmOrder` returns `PAYER_ACTION_REQUIRED`, it launches `initiatePayerAction({ orderId })` automatically.
- **onApprove fires only after authentication succeeds**, so merchants never capture an unauthenticated order. Its data now carries the resulting `liabilityShift` (new `GooglePayOnApproveData` export).
- Buyer cancel / authentication failure is routed to `onError` (the core SDK rejects both with the same error).
- 3DS is **not** awaited inside Google's `onPaymentAuthorized` callback: the Google Pay payment sheet must close before the 3DS modal can open, so the callback returns `SUCCESS` immediately and the flow completes out-of-band (guarded against post-unmount callbacks).

### Tests
- Rewrote the `PAYER_ACTION_REQUIRED` test to assert the `{ orderId }` call and out-of-band `onApprove` with `liabilityShift`.
- Added coverage for the reject → `onError` path and the post-unmount no-op guard.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/useGooglePayOneTimePaymentSession.ts
+++ b/packages/react-paypal-js/src/v6/hooks/useGooglePayOneTimePaymentSession.ts
@@ -14,8 +14,19 @@ import type {
   GooglePayButtonOptions,
   GooglePayApprovePaymentResponse,
   GooglePayTransactionInfo,
+  LiabilityShiftType,
   BasePaymentSessionReturn,
 } from "../types";
+
+/**
+ * Data passed to `onApprove`. When the order required 3DS
+ * (`PAYER_ACTION_REQUIRED`), this fires only after authentication succeeds and
+ * includes the resulting `liabilityShift`.
+ */
+export type GooglePayOnApproveData = GooglePayApprovePaymentResponse & {
+  /** The 3DS liability-shift outcome, present only when 3DS ran. */
+  liabilityShift?: LiabilityShiftType;
+};
 
 export type UseGooglePayOneTimePaymentSessionProps = {
   /**
@@ -39,8 +50,11 @@ export type UseGooglePayOneTimePaymentSessionProps = {
   createOrder: () => Promise<{ orderId: string }>;
   /**
    * Callback invoked when the payment is successfully approved.
+   *
+   * When 3DS (`PAYER_ACTION_REQUIRED`) is required, this fires only after the
+   * buyer completes authentication, and the data includes `liabilityShift`.
    */
-  onApprove: (data: GooglePayApprovePaymentResponse) => void | Promise<void>;
+  onApprove: (data: GooglePayOnApproveData) => void | Promise<void>;
   /**
    * Optional callback invoked when the payment is cancelled.
    */
@@ -91,6 +105,12 @@ export type UseGooglePayOneTimePaymentSessionReturn =
  * The hook manages the entire session lifecycle including order creation, payment confirmation,
  * 3DS (PAYER_ACTION_REQUIRED) handling, and error management.
  *
+ * When an order requires 3DS, the hook launches the payer-action (SCA) flow automatically and
+ * calls `onApprove` only after the buyer completes authentication — so `onApprove` (which
+ * merchants typically use to capture) never fires for an unauthenticated order. On that path the
+ * `onApprove` data also includes `liabilityShift`. If the buyer cancels or authentication fails,
+ * `onError` is called instead.
+ *
  * @example
  * ```typescript
  * function GooglePayCheckoutButton() {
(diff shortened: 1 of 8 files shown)


packages/react-paypal-js/src/v6/hooks/useGooglePayOneTimePaymentSession.ts
```diff
@@ -261,10 +284,41 @@ export function useGooglePayOneTimePaymentSession({
...
-                paypalSession.initiatePayerAction();
+                // Do NOT await here: the 3DS modal can only open once Google's
+                // payment sheet closes, and that sheet stays open until this
+                // callback resolves. So we kick off the 3DS flow, return SUCCESS
+                // to close the sheet, and notify the merchant out-of-band once
+                // authentication resolves. onApprove fires only on success, so
+                // merchants never capture an unauthenticated order.
+                paypalSession
+                  .initiatePayerAction({ orderId: order.orderId })
+                  .then((payerActionResult) => {
```
HackTheW2d, line 316: Here we think `onApprove` and `captureOrder` can be called when the promise is **resolved** rather than checking the result. 
And do we want to leave some options for merchants to decide if they want to continue approving by providing them full context of `enrollment_status`, `authentication_status` and `liability_shift`?

HackTheW2d: Check https://developer.paypal.com/platforms/checkout/advanced/customize/3d-secure/response-parameters for more context
```
 {
    "enrollment_status": "Y",
    "authentication_status": "N",
    "liability_shift": "NO"
  }
```
is a solid reject and should not continue with capturing.

But 
```
 {
    "enrollment_status": "N",
    "authentication_status": null,
    "liability_shift": "NO"
  }
```
Can be decided by merchants if they want to continue.

HackTheW2d: I think `state: "...",` is discarded. We only check liability_shift here 🤔

EvanReinstein: These are some good questions, but basically the SDK only passes `liability_shift` back to the consumer, so we can only surface what the SDK gives us. As for whether or not to call onApprove (a stand in for a capture call in this use case), we want to mirror what the vanilla integration is suggesting which is to fire the merchant's capture call unconditionally.

HackTheW2d (approved):
LGTM! Nice work!

PR #1034, opened 2026-08-25
Fix: Apple Pay Global Type

EvanReinstein:
This PR addresses a Typescript clash between `@types/applepayjs` and the v6 Apple Pay component types in `paypal-js`. The JS SDK and its associated wrappers should not handle types for other payment handlers that are not strictly owned/governed by PayPal. In this case the global `ApplePaySession` has been removed and replaced with `@types/applepayjs` as a devdependency, with instructions for integrations that rely on the Apple Pay window object to leverage the same DefinitelyTyped package.

### DIFF
--- a/.changeset/loose-plums-repair.md
+++ b/.changeset/loose-plums-repair.md
@@ -1,0 +1,7 @@
+---
+"@paypal/paypal-js": minor
+---
+
+Remove PayPal's homegrown `ApplePaySession` browser-global typing from the v6 types.
+
+`@paypal/paypal-js/sdk-v6` no longer declares a stripped `ApplePaySession` class or augments the global `Window` interface with `ApplePaySession`. That augmentation conflicted with the community `@types/applepayjs` package (`TS2717` / `TS2687`), and because it shipped through the `sdk-v6` barrel it landed in every consumer's compilation — even projects that never use Apple Pay.
--- a/.changeset/loose-plums-repair.md
+++ b/.changeset/loose-plums-repair.md
@@ -1,7 +1,7 @@
 ---
-"@paypal/paypal-js": minor
+"@paypal/paypal-js": major
 ---
 
-Remove PayPal's homegrown `ApplePaySession` browser-global typing from the v6 types.
+A breaking change that removes PayPal's homegrown `ApplePaySession` browser-global typing from the v6 types.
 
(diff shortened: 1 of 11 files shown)


packages/paypal-js/types/v6/components/applepay-payments.d.ts
```diff
@@ -230,7 +235,7 @@ export interface ApplePayPaymentsInstance {
    * ```typescript
    * // Check if Apple Pay is available
    * const isApplePayAvailable =
-   *   window.ApplePaySession && ApplePaySession.canMakePayments();
+   *   typeof ApplePaySession !== "undefined" && ApplePaySession.canMakePayments();
```
gregjopa: It would be good to keep `window` to know where this `ApplePaySession` comes from?
```suggestion
   *   typeof window.ApplePaySession !== "undefined" && ApplePaySession.canMakePayments();
```

.changeset/loose-plums-repair.md
```diff
@@ -0,0 +1,7 @@
+---
+"@paypal/paypal-js": minor
```
HackTheW2d: Do you mind update this to a major change? Any merchants who used this before would be notified they'll need to install applepayjs package

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
```diff
@@ -329,7 +327,10 @@ export function useApplePayOneTimePaymentSession({
           const confirmResult = await paypalSession.confirmOrder({
             orderId: order.orderId,
             token: event.payment.token,
-            billingContact: event.payment.billingContact,
+            // Apple types billingContact as optional, but PayPal's SDK requires it to
+            // proxy the order to Apple Pay; the merchant requests billing fields, so it
+            // is present at runtime.
+            billingContact: event.payment.billingContact as ApplePayContact,
```
HackTheW2d, line 333: Can we add tests of missing `billingContact`? This should be validated its existence before passing to `confirmOrder`

gregjopa (approved):

packages/react-paypal-js/src/v6/hooks/useApplePayOneTimePaymentSession.ts
EvanReinstein: Yea looking more closely at the SDK it looks like we just pass these values along. I'd like to investigate more and surface the errors that get thrown when it isn't passed

Issue #1020, opened 2026-07-29
[Bug] PayPalOneTimePaymentSessionOptions makes `onApprove` required despite declaring it optional


MaxBag:
`onApprove` is declared as optional in `PayPalOneTimePaymentSessionOptions`, but it is still required in practice.
                                                                                                                                                                                                  
The type intersects `BasePaymentSessionOptions` — where `onApprove` is required — and re-declares `onApprove?` without `Omit`-ing it from the base first. In a TypeScript intersection a  property is optional only if it is optional in **every** constituent, so the `?` has no effect.                                                                                                 
                                                                                                                                                                                                  
  As a result, neither `<PayPalOneTimePaymentButton>` nor `usePayPalOneTimePaymentSession()` can be used without `onApprove`, and the documented examples do not compile.                         
                                                                                                                                                                                                  
  ### Affected declaration

  `packages/paypal-js/types/v6/components/paypal-payments.d.ts:82-93`                                                                                                                             
                                                                                                                                                                                                  
  ```ts                                                                                                                                                                                           
  export type PayPalOneTimePaymentSessionOptions = BasePaymentSessionOptions & {                                                                                                                  
    commit?: boolean;                                                                                                                                                                             
    orderId?: string;                                                                                                                                                                             
    onApprove?: (data: OnApproveDataOneTimePayments) => Promise<void>; // ← no effect                                                                                                             
    // ...                                                                                                                                                                                        
  };                                                                                                                                                                                              
  ```                                                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                        
  ### Reproduction                                                                                                                                                                                
                                                                                                                                                                                                  
  ```tsx                                                                                                                                                                                          
  import {PayPalProvider, PayPalOneTimePaymentButton} from '@paypal/react-paypal-js/sdk-v6';                                                                                                      
                                                                                                                                                                                                  
  const Checkout = ({clientToken}: {clientToken: string}) => {                                                                                                                                    
    const createOrder = () => Promise.resolve({orderId: 'ORDER-123'});                                                                                                                            
                                                                                                                                                                                                  
    return (                                                                                                                                                                                      
      <PayPalProvider clientToken={clientToken} environment="sandbox" pageType="checkout">                                                                                                        
        <PayPalOneTimePaymentButton createOrder={createOrder} />                                                                                                                                  
      </PayPalProvider>                                                                                                                                                                           
    );                                                                                                                                                                                            
  };                                                                                                                                                                                              
[...]

PR #1037, opened 2026-08-26
Fix: PayPal One Time Payment Session onApprove property

EvanReinstein:
This addresses the issue raised [here](https://github.com/paypal/paypal-js/issues/1020).


packages/paypal-js/types/v6/components/paypal-payments.d.ts
```diff
@@ -82,7 +81,6 @@ export type OnCancelDataSavePayments = {
 export type PayPalOneTimePaymentSessionOptions = BasePaymentSessionOptions & {
   commit?: boolean;
   orderId?: string;
-  onApprove?: (data: OnApproveDataOneTimePayments) => Promise<void>;
```
EvanReinstein, line 85: Removing this allows `PayPalOneTimePaymentSessionOptions` to fall back to the correct `onApprove` type definition in `BasePaymentSessionOptions`

gregjopa: One thing to note is that `onApprove()` is not required when using `presentationMode: "redirect"`. It is required for all other presentation modes.

HackTheW2d: Is this the right case: If `onApprove` does not exist, only allow `redirect`, only if `onApprove` presents, all the presentation modes can be selected.

packages/react-paypal-js/README.md
```diff
@@ -296,7 +296,14 @@ function CheckoutPage() {
     );
   }
 
-  return <PayPalOneTimePaymentButton orderId="ORDER-123" />;
+  return (
+    <PayPalOneTimePaymentButton
+      orderId="ORDER-123"
+      onApprove={async ({ orderId }) => {
+        await fetch(`/api/capture/${orderId}`, { method: "POST" });
+      }}
```
HackTheW2d, line 310:
```suggestion
      onApprove={async ({ orderId }) => {
        const response = await fetch(`/api/capture/${orderId}`,
        {
          method: "POST",
        });
    
        if (!response.ok) {
          throw new Error(`Failed to capture order:
          ${response.status}`);
        }
      }}
```

packages/paypal-js/types/v6/components/paypal-payments.d.ts
HackTheW2d: Found that `usePayPalSavePaymentSession` use `auto` by default. But `onApprove` is also optional.
This would cause:
```
  usePayPalSavePaymentSession({
    createVaultToken,
    // presentationMode omitted => defaults to auto
    // onApprove omitted
  });
```
to pass the static check

packages/paypal-js/types/v6/components/paypal-payments.d.ts
```diff
@@ -79,7 +79,10 @@ export type OnCancelDataSavePayments = {
...
 };
 
-export type PayPalOneTimePaymentSessionOptions = BasePaymentSessionOptions & {
+export type PayPalOneTimePaymentSessionOptions = Omit<
+  BasePaymentSessionOptions,
+  "onApprove"
+> & {
   commit?: boolean;
   orderId?: string;
   onApprove?: (data: OnApproveDataOneTimePayments) => Promise<void>;
```
HackTheW2d, line 88: I think the fix here is onApprove only works when `presentationMode=redirect` represents.
But this type looks like with `auto`, it still won't require `onApprove`
```
  type OnApprove = NonNullable<
    PayPalOneTimePaymentSessionOptions["onApprove"]
  >;

  type CallbackRequirement =
    | {
        presentationMode: "redirect";
        onApprove?: OnApprove;
      }
    | {
        presentationMode?: "auto";
        onApprove: OnApprove;
      }
    | {
        presentationMode:
          | "popup"
          | "modal"
          | "payment-handler"
          | "direct-app-switch";
        onApprove: OnApprove;
      };
```
This might be a possible fix?

.changeset/otp-button-readme-onapprove.md
HackTheW2d, line 1: can we merge 2 changesets?
One mentioned  `include the required onApprove callback`, the other mentioned `Make onApprove genuinely optional`

packages/paypal-js/types/v6/tests/paypal-payments.test.ts
HackTheW2d, line 1: Can we add some negative tests?
eg.
```
  // @ts-expect-error - onApprove must return
  Promise<void>
  sdkInstance.createPayPalOneTimePaymentSession({
    onApprove: () => {},
  });


  // Valid: redirect does not require onApprove
  <PayPalOneTimePaymentButton
    orderId="ORDER-123"
    presentationMode="redirect"
  />;

  // @ts-expect-error - default auto requires onApprove
  <PayPalOneTimePaymentButton orderId="ORDER-123" />;

  // @ts-expect-error - popup requires onApprove
  <PayPalOneTimePaymentButton
    orderId="ORDER-123"
    presentationMode="popup"
  />;
```

packages/paypal-js/types/v6/components/paypal-payments.d.ts
EvanReinstein: I'm not sure what you mean here. We decided offline that we'll make onApprove optional.

.changeset/otp-button-readme-onapprove.md
EvanReinstein: I updated the changeset to indicate we want an optional property.

HackTheW2d (approved):
Thanks for addressing the issue!

packages/paypal-js/types/v6/tests/paypal-payments.test.ts
EvanReinstein: In this case we won't because we are making onApprove optional

PR #1039, opened 2026-08-27
fix(react-paypal-js): handle wallet button hydration and async replacement

OskarEichler:
## Fixes

- Apple Pay: attach the native click listener when hydration renders the actual apple-pay-button. The initial placeholder previously caused the mount-only effect to return without ever attaching a listener.
- Google Pay: give each asynchronous button mount an effect-local cancellation flag. An older readiness response can no longer replace a newer button or mutate its detached container after unmount. Remove the redundant mounted ref and callback layer.

## Compatibility

No public API/type changes. Payment/session behavior is unchanged; these fixes only correct button event attachment and async DOM ownership.

## Verification

React 19/JSDOM mocked SDK: placeholder-to-hydrated Apple Pay click starts exactly one session (previously zero). Google Pay black-to-white option change with readiness promises resolved in reverse order retains white (previously reverted to black). Combined reviewed patch set: existing React suite has 914 passes, 2 pre-existing HostedFieldsProvider failures and 17 passing snapshots, identical to unmodified main. Lint/typecheck, Rollup builds, declarations and diff/format checks pass (five existing JSDoc warnings). No real wallet transactions or device/browser payment certification. No test files changed.

### DIFF
--- a/.changeset/wallet-button-lifecycle.md
+++ b/.changeset/wallet-button-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+Handle wallet button hydration and asynchronous replacement without stale mounts or missing click handlers.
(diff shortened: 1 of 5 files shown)


EvanReinstein:
Hey @OskarEichler would you mind providing tests or a demo that demonstrates the issues that this PR is fixing. Thank you!

PR #1043, opened 2026-08-27
fix(react-paypal-js): compare own keys when memoizing eligibility payloads

OskarEichler:
## Fix

Require each compared object key to be an own property of the other object before recursively comparing its value. Equal key counts and equal property lookups alone do not establish equality: {a: undefined} and {b: undefined} previously compared equal, as could an inherited property standing in for a missing own key.

This prevents deep-comparison memoization from reusing a different eligibility payload. Existing reference, array, depth-limit and value comparison behavior is preserved; no new dependency or public API.

## Verification

Inline reproductions confirm both distinct-undefined-key and inherited-key false positives change from true to false. Combined reviewed patch set: existing React suite has 914 passes, 2 pre-existing HostedFieldsProvider failures and 17 passing snapshots, identical to unmodified main. Lint/typecheck, Rollup builds, declarations and diff/format checks pass (five existing JSDoc warnings). No test files changed.

### DIFF
--- a/.changeset/eligibility-own-key-equality.md
+++ b/.changeset/eligibility-own-key-equality.md
@@ -1,0 +1,5 @@
+---
+"@paypal/react-paypal-js": patch
+---
+
+Compare own keys when memoizing v6 eligibility payloads.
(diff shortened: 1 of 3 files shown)


EvanReinstein:
Can you add tests that demonstrate the issue this is fixing 👍

PR #1045, opened 2026-08-31
Feature/v6 react lpm tree shaking

HackTheW2d:
Add tree-shaking feature to eliminate subpaths.


packages/react-paypal-js/README.md
```diff
@@ -983,7 +983,7 @@ export default function App() {
 }
 ```
 
-> **Note:** `PayPalProvider` and `usePayPal` must be imported from `@paypal/react-paypal-js/sdk-v6/local-payment-methods` (not `@paypal/react-paypal-js/sdk-v6`) when using LPMs — the LPM subpath is a separate bundle with its own React context instance, so a provider from the core subpath will not satisfy the LPM hooks.
+> **Note:** Import `PayPalProvider`, PayPal components, and LPM components from the same `@paypal/react-paypal-js/sdk-v6` entry. When rendered under one `PayPalProvider`, they share one React context and SDK instance.
```
EvanReinstein: I know that this is part of the problem that this PR is solving, but isn't it a little bit of a redundant reminder when there is only one path for v6 components anyhow?

.changeset/lpm-react-wrapper-core.md
```diff
@@ -2,7 +2,7 @@
 "@paypal/react-paypal-js": minor
 ---
 
-Add a React wrapper for all 50 v6 SDK Local Payment Methods (LPM), shipped via a new `@paypal/react-paypal-js/sdk-v6/local-payment-methods` subpath so the default `@paypal/react-paypal-js` export is unaffected.
+Add a React wrapper for all 50 v6 SDK Local Payment Methods (LPM) to the existing `@paypal/react-paypal-js/sdk-v6` entry. The legacy `@paypal/react-paypal-js` entry remains unaffected, while static named v6 imports allow compatible production bundlers to remove unused LPM exports.
```
EvanReinstein: I've seen explanations of the update re: treeshaking here and in the README. I want to make sure that we're not overstating anything. Here is a suggestion to update, but feel free to modify in any way you see fit 👍 
```suggestion
Add a React wrapper for all 50 v6 SDK Local Payment Methods (LPM) to the existing `@paypal/react-paypal-js/sdk-v6` entry. The legacy `@paypal/react-paypal-js` entry remains unaffected, while static named imports let compatible bundlers drop the React wrapper code for unused LPMs. (The shared `LPM_REGISTRY` config table is retained in full, since methods are looked up by dynamic key.).
```

.changeset/lpm-react-wrapper-core.md
```diff
@@ -12,14 +12,14 @@ import {
...
-} from "@paypal/react-paypal-js/sdk-v6/local-payment-methods";
+} from "@paypal/react-paypal-js/sdk-v6";
 ```
 
 - An all-in-one named button, e.g. `IdealOneTimePaymentButton`.
 - A split hook + standalone button, e.g. `useIdealOneTimePaymentSession` + `IdealPaymentButton`.
 - A generic dynamic-selector button, `LPMOneTimePaymentButton`.
 
-The LPM subpath is a separate bundle with its own React context instance, so it re-exports `PayPalProvider`, `usePayPal`, and `INSTANCE_LOADING_STATE` — LPM consumers should obtain the provider from this subpath rather than `@paypal/react-paypal-js/sdk-v6` to ensure their `PayPalProvider` and LPM hooks/buttons share the same context.
+PayPal and LPM components imported from `@paypal/react-paypal-js/sdk-v6` share the same `PayPalProvider`, React context, and SDK instance. LPM factory exports are annotated as side-effect free so unused payment methods can be tree-shaken from the single published client bundle.
```
EvanReinstein: Same thing for the last sentence here as I mentioned above.

packages/react-paypal-js/README.md
```diff
@@ -929,23 +929,23 @@ export default function App() {
 
 ## Local Payment Methods (LPM)
 
-`react-paypal-js` ships pre-built components and hooks for 45+ regional Local Payment Methods (LPMs) — iDEAL, Bancontact, BLIK, Pix, Klarna, and more — each imported by name from the dedicated `@paypal/react-paypal-js/sdk-v6/local-payment-methods` subpath. LPMs are bundled separately from the core SDK to keep the main bundle small, since most integrations only need a handful of region-specific methods.
+`react-paypal-js` ships pre-built components and hooks for 45+ regional Local Payment Methods (LPMs) — iDEAL, Bancontact, BLIK, Pix, Klarna, and more — from the same `@paypal/react-paypal-js/sdk-v6` entry as the other V6 payment methods. Static named imports allow compatible production bundlers to remove unused LPM exports.
```
EvanReinstein: Similar nit as [this one](https://github.com/paypal/paypal-js/pull/1045/changes#r3897183415) above.

packages/react-paypal-js/scripts/test-v6-tree-shaking.mjs
```diff
@@ -0,0 +1,177 @@
...
+const publicV6Specifier = `${packageJson.name}/sdk-v6`;
+
+const expectedSideEffects = new Set([
+  "./index.js",
+  "./dist/cjs/**",
+  "./dist/esm/**",
+  "./dist/v6/esm/server.js",
+  "./dist/v6/esm/server.min.js",
+]);
+const expectedPureAnnotationCount = 152;
```
EvanReinstein: Should this use the registry to derive the pure annotation count?

packages/react-paypal-js/scripts/test-v6-tree-shaking.mjs
```diff
@@ -0,0 +1,177 @@
...
+    "package.json#sideEffects must keep legacy and server artifacts protected without marking the v6 client as side-effectful",
+  );
+  assert.equal(
+    packageJson.exports["./sdk-v6/local-payment-methods"],
+    undefined,
+    "the removed LPM subpath must not return to the public export map",
+  );
+
+  pass("package metadata keeps the intended side-effect and export boundaries");
+}
```
EvanReinstein: The idea of this test is good, but I think in this case it is merely a hardcoded copy being checked against the real thing. Following this strategy also results in a update dependency if there are changes in package.json to `sideEffects`


Claude suggested adding `minimatch v10.2.2^`as a devDep, it is already installed but not imported directly by `react-paypal-js`, you can read more about it [here](https://www.npmjs.com/package/minimatch). This will help you compare the glob that a bundler would use. Happy to pair on this if you'd like

EvanReinstein: To clarify, I am talking about the first `deepEqual` assertion. I think the second `assert.equal` is not necessary.

HackTheW2d: I have moved all tests to a local script. I think it's better to not put any build/packing tests here :)

EvanReinstein (approved):

PR #1050, opened 2026-09-01
fix(react-paypal-js): keep latest message fetch state

OskarEichler:
## Problem

`usePayPalMessages` can retain a failed fetch error after a later fetch succeeds. Concurrent fetches can also complete out of order, allowing an older empty-content response to overwrite the state of a newer successful request. A request finishing after its session changes can update the current hook for an obsolete session.

## Fix

- Clear the current fetch error when a new request starts.
- Track monotonically increasing fetch requests and update error state only for the latest active request.
- Invalidate pending requests when the Messages session changes or is removed.
- Preserve each caller promise result even when its state update is obsolete.
- Add separate regressions for successful retry recovery and reversed completion order, plus a patch changeset.

No public API, returned content, or dependency changes are introduced.

## Verification

- Focused usePayPalMessages suite: 18 tests pass.
- Full React package suite: 72 suites, 918 tests and 17 snapshots pass.
- Package ESLint and TypeScript pass with five pre-existing JSDoc warnings.
- Core and all four React package bundles build.
- Focused Prettier and `git diff --check` pass.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/usePayPalMessages.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/usePayPalMessages.test.ts
@@ -335,6 +335,85 @@ describe("usePayPalMessages", () => {
 
       expect(result.current.error).toBeNull();
     });
+
+    test("should clear a previous fetch error after a successful retry", async () => {
+      const emptyContent = {
+        messageItems: { mainItems: [], actionItems: [] },
+        update: jest.fn(),
+      };
+      const populatedContent = {
+        messageItems: { mainItems: [{ type: "text" }], actionItems: [] },
+        update: jest.fn(),
+      };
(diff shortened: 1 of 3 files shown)


packages/react-paypal-js/src/v6/hooks/usePayPalMessages.test.ts
```diff
@@ -335,6 +335,85 @@ describe("usePayPalMessages", () => {
...
+          logoType: "MONOGRAM",
+        });
+      });
+
+      await act(async () => {
+        resolveFirst(emptyContent);
+        await firstFetch;
+      });
+
+      expect(result.current.error).toBeNull();
```
EvanReinstein, line 419: Similar to my comment on PR 1051, can we also assert that the content is the populatedContent?

PR #1051, opened 2026-09-01
fix(react-paypal-js): keep latest Braintree message fetch state

OskarEichler:
## Problem

`useBraintreePayPalMessages` allows overlapping `fetchContent` calls. If an older call resolves with the empty error sentinel after a newer call succeeds, the older response replaces the current hook state with an error. A completion from a replaced Messages instance can do the same.

## Fix

- Track monotonically increasing fetch request IDs.
- Only let the latest request update hook state.
- Invalidate pending requests when the Messages instance is cleaned up.
- Continue returning each caller’s resolved content, including the empty sentinel.
- Add a reversed-completion regression test.
- Add a patch changeset.

## Verification

- Focused Braintree Messages suite: 13 tests passed.
- Full React suite: 71 suites and 915 tests passed; the two existing timer-sensitive Hosted Fields tests failed unchanged.
- React ESLint and TypeScript checks passed; ESLint reports the same five pre-existing JSDoc warnings.
- Core and all four React package bundles built successfully.
- Changed files pass Prettier and `git diff --check`.

### DIFF
--- a/packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.test.ts
+++ b/packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.test.ts
@@ -299,6 +299,43 @@ describe("useBraintreePayPalMessages", () => {
 
       expect(result.current.error).toBeNull();
     });
+
+    test("should ignore an older fetch error after a newer fetch succeeds", async () => {
+      const emptyContent = {
+        messageItems: { mainItems: [], actionItems: [] },
+        update: jest.fn(),
+      } as unknown as BraintreeMessageContent;
+      const populatedContent = {
+        messageItems: { mainItems: [{ type: "text" }], actionItems: [] },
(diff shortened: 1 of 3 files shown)


packages/react-paypal-js/src/v6/hooks/Braintree/useBraintreePayPalMessages.test.ts
```diff
@@ -299,6 +299,43 @@ describe("useBraintreePayPalMessages", () => {
...
+        firstFetch = result.current.handleFetchContent({ amount: "100" });
+        await result.current.handleFetchContent({ amount: "200" });
+      });
+
+      await act(async () => {
+        resolveFirst(emptyContent);
+        await firstFetch;
+      });
+
+      expect(result.current.error).toBeNull();
```
EvanReinstein, line 343: Can we also check to see which content is returned as well, in order to know that the latest fetched content is used?

PR #1053, opened 2026-09-01
fix(paypal-js): match pending v6 script options

OskarEichler:
## Problem

`loadCoreSdkScript()` currently reuses any pending script whose URL contains `/web-sdk/v6/core`. That selector ignores the environment host, `debug` query, namespace, and integration-source attribute. A request can therefore wait on an incompatible script, resolve the wrong configuration, or reject because its expected namespace was never created.

## Fix

- Match pending scripts by the complete resolved URL.
- Also require the effective namespace and integration source to match.
- Preserve the default `paypal` namespace equivalence when the attribute is omitted.
- Add regressions for environment, debug mode, namespace, and integration-source mismatches.
- Add a patch changeset.

## Verification

- Core source suite: 5 files and 58 tests passed.
- ESLint and TypeScript checks passed.
- ESM, CJS, IIFE, legacy, and v6 bundles built successfully.
- Changed files pass Prettier and `git diff --check`.

### DIFF
--- a/.changeset/v6-match-pending-script-options.md
+++ b/.changeset/v6-match-pending-script-options.md
@@ -1,0 +1,5 @@
+---
+"@paypal/paypal-js": patch
+---
+
+Only reuse an in-flight v6 SDK script when its environment, debug mode, namespace, and integration source match.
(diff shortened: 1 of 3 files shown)


EvanReinstein:
We won't pursue this change currently, thank you.

PR #1058, opened 2026-09-08
Add new fundingSource attribute to OnApproveData types for V6

gregjopa:
The `fundingSource` attribute is now available in the `onApprove(data)` payload for JS SDK V6. This PR updates the types to reflect that.


packages/paypal-js/types/v6/components/base-component.d.ts
```diff
@@ -6,6 +8,7 @@
 /* eslint-enable tsdoc/syntax */
 
 export type OnApproveDataOneTimePayments = {
+  fundingSource: FundingSource;
```
dtjones404, line 11: Are we always passing fundingSource in onApproveData now?  Just wondering if this should be optional

gregjopa: This should always be passed into onApproveData. We can discuss the implementation details offline.

dtjones404 (approved):

EvanReinstein (approved):

HackTheW2d (approved):

PR #1065, opened 2026-09-14
Feat: v6 Venmo sandbox support

EvanReinstein:
This enhances the v6 Venmo types and react hook to handle sandbox support.


HackTheW2d (approved):

AleGastelum (approved):
LGTM

nityasp (approved):
Looks good. Thanks for taking this up.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
```diff
@@ -137,6 +139,7 @@ export function useVenmoOneTimePaymentSession({
     const startOptions = {
       presentationMode,
       fullPageOverlay,
+      sandboxSupport: sandboxSupportRef.current,
```
AleGastelum: If `sandboxSupport` is undefined, would the start options look like the following?:

```
startOptions = {
  //  ...prevOptions
  sandboxSupport: undefined,
}
```

if yes, is this expected? Not sure if we would rather omit `sandboxSupport` if undefined, but It might be functional either way

EvanReinstein: I guess yea it would, which follows the same pattern for fullPageOverlay. I'll update to make this more explicit though.

packages/react-paypal-js/src/v6/hooks/useVenmoOneTimePaymentSession.ts
AleGastelum, line 1: Should we re add the stable ref `useRef` for `sandboxSupport`? My understanding is that the current code would rerun the `useEffect` on any re-ender as `sandboxSupport` dependency would be a new object each time, unless I'm missing something

EvanReinstein: Not necessarily. This value is being used in a `useCallback` rather than a `useEffect`. Nothing in the hook depends on `sandboxSupport` so it won't trigger any of the teardown/re-creation issues that would re-trigger the session creation `useEffect`

PR #994, opened 2026-07-13
feat(react-paypal-js-storybook): add v6 LPM stories

Mathanjeevika:
## Summary
- Adds Storybook stories, a shared story factory (`shared/lpm/{code,storyFactory,utils}.ts`), and a dedicated provider decorator (`LPMPayPalProviderDecorator.tsx`) for all 50 v6 SDK Local Payment Methods.
- Adds a `generate:lpm-stories` script that regenerates the 50 per-LPM story files from the LPM registry.
- Small hand-merged additions to existing shared files: a `skipPayPalProvider` opt-out on `withPayPalProvider`, an `LPM` group in the Storybook sidebar order, and the new decorator's barrel export.

## Context
This is PR 3 of 3 split out from #926 per review feedback requesting the original monolithic PR be broken up:
1. #992 — `paypal-js` v6 LPM type declarations
2. #993 — `react-paypal-js` core LPM wrapper
3. **this PR** — Storybook stories for LPM components (this PR)

**This branch is stacked on #993's branch** (`feature/lpm-react-wrapper-core`), since the LPM stories import `LPM_REGISTRY`-driven exports from `@paypal/react-paypal-js/sdk-v6/local-payment-methods`. Because the head is on a fork, GitHub can only diff against `main`, so **this PR's diff currently also includes #993's commit** until #993 merges — please review only the last commit (`feat(react-paypal-js-storybook): add v6 LPM stories`) for now, or hold review until #992 and #993 land and this is rebased to show a clean diff.

## Test plan
- [x] `npx tsc --noEmit` passes for the storybook v6 package
- [ ] no `lint`/test script exists for this Storybook package (not part of CI validation for this package)

PR #1066, opened 2026-09-16
feat(storybook): add stories for all v6 LPMs

HackTheW2d:
## Summary

- add Storybook examples for all 50 registered v6 Local Payment Methods
- share eligibility, currency, and session setup through a common factory and generated story files
- align LPM session-field metadata with the current core Web SDK implementation

Rebuilds the LPM Storybook coverage from current `main`, using the reusable story pattern explored in #994.

## Testing

- LPM story generation check
- React LPM registry tests
- Storybook TypeScript check
- Storybook production build

https://github.com/user-attachments/assets/1b1848e7-a6c0-4baf-a600-27ba1e0408c7


EvanReinstein (approved):
Looks good 👍

